PR #32, opened 2016-04-12
Change the way of specifying jar versions

timyitong:
We need the ability to control the version number we want to publish differently between internal and external builds. gradle.properties isn't so easy to override via an adapt script in internal builds. Changing it to using specific configuration in jar tasks seems to be doing the trick. Tested both in this build and internally.


TDDFT:
LGTM

PR #33, opened 2016-04-13
Add jacoco and coverall integration which provides test-coverage check

timyitong:
I'm still experimenting via this PR to make sure all set ups are correct. (Because PR could help me trigger Travis CI build : P)

Hopefully, if this goes well, we could have test-coverage tag in addition to build pass tag.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,4 +24,6 @@ before_script:
   - sudo service rsync stop || true
   - sudo service x11-common stop || true
 script:
-  - ./gradlew check -x integTest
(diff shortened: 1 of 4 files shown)


timyitong:
jacoco can somehow significantly drag down our test speed (from 4 min to 28 min in Travis). Not sure whether this is okay actually. Maybe we should try some alternatives other than jacoco?

XianXing:
Nice, at least Travis CI is working. J

fastier-li:
Replaced by #249.

Issue #35, opened 2016-04-19
"afterMethod" in TestTemplateWithTmpDir doesn't clean the created tmp directories.


XianXing:
In trait ```TestTemplateWithTmpDir```, the ```getTmpDir``` method would create a tmp directory with the current time in milliseconds as part of the directory's name.

As a result, when ```getTmpDir``` is called in the ```afterMethod``` method, it would create a new directory which has a different name compared to the directories created previously,  which are supposed to be cleaned. Consequently, those previously created tmp directories won't be cleaned after the ```afterMethod``` is called.

PR #36, opened 2016-04-20
Bug fix for Issue #35: afterMethod in TestTemplateWithTmpDir doesn't …

XianXing:
…clean the created tmp directories properly. Also made some minor Scala code style changes.

### DIFF
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/util/PhotonLoggerIntegTest.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/util/PhotonLoggerIntegTest.scala
@@ -17,17 +17,17 @@ package com.linkedin.photon.ml.util
 import org.testng.annotations.{DataProvider, Test}
 import org.testng.Assert._
 
-import java.nio.file.{Files, FileSystems, Path}
+import java.nio.file.{Files, FileSystems}
 import scala.io.Source
 
-import com.linkedin.photon.ml.test.{CommonTestUtils, SparkTestUtils, TestTemplateWithTmpDir}
+import com.linkedin.photon.ml.test.{SparkTestUtils, TestTemplateWithTmpDir}
 
 class PhotonLoggerIntegTest extends SparkTestUtils with TestTemplateWithTmpDir {
 
   class TestException extends Exception
 
   @Test
-  def testSingleLogMessage = sparkTest("singleLogMessage") {
+  def testSingleLogMessage() = sparkTest("singleLogMessage") {
     val logFile = s"$getTmpDir/singleLogMessage"
     val logger = new PhotonLogger(logFile, sc)
 
@@ -35,19 +35,19 @@ class PhotonLoggerIntegTest extends SparkTestUtils with TestTemplateWithTmpDir {
       logger.error("test message")
 
     } finally {
-      logger.close
+      logger.close()
     }
 
-    val fs = FileSystems.getDefault()
+    val fs = FileSystems.getDefault
     assertTrue(Files.exists(fs.getPath(logFile)))
 
-    val lines = Source.fromFile(logFile).getLines.toArray
-    assertEquals(lines.size, 1)
+    val lines = Source.fromFile(logFile).getLines().toArray
+    assertEquals(lines.length, 1)
     assertTrue(lines(0).matches("^[0-9T\\-\\:\\.]* \\[ERROR\\] test message$"))
   }
 
   @Test
-  def testMultipleLogMessages = sparkTest("multipleLogMessages") {
+  def testMultipleLogMessages() = sparkTest("multipleLogMessages") {
     val logFile = s"$getTmpDir/multipleLogMessages"
     val logger = new PhotonLogger(logFile, sc)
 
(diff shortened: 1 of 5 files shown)


XianXing:
@TDDFT Please take a look.

photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
```diff
@@ -24,15 +27,25 @@ import org.testng.annotations.AfterMethod
...
+    tmpDirs.synchronized {
+      tmpDirs += tmpDirName
+    }
+    tmpDirName
+  }
 
   @AfterMethod
   def afterMethod(): Unit = {
-    FileUtils.cleanDirectory(new File(getTmpDir))
+    tmpDirs.foreach(tmpDir => FileUtils.cleanDirectory(new File(tmpDir)))
```
TDDFT: According to the doc, I think you need this:
https://commons.apache.org/proper/commons-io/apidocs/org/apache/commons/io/FileUtils.html#deleteDirectory(java.io.File)

photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
```diff
@@ -24,15 +27,25 @@ import org.testng.annotations.AfterMethod
...
-  def getTmpDir: String = TestTemplateWithTmpDir._tmpDirThreadLocal.get()
+  def getTmpDir: String = {
+    val tmpDirName = TestTemplateWithTmpDir._tmpDirThreadLocal.get()
+    tmpDirs.synchronized {
+      tmpDirs += tmpDirName
+    }
+    tmpDirName
+  }
 
   @AfterMethod
```
TDDFT: I think in multithreading scenario, AfterMethod should probably changed to AfterClass. Currently, we are setting parallel level of "methods" (https://github.com/linkedin/photon-ml/blob/master/build.gradle#L115). Different methods in the same class will share the same class variables. These directory should only be deleted once. For "methods" parallelism, AfterClass is only executed once, while AfterMethod will be executed multiple times (by different threads). So I think this should be AfterClass.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/util/PhotonLoggerIntegTest.scala
```diff
@@ -62,14 +62,14 @@ class PhotonLoggerIntegTest extends SparkTestUtils with TestTemplateWithTmpDir {
...
     }
 
-    val fs = FileSystems.getDefault()
+    val fs = FileSystems.getDefault
     assertTrue(Files.exists(fs.getPath(logFile)))
 
-    val lines = Source.fromFile(logFile).getLines.toArray
-    assertEquals(lines.size, 5)
+    val lines = Source.fromFile(logFile).getLines().toArray
+    assertEquals(lines.length, 5)
```
joshvfleming: @XianXing I'm curious about the rationale for both these changes. ```getLines``` is not side-effecting. Also is there a conventional preference for ```length``` over ```size```?

photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
joshvfleming: Yeah. This could potentially delete another thread's temporary folder before that thread is done with its tests.

It's strange that ```FileUtils.forceDeleteOnExit``` isn't working here. Any idea why?

photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
XianXing: Good point!

photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
XianXing: @TDDFT Good to know, thanks!
@joshvfleming As mentioned in Issue #35, the problem we would like to address in this PR is that the afterMethod was not working properly. But FileUtils.forceDeleteOnExit is actually working fine, although it provides same functionality as the AfterClass method. So we may want to keep either the AfterClass method, or the FileUtils.forceDeleteOnExit method, but maybe not both.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/util/PhotonLoggerIntegTest.scala
XianXing: @joshvfleming For this change, I actually agrees with my IntelliJ editor that getLines() is defined as an empty paren method (in http://www.scala-lang.org/api/rc2/scala/io/BufferedSource.html), and was accessed as parameterless.

For the preference for length over size: calling size may require additional implicit conversion to SeqLike to be made. Calling length on Array or Strings may provide advantages. To further verify this, I found a few similar questions asked on StackOverflow, for example: http://stackoverflow.com/questions/19826532/do-size-and-length-in-scala-collections-generate-the-same-byte-code. There are a few more discussions there :) 

But I agree with you that these are very minor things, and I don't have very strong preference neither.

TDDFT:
LGTM

PR #37, opened 2016-04-20
Fix scalaStyle warnings

joshvfleming:


photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/NameAndTerm.scala
```diff
@@ -18,13 +18,21 @@ package com.linkedin.photon.ml.avro.data
...
+//after Avro related classes/functons are decoupled from the rest of code
 protected[ml] case class NameAndTerm(name: String, term: String) {
 
   override def hashCode: Int = {
     (name + NameAndTerm.DELIMITER + term).hashCode
   }
 
+  override def equals(that: Any): Boolean = {
+    that match {
+      case other: NameAndTerm => hashCode == other.hashCode
```
XianXing: Why do we define two NameAndTerm instances to be the same if their hashCodes are the same? I don't oppose this change, but would like to understand the motivation or use case better.

XianXing:
LGTM.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/NameAndTerm.scala
joshvfleming: Generally you're always supposed to override these two in pairs. The reason is that ```hashCode``` determines an object's identity (in the place of objectId), and ```equals``` (in most cases) is an identity check. Otherwise you can get into weird situations where e.g. multiple objects are collapsed into one in a set, but they're not "equal" to each other. So it's usually best to keep these two in sync.

This could be an exception though -- e.g., perhaps two NameAndTerm values should be considered distinct, even if their name and term values are the same?

XianXing: When I looked at this change, the first two things came to my mind were Long and String. In some cases, two longs (or two strings) may share the same hash code, but they are still different when compared by the ```equals``` method, because the ```equals``` method has nothing to do with the ```hashCode``` in those two cases.

For NameAndTerm, it appears to me that two NameAndTerm instances may have hash code collision (due to String hash code collision) in some rare situations, but they still should be considered as different NameAndTerm instances.

Anyway, I feel like I am now trying to dig into some small details that may not matter at all in real life. Please go ahead and merge this RB when it's ready. Thanks for fixing all the Scala stye issues by the way :)

mtth: @XianXing is definitely right. This `equals` method has the potential to cause confusion at best, data corruption at worst. (Collisions aren't that uncommon either: `"Aa"` and `"BB"` have the same hash code.)

joshvfleming: @XianXing @mtth You're right. I wasn't paying attention to what the ```hashCode``` calculation was actually doing. Will fix.

Issue #42, opened 2016-04-21
Adding local AUC computation to the evaluation package


XianXing:
Currently the only AUC computation supported in Photon/GAME is ```RDD``` based, e.g., the input scores and labels are in the ```RDD[(Double, Double)]``` format.

However, in order to enable per random effect AUC computation as in issue #41, computing AUC locally in the ```Array[(Double, Double)]``` or ```Seq[(Double, Double)]``` format will be a useful feature to have.

Issue #41, opened 2016-04-21
Enable AUC per random effect computation.


XianXing:
Currently the AUC evaluation metric is computed on the overall testing data set. However, it's useful to include a per-random effect level AUC computation, e.g., compute AUC at the memberId level for each member's data, and take a weighted average to obtain the final AUC.

This per random effect AUC evaluation metric is similar in spirit to the row/col oriented mean percentile rank proposed in [1] and the row/col oriented AUC found in [2], with row and col Ids being the corresponding random effect Id.

[1] Steffen Rendle, Christoph Freudenthaler, Zeno Gantner, Lars Schmidt-Thieme. BPR: Bayesian Personalized Ranking from Implicit Feedback 
[2] LibMF: https://www.csie.ntu.edu.tw/~cjlin/libmf/

XianXing:
@ajithmurali @ankans @TDDFT Since you guys are also looking at this metric, please provide any of your requests or suggestions on this to-be-added feature here, then we can figure out a detailed plan before adding this feature to GAME.

XianXing:
Since this issue is tightly related to PR #162 #164 #166 #169 and issue #151, I will grab it and get it done.

PR #43, opened 2016-04-21
Adding local AUC computation to the evaluation package

XianXing:
As discussed in issue #42, currently the only AUC computation supported in Photon/GAME is ```RDD``` based, e.g., the input scores and labels are in the ```RDD[(Double, Double)]``` format.

However, in order to enable per random effect AUC computation as in issue #41, computing AUC locally in the ```Array[(Double, Double)]``` or ```Seq[(Double, Double)]``` format will be a useful feature to have.

Besides, the local AUC computation will be helpful to our effort in evaluating different matrix factorization packages as well.

One more change made in this PR is that, now the local AUC computation will take the weight of each data point into consideration when computing AUC.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/AreaUnderROCCurveLocalEvaluator.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/AreaUnderROCCurveLocalEvaluator.scala
@@ -34,8 +34,6 @@ protected[ml] class AreaUnderROCCurveLocalEvaluator(
 
   def this(labels: Map[Long, Double]) = this(labels.mapValues(label => (label, DEFAULT_OFFSET, DEFAULT_WEIGHT)))
 
-  val numLabels = labelAndOffsetAndWeights.size
-
   /**
    * Evaluate the scores of the model
    *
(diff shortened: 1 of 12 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/AreaUnderROCCurveLocalEvaluator.scala
```diff
@@ -0,0 +1,66 @@
...
+      if (label > MathConst.POSITIVE_RESPONSE_THRESHOLD) {
+        positiveCount += weight
+      } else {
+        aucAccum += positiveCount
+        negativeCount += weight
+      }
+    }
+
+    //normalize ACU over the total area
+    aucAccum / (positiveCount * negativeCount)
```
joshvfleming: @XianXing I believe there's a subtle flaw in this computation. If we have two identical scores with conflicting ground-truth labels, we're accumulating a little too much area for the "true" one. For example, in the following case:

```
labels = [0, 1, 0, 0, 1, 1, 1]
scores = [-0.1, -1, 0, -1, 2, 6, 8]
```

we compute an AUC of **0.8333** here, but it should be **0.7917**.

joshvfleming: @XianXing I believe there's a subtle flaw in this computation. If we have two identical scores with conflicting ground-truth labels, we're accumulating a little too much area for the "true" one. For example, in the following case:

```
labels = [0, 1, 0, 0, 1, 1, 1]
scores = [-0.1, -1, 0, -1, 2, 6, 8]
```

we compute an AUC of **0.8333** here, but it should be **0.7917**.

joshvfleming:
@XianXing I believe there's a subtle flaw in this computation. If we have two identical scores with conflicting ground-truth labels, we're accumulating a little too much area for the "true" one. For example, in the following case:

```
labels = [0, 1, 0, 0, 1, 1, 1]
scores = [-0.1, -1, 0, -1, 2, 6, 8]
```

we compute an AUC of **0.8333** here, but it should be **0.7917**.

ashelkovnykov:
LGTM, aside from the AUC concerns.

XianXing:
Very good point @joshvfleming, the previous algorithm used indeed ignored the case you mentioned. I've updated the algorithm and the unit tests accordingly to cover it. Please take a look.

namitk:
@XianXing: Before merging, could you rebase -i the commits to squash the extraneous, minor fixes? I believe this PR can be two commits, the first one and then the rest merged under "fix bug" or "more rigorous and robust algorithm..."

XianXing:
Good suggestion @namitk, now the commit history is cleaned a bit.

joshvfleming:
Looks good!

Issue #45, opened 2016-04-26
Coefficients type should be flexible when loaded from Avro model file


XianXing:
When the model coefficients are loaded from Avro model file via the ```AvroUtils.loadMeanVectorFromBayesianLinearModelAvro``` function, the type of the returned coefficients is ```DenseVector```. 

However in some cases when the underlying coefficients has large dimension but relatively small support, they should be represented as a ```SparseVector```.

It would be helpful if the coefficients type can be flexible when loaded from Avro model file, instead of always being a ```DenseVector[Double]```, for example, the coefficient type could be either dense vector or sparse vector based on the sparsity of the underlying array.

PR #46, opened 2016-04-27
Coefficients type should be flexible when loaded from Avro model file

XianXing:
As discussed in Issue #45, it would be helpful if the coefficients type can be flexible when loaded from Avro model file, instead of always being a ```DenseVector[Double]```, for example, the coefficient type could be either dense vector or sparse vector based on the sparsity of the underlying array. 

This will be helpful in cases when the underlying dimension of the coefficient space (array length) is large, but the actual support (active size) is small.

Also some unit tests are added in this PR.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
@@ -17,6 +17,7 @@ package com.linkedin.photon.ml.avro
 
 import scala.collection.JavaConversions._
 import scala.collection.Map
+import scala.collection.mutable
 
 import breeze.linalg.{DenseVector, SparseVector, Vector}
 import org.apache.avro.generic.GenericRecord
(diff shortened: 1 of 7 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -91,27 +92,28 @@ object AvroUtils {
...
       modelId: String,
       intToNameAndTermMap: Map[Int, NameAndTerm]): BayesianLinearModelAvro = {
 
-    val meansAvros = vectorToArrayOfNameTermValueAvros(model.means, intToNameAndTermMap)
-    val variancesAvrosOption = model.variancesOption.map(variances => vectorToArrayOfNameTermValueAvros(variances,
-      intToNameAndTermMap))
+    val meansAvros = vectorToArrayOfNameTermValueAvros(coefficients.means, intToNameAndTermMap)
+    val variancesAvrosOption = coefficients.variancesOption.map(variances =>
+      vectorToArrayOfNameTermValueAvros(variances, intToNameAndTermMap)
+    )
```
joshvfleming: Minor: it's debatable whether this is actually an improvement, but I'm a big fan of point-free style:

```scala
val variancesAvrosOption = coefficients.variancesOption.map(
  vectorToArrayOfNameTermValueAvros(_, intToNameAndTermMap))
```

joshvfleming:
LGTM

ashelkovnykov:
LGTM

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
XianXing: Thanks for pointing this out, I also like the code to be concise!

XianXing:
Cleaned the commit history a bit.

Issue #49, opened 2016-05-02
Bumping up tolerance parameter for GAME's per-random effect feature selection


XianXing:
Currently the per-random effect feature selection algorithm used in GAME is based on computing Pearson correlation score, and in particular, the [one pass algorithm](https://en.wikipedia.org/wiki/Pearson_product-moment_correlation_coefficient) is adopted.

However, since the one pass algorithm is numerically unstable for certain cases, the current tolerance parameter used is 1e-12 which might not work super well together with the one pass algorithm.

We may want to bump up the tolerance parameter (to be addressed in the issue) and figure out a numerically more stable algorithm down the road (to be addressed in issue #50).

PR #51, opened 2016-05-02
Bumping up the tolerance used with per-random effect Pearson correlat…

XianXing:
…ion score one-pass algorithm, and make the diagnostic information more useful. More details can be found in issue #49.


ashelkovnykov:
LGTM

PR #52, opened 2016-05-03
Enable integration tests on Travis CI

ashelkovnykov:
- Modified PhotonLoggerIntegTest to have a more accurate regex (previously it assumed the time zone would always be West of GMT)
- Added custom script for testing on Travis CI
- Split unit tests and integration tests to run in parallel
- Integration test output is redirected to a tmp file, to get around the 4MB log file size issue. Setup up a dummy process to ping Travis every 30s so it doesn't kill the test process for inactivity. Dump the last 500 lines whenever tests complete successfully or an error occurs.

Much of the script comes from here:
http://stackoverflow.com/questions/26082444/how-to-work-around-travis-cis-4mb-output-limit

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 os:
   - linux
-language: java
+language: scala
 jdk:
   - oraclejdk8
 before_cache:
@@ -9,8 +9,6 @@ cache:
   directories:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
-install:
-  - ./gradlew assemble
 before_script:
   - sudo service postgresql stop || true
   - sudo service mysql stop || true
@@ -23,5 +21,8 @@ before_script:
   - sudo service riak stop || true
   - sudo service rsync stop || true
   - sudo service x11-common stop || true
+env:
+  - TESTS=unit
+  - TESTS=integration
 script:
-  - ./gradlew check -x integTest
+  - ./tests.sh $TESTS
--- a/build.gradle
+++ b/build.gradle
@@ -86,7 +86,8 @@ rat {
     'photon-ml/LINEAR_REGRESSION*/**',
     'photon-ml/LOGISTIC_REGRESSION*/**',
     'photon-ml/POISSON_REGRESSION*/**',
-    'photon-test/src/test/resources/**'
+    'photon-test/src/test/resources/**',
+    'tests.sh'
   ]
 }
 
--- a/tests.sh
+++ b/tests.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+#
+# This script is used to customize Travis CI testing as follows:
+# - Wrapper for gradle call to run unit and integration tests in parallel
+# - Redirect integTest output to tmp file to avoid Travis CI 4MB log file limit. In addition, ping Travis CI while
+#   testing to avoid the 10 minute no-output timeout. This code modified from the base script found here:
+#   http://stackoverflow.com/questions/26082444/how-to-work-around-travis-cis-4mb-output-limit
+#
+
+NUM_LINES=500
+
+# Helper functions
+dump_output() {
+  echo "Dumping the last $NUM_LINES lines of output:"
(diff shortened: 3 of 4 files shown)


tests.sh
```diff
@@ -0,0 +1,65 @@
...
+  # Kill ping process, output final
+  clean_up
+
+elif [[ "$1" == "unit" ]]; then
+  ./gradlew check -x integTest
+
+else
+  echo "ERROR: Invalid test type specified; must be either 'unit' or 'integration'"
+  exit 1
+fi
```
XianXing: Nitpick: Can we add a newline at the end of the file?

I once read from somewhere that adding a newline at the end of the file is a good practice for java/scala/c++/..., but couldn't find it. Instead, I found the following discussion:
http://stackoverflow.com/questions/729692/why-should-text-files-end-with-a-newline

XianXing:
LGTM!

tests.sh
joshvfleming: Yep. Also, scalaStyle will produce a warning for files that don't end with newline.

tests.sh
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env bash
+# Modified version of script from here:
+#   http://stackoverflow.com/questions/26082444/how-to-work-around-travis-cis-4mb-output-limit
```
joshvfleming: It's unfortunate that we need a wrapper script here. Could we just suppress the gradle output instead? E.g.:

```groovy
  testLogging {
    showStandardStreams = false
  }
```

We could always set this back to ```true``` when we need to debug locally.

build.gradle
```diff
@@ -86,7 +86,8 @@ rat {
     'photon-ml/LINEAR_REGRESSION*/**',
     'photon-ml/LOGISTIC_REGRESSION*/**',
     'photon-ml/POISSON_REGRESSION*/**',
-    'photon-test/src/test/resources/**'
+    'photon-test/src/test/resources/**',
+    'tests.sh'
```
timyitong: Maybe the build script should also append a license instead of getting ignored?

tests.sh
ashelkovnykov: Will do, slipped my mind.

tests.sh
ashelkovnykov: I did a test with suppressed output and it doesn't produce any input unless an error occurs. We'd need to ping Travis or split the integTest task into several tasks that take less than 10 min to run to avoid the output timeout.

build.gradle
ashelkovnykov: I asked the Open Source department about this and they agree that it's not mandatory, since we include the license and notice files for the distribution. However, they're split on whether it is best practice to include a header anyway.

I'll add it, better safe than sorry

ashelkovnykov: Open Source department messaged me again, they've resolved their disagreements. They feel that it's best practice for build/test scripts, project management, and non-core code files in general to not have license headers. That was my line of reasoning originally, so I'm going to leave it as is.

.travis.yml
```diff
@@ -9,8 +9,6 @@ cache:
   directories:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
-install:
-  - ./gradlew assemble
```
XianXing: @ashelkovnykov Sorry to ping you on this again, just out of curiosity, what were the following two lines used for in travis, did travis rely on them to compile Photon? And why did we remove them in this PR?
```
install:
  - ./gradlew assemble
```

ashelkovnykov: Those two lines are run to install dependencies before tests:
https://docs.travis-ci.com/user/customizing-the-build/#Customizing-the-Installation-Step

For Travis CI, Scala projects that don't use sbt behave the same as Java projects. If Travis CI for a Java project detects the gradle wrapper, it will automatically call `./gradlew assemble` during install:
https://docs.travis-ci.com/user/languages/java#Projects-Using-Gradle

I chose to cut those two lines because they were redundant. Hope this helps!

XianXing: This helps a lot, thanks for taking time finding those references!

PR #57, opened 2016-05-04
Fix utf8 string output in IOUtils

TDDFT:
The current implementation uses DataOutputStream.writeBytes interface which casts char to byte and write only one byte per char. 

`   public final void writeBytes(String s) throws IOException {
        int len = s.length();
        for (int i = 0 ; i < len ; i++) {
            out.write((byte)s.charAt(i));
        }
        incCount(len);
    }
`
But Java char uses UTF encoding so sizeof(char) = 2 bytes (https://docs.oracle.com/javase/8/docs/api/java/lang/Character.html) and the casting will cause troubles for international languages used in feature names and terms.


XianXing:
LGTM.

ashelkovnykov:
LGTM

PR #67, opened 2016-05-09
Fix feature name collision / strictness issues in box constraints

joshvfleming:
This pull request addresses the following two issues:

 * A bug in box constraints where we'd incorrectly treat two features as identical if there was a substring match between the feature names
 * A minor issue where we'd allow users to specify empty constraints (e.g. for typos in upperBound, lowerBound or both)


joshvfleming:
@namitk Please review

namitk:
Thanks for the fix, that was quite a rookie mistake :) 
LGTM, will merge after travis checks complete

joshvfleming:
Not at all, @namitk. Otherwise I must be the oldest rookie ever, because I still do stuff like this all the time. :-)

PR #68, opened 2016-05-09
Modify repository declarations for root Gradle buildscript

ashelkovnykov:
I think this may be the cause for our occasional deployment issue. See me for more details.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,17 +18,9 @@
 buildscript {
   repositories {
     jcenter()
-  }
-
-  // TODO: is this the right way to incluse gradle plugins?
-  repositories {
-    repositories {
-      maven {
-        // For license plugin gradle-rat.
-        url 'http://dl.bintray.com/content/netflixoss/external-gradle-plugins/'
-        // For avro compiler plugin gradle-avro-plugin
-        url 'https://plugins.gradle.org/m2/'
-      }
+    maven {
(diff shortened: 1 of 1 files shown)


ashelkovnykov:
@joshvfleming Please review

build.gradle
```diff
@@ -18,17 +18,13 @@
...
-      }
+    maven {
+      // For license plugin gradle-rat.
+      url 'http://dl.bintray.com/content/netflixoss/external-gradle-plugins/'
+    }
+    maven {
+      // For avro compiler plugin gradle-avro-plugin
+      url 'https://plugins.gradle.org/m2/'
     }
   }
```
XianXing: Can you please briefly explain what is the problem of the previous approach, so that latter folks who look at this PR can get a rough idea?

ashelkovnykov: - Multiple repositories can be listed in one block, breaking them into 2 blocks is unnecessary.
- Repository blocks shouldn't be nested
- Each Maven repository block should contain only 1 URL
- The repository at the first Maven URL above is unnecessary: the rat plugin jar is available on jcenter (and it was likely being overwritten by the second URL anyway)

I believe that's it

ashelkovnykov: Just realized the above comments are on an outdated diff. Check the most recent commit, there's only 1 Maven block in it.

XianXing: Thanks!

Issue #56, opened 2016-05-04
Save MatrixFactorizationModel to HDFS as Avro files


XianXing:
This issue is a sub-task of #58.

We may need this issue resolved before adding the actual learning algorithm for MF because that, this functionality can be used to save the MF model learned from org.apache.spark.ml.recommendation.ALS to HDFS in Avro format. And such learned MF model can then be used in other downstream scenarios, for example, when the row and column latent factors of the MF model are used as features in a response prediction model.

XianXing:
This is a sub-task

XianXing:
This issue should be addressed after #59 is full filled by a PR.

PR #72, opened 2016-05-12
Adding Avro I/O util functions for matrix factorization model

XianXing:
This PR has three commits:

1. Fix the bug when creating the temporary dir
2. Change the function name in Avro I/O util to make it more consistent
3. Addresses issue #56 and provides Avro I/O util functions for the matrix factorization model.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
@@ -15,19 +15,21 @@
 package com.linkedin.photon.ml.avro
 
 
-import scala.collection.JavaConversions._
-import scala.collection.Map
-import scala.collection.mutable
+import java.lang.{Double => JDouble}
+import java.util.{List => JList}
 
-import breeze.linalg.{DenseVector, SparseVector, Vector}
+import scala.collection.JavaConversions._
+import scala.collection.{Set, Map, mutable}
+
+import breeze.linalg.{Vector, DenseVector, SparseVector}
 import org.apache.avro.generic.GenericRecord
 import org.apache.avro.mapred.{AvroInputFormat, AvroWrapper}
 import org.apache.hadoop.io.NullWritable
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 
-import com.linkedin.photon.ml.avro.generated.{BayesianLinearModelAvro, NameTermValueAvro}
-import com.linkedin.photon.ml.avro.data.NameAndTerm
+import com.linkedin.photon.ml.avro.generated.{BayesianLinearModelAvro, NameTermValueAvro, LatentFactorAvro}
+import com.linkedin.photon.ml.avro.data.{NameAndTermFeatureSetContainer, NameAndTerm}
 import com.linkedin.photon.ml.constants.MathConst
 import com.linkedin.photon.ml.model.Coefficients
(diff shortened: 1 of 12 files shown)


photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
```diff
@@ -35,7 +35,7 @@ trait TestTemplateWithTmpDir {
 }
 
 private object TestTemplateWithTmpDir {
-  private val _tmpDirThreadLocal: ThreadLocal[String] = new ThreadLocal[String] {
+  private def _tmpDirThreadLocal: ThreadLocal[String] = new ThreadLocal[String] {
```
joshvfleming: Just curious: what was the bug? This is a pretty big change to the semantics of ```getTmpDir```. Previously, it defined a temporary base dir that would be constant for all tests running in a particular test class and thread. With this change, a new directory is created for each call to ```getTmpDir```. Also, this doesn't really need to be a ```ThreadLocal``` anymore, since it's not shared between the threads.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -86,28 +97,89 @@ object AvroUtils {
...
+   */
+  protected[avro] def readNameAndTermSetFromGenericRecords(
+    genericRecords: RDD[GenericRecord],
+    featureSectionKey: String): Set[NameAndTerm] = {
+
+    genericRecords.flatMap(_.get(featureSectionKey) match { case recordList: JList[_] =>
+      val nnz = recordList.size
+      val featureNameAndTermBuf = new mutable.ArrayBuffer[NameAndTerm](nnz)
+      val iterator = recordList.iterator
+      while (iterator.hasNext) {
```
joshvfleming: Using JavaConverters here would yield slightly more functional / idiomatic Scala code, e.g.:

```scala
import collection.JavaConverters._

val featureNameAndTermBuf = recordList.asScala.map {
  case record: GenericRecord => AvroUtils.readNameAndTermFromGenericRecord(record)
  case any => throw new IllegalArgumentException(s"$any in features list is not a record")
}
```

I wonder if there's a performance concern, though?

photon-test/src/main/scala/com/linkedin/photon/ml/test/TestTemplateWithTmpDir.scala
XianXing: In some cases we may call getTmpDir a few times for the same test, for example, in a test with ```DataProvider``` that provides multiple parameters and thus multiple run of the same test. In such cases, different runs of the same test would have to share the same tmpDir (because it's defined as a val of a singleton) and exceptions like "directory *** already exists" may arise.

In such case, call by name might be a better choice than call by value.

For ```ThreadLocal```, it might still be useful in case two threads trying to use the same directory, although the chance for that is very slim given the way we name the tmp directory.

But I agree that we can remove ThreadLocal from this function.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
XianXing: This is a good suggestion. The overhead comes with JavaConverters will have minimal impact, because this it is not the bottleneck of this function.

PR #74, opened 2016-05-13
Improve error reporting for empty training set

joshvfleming:
This arose in response to a recent user issue. The feature vectors in
their training set were empty, and we failed on the call to
trainingData.first() with an unhelpful error message. This change
improves the error reporting for this case.


photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
```diff
@@ -169,19 +169,26 @@ protected[ml] class Driver(
...
+
+    // Read training data
     trainingData = suite.readLabeledPointsFromAvro(
       sc, params.trainDir, params.selectedFeaturesFile, params.minNumPartitions)
       .persist(trainDataStorageLevel).setName("training data")
-    featureNum = trainingData.first().features.size
+
     trainingDataNum = trainingData.count().toInt
+    require(trainingDataNum > 0,
+      "No training data found. Ensure that training data exists and feature vectors are not empty.")
```
XianXing: Thanks for making the error reporting more informative.

One suggestion:  ```trainingDataNum > 0``` and "feature vectors are not empty" are two independent things, because we can still have the case where ```trainingDataNum > 0``` while ```featureNum == 0```. Shall we decouple those two things from the error message, and only say "No training data found. Ensure that training data exists."?

ashelkovnykov:
LGTM

photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
XianXing: After a second look at ```GLMSuite```, I realized that the problem is caused by the fact that the created data will be empty if the features size is 0. So this error message actually reflects how we deal with data set with empty features in ```GLMSuite```.

I didn't know why GLMSuite was designed in that way, and will open an issue to separate logic of empty feature and empty data set checks.

joshvfleming: That's right, @XianXing. We map over the records and return ```None``` if the feature vector is empty, and then we ```flatMap``` over the result. This means that a dataset of empty feature vectors becomes an empty collection.

Issue #77, opened 2016-05-17
Default value for ```isNullOK``` in ```getStringAvro``` should not be true.


XianXing:
Currently the default value of ```isNullOK``` in the ```getStringAvro``` function of ```com.linkedin.photon.ml.util.Utils``` is set to ```true```. 

However, sometimes Photon's user/developer may not be aware of or overlook such default setting, which would lead to unexpected code behavior, because when the corresponding field of the key is null in the Avro record, an empty string will always be returned as the retrieved result. And such behavior is inconsistent with other ```get*Avro``` functions, where an exception will be thrown if the field is null.

As a result, we may want to set the default value of ```isNullOK``` in the ```getStringAvro``` function to ```false```.

PR #78, opened 2016-05-17
Changing the default value for isNullOk to false.

XianXing:
This PR changes the default value for ```isNullOK``` to false, in order to make it consistent with other ```get*Avro``` functions.

This PR also makes sure that ```getFeatureKey``` and ```readNameAndTermFromGenericRecord``` functions are sharing the same the same logic, which is that the term is allowed to be null.

See #77 for more details.


joshvfleming:
LGTM

Issue #84, opened 2016-05-19
Output best GAME model even when 'save-models-to-hdfs' is false


ashelkovnykov:
Currently when 'save-models-to-hdfs' is false, no models are output at all. The correct behaviour should be to always attempt to output the best model, and instead have 'save-models-to-hdfs' affect whether all models are output or not.

PR #86, opened 2016-05-20
Deprecate 'save-models-to-hdfs'. Add new parameter to control model output.

ashelkovnykov:
Addresses issue #84.

Tested on the LinkedIn internal grid. With 'save-models-to-hdfs' set to false, only the best model was output.

### DIFF
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverGameIntegTest.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverGameIntegTest.scala
@@ -15,19 +15,20 @@
 package com.linkedin.photon.ml.cli.game.training
 
 import collection.JavaConversions._
-import java.nio.file.{Files, FileSystems, Path}
-import org.apache.spark.SparkConf
-import org.testng.annotations.Test
-import org.testng.Assert._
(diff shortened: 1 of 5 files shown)


XianXing:
Hmm, I like the idea to separate the logic of saving *all* models versus saving the *best* model.

But there is one more thing to note: sometimes the main purpose of a modeler to run GAME is to tune different hyper-parameters and try different modeling ideas, in order to determine the best hyper-parameter(s) to use for offline/online scoring and model serving. And in those use cases, the user may or may not want GAME to output any models, even the "best" model, to HDFS. Because sometimes the model can be large and would unnecessarily consume disk space, and the modeler may just want to take a look at the evaluation metric and check her modeling idea and nothing more.

Is there any way that we can separate the saving all models from the saving best model logic, while still  providing the modelers the option of not saving any models?

joshvfleming:
Yeah. This seems like a multi-mode kind of situation. E.g.:

```latex
saveModels \in \{ all, bestOnly, none \}
```

ashelkovnykov:
Going to do another test, will post results here when available.

ashelkovnykov:
Added some integration tests and tested on the LinkedIn internal grid, works fine.

XianXing:
LGTM

Issue #87, opened 2016-05-20
Check whether the output already exists or not before running the driver


XianXing:
Currently both the training and scoring driver for GAME and the training driver for Photon would automatically delete the user specified output directory if it exists already, which is equivalent to setting the force overwrite to always true for the output directory.

Such behavior is inconsistent with both Pig and Spark, as both Pig and Spark by default (e.g., http://spark.apache.org/docs/latest/configuration.html) will throw an exception if the output directory already exists.

Since most of our customers are Pig or Spark users as well, it might be a good idea to keep things consistent and have Photon throw an exception if the specified output directory already exists.

PR #89, opened 2016-05-24
Throw exception if output directory already exists.

XianXing:
Currently Photon/GAME will silently delete the user specified output directory if it exists. However, such behavior is inconsistent with Pig and Spark, where an exception will be thrown if the output directory exists.

More details on this PR are discussed in #87.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
@@ -37,7 +37,7 @@ import com.linkedin.photon.ml.supervised.TaskType._
 import com.linkedin.photon.ml.supervised.classification.{LogisticRegressionModel, SmoothedHingeLossLinearSVMModel}
 import com.linkedin.photon.ml.supervised.model.{GeneralizedLinearModel, ModelTracker}
 import com.linkedin.photon.ml.supervised.regression.{LinearRegressionModel, PoissonRegressionModel}
-import com.linkedin.photon.ml.util.{PalDBIndexMapLoader, PhotonLogger, Utils}
+import com.linkedin.photon.ml.util.{IOUtils, PalDBIndexMapLoader, PhotonLogger, Utils}
 import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.spark.rdd.RDD
@@ -144,7 +144,6 @@ protected[ml] class Driver(
 
     Utils.createHDFSDir(params.outputDir, sc.hadoopConfiguration)
     val finalModelsDir = new Path(params.outputDir, LEARNED_MODELS_TEXT).toString
-    Utils.deleteHDFSDir(finalModelsDir, sc.hadoopConfiguration)
     suite.writeModelsInText(sc, lambdaModelTuples, finalModelsDir.toString)
 
     logger.info(s"Final models are written to: $finalModelsDir")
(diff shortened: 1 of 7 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
```diff
@@ -517,7 +514,16 @@ object Driver {
     val params = PhotonMLCmdLineParser.parseFromCommandLine(args)
 
     /* Configure the Spark application and initialize SparkContext, which is the entry point of a Spark application */
-    val sc: SparkContext = SparkContextConfiguration.asYarnClient(new SparkConf(), params.jobName, params.kryo)
+    val sc = SparkContextConfiguration.asYarnClient(new SparkConf(), params.jobName, params.kryo)
+    val configuration = sc.hadoopConfiguration
+
+    assert(!IOUtils.isDirExisting(params.outputDir, configuration),
+      s"Output directory ${params.outputDir} already exists!" )
```
joshvfleming: This feels more like a ```require``` to me. The difference between ```assert``` and ```require``` is subtle, but I'd characterize it like this:

**assert** - ensure that we are in a sane state as it relates to internal code consistency (i.e. in cases where a failure might be the code's fault)

**require** - ensure that we can make certain assumptions about user-supplied information (i.e. in cases where a failure might be the user's fault)

The other distinction is that ```assert``` is elidable, meaning that assertions can be disabled during build.

photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
```diff
@@ -517,7 +514,16 @@ object Driver {
...
+
+    params.summarizationOutputDirOpt.foreach { dir =>
+      assert(!IOUtils.isDirExisting(dir, configuration),
+        s"Summarization output directory $dir already exists!" )
```
joshvfleming: Same here.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/NameAndTermFeatureSetContainer.scala
```diff
@@ -168,6 +168,9 @@ object NameAndTermFeatureSetContainer {
 
     println(params + "\n")
     val sparkContext = SparkContextConfiguration.asYarnClient(applicationName, useKryo = true)
+    val configuration = sparkContext.hadoopConfiguration
+    assert(!IOUtils.isDirExisting(params.featureNameAndTermSetOutputPath, configuration),
+      s"Output path ${params.featureNameAndTermSetOutputPath} already exists!" )
```
joshvfleming: And here.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -513,6 +511,9 @@ object Driver {
     import params._
 
     val sc = SparkContextConfiguration.asYarnClient(applicationName, useKryo = true)
+    val configuration = sc.hadoopConfiguration
+
+    assert(!IOUtils.isDirExisting(outputDir, configuration), s"Output directory $outputDir already exists!" )
```
joshvfleming: Here too.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -21,16 +21,34 @@ import scala.collection.mutable
...
 
 /**
  * Some basic IO util functions to be merged with the other util functions
- * @todo merge this class with the IOUtil function in Photon
- * @author xazhang
  */
 protected[ml] object IOUtils {
 
+  /**
+    * Delete a given directory on HDFS, it will be silent if the directory does not exist
```
ashelkovnykov: Comment doesn't line up with method

photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
XianXing: Good to know, thanks!

PR #90, opened 2016-05-24
Fix skipped test

joshvfleming:
TestNG was skipping this test because the DataProvider failed to create a null date range.

This PR also includes a change that would prevent this from happening in the future. The idea is to globally transform all tests that are skipped on account of a Throwable into failures.

### DIFF
--- a/build-scripts/integration-test.gradle
+++ b/build-scripts/integration-test.gradle
@@ -57,6 +57,7 @@ plugins.withType(JavaPlugin) {
     options {
       parallel = 'methods'
       threadCount = 4
+      listeners << 'com.linkedin.photon.ml.test.FailOnSkipListener'
     }
 
     afterSuite { desc, result ->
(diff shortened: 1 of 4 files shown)


ashelkovnykov:
LGTM

build-scripts/integration-test.gradle
```diff
@@ -57,6 +57,7 @@ plugins.withType(JavaPlugin) {
     options {
       parallel = 'methods'
       threadCount = 4
+      listeners << 'com.linkedin.photon.ml.test.FailOnSkipListener'
```
XianXing: Nice!

photon-ml/src/integTest/scala/com/linkedin/photon/ml/util/IOUtilsIntegTest.scala
```diff
@@ -48,7 +48,7 @@ class IOUtilsIntegTest extends SparkTestUtils with TestTemplateWithTmpDir {
       Array(baseDir.toString, DateRange.fromDates("20160101-20160401"), Seq(path1, path2, path3)),
       Array(baseDir.toString, DateRange.fromDates("20160101-20160301"), Seq(path1, path2, path3)),
       Array(baseDir.toString, DateRange.fromDates("20160101-20160201"), Seq(path1, path2)),
-      Array(baseDir.toString, DateRange.fromDates("20160101-20160101"), Seq(path1)),
+      Array(baseDir.toString, DateRange.fromDates("20160101-20160102"), Seq(path1)),
```
XianXing: hmm, we still have some use cases out there that require the start date to be the same as the end date (e.g., in some weekly/monthly data used by feed and jymbii). It looks like the problem is caused by the ```require``` statement in ```DateRange```.

But I will open an issue and address this problem in a separate PR.

Issue #91, opened 2016-05-24
Allow the start date to be the same as end date when the user specify date range.


XianXing:
Currently in the ```DateRange``` class it is required that the ```startDate``` always comes strictly before ```endDate```.

However, we still have some use cases out there that require the start date to be the same as the end date (e.g., in some weekly/monthly data used by feed and jymbii).

PR #92, opened 2016-05-24
Allow the start date and end date to be configured as the same day

XianXing:
As discussed in issue #91, it is useful to allow the start date and end date to be configured as the same day in certain use cases.


joshvfleming:
LGTM. Will merge after test run.

Issue #93, opened 2016-05-25
GAME's model selection logic is wrong


XianXing:
Current GAME decides the "best" model by looking at the evaluation metrics computed on the validating data with each model, then select the one with **largest** value.

Such logic works fine with evaluation metrics like AUC, however, for other evaluation metrics such as RMSE, larger value actually corresponds to worse model behavior.

As a result, GAME is now actually reporting the model with highest RMSE as the best model when the task type is specified to be ```LINEAR_REGRESSION```.

This is a bug and should be fixed.

PR #96, opened 2016-05-26
Fix GAME model selection logic bug

joshvfleming:
This addresses issue https://github.com/linkedin/photon-ml/issues/93

Previously we selected the "best" model by largest score value. This is
wrong for some metrics, however, such as RMSE where larger values
indicated greater error.

The fix is a change to the Evaluator interface that allows evaluator
implementations to specify how scores should be compared.


ashelkovnykov:
LGTM

XianXing:
LGTM, thanks for fixing the bug!

PR #97, opened 2016-05-28
Improves GAME's scoring module, mostly from the following two perspec…

XianXing:
…tives: 1. Separate the logic of scoring and model evaluation 2. Created a more user-friendly output of the scoring job. Also improved test coverage for scoring related components.

More details can be found in issue #71.

PR #102, opened 2016-05-31
Minor doc fix for PR#100

XianXing:
1. Adding the format example for ```featureShardIdToInterceptMap```.
2. Reformat the code.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
@@ -37,6 +37,10 @@ import scopt.OptionParser
  * @param featureShardIdToFeatureSectionKeysMap A map between the feature shard id and it's corresponding feature
  *                                              section keys in the following format:
  *                                              shardId1:sectionKey1,sectionKey2|shardId2:sectionKey2,sectionKey3.
+ * @param featureShardIdToInterceptMap A map between the feature shard id and a boolean variable that decides whether a
+ *                                     dummy feature should be added to the corresponding shard in order to learn an
+ *                                     intercept, for example, in the following format: shardId1:true|shardId2:false.
+ *                                     The default is true for all shard ids.
  * @param randomEffectIdSet A set of random effect ids of the corresponding random effect models in the following
  *                          format: randomEffectId1,randomEffectId2,randomEffectId3,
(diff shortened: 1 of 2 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
```diff
@@ -100,7 +104,6 @@ object Params {
               s"are assumed to be flat directories of input files (e.g., inputDir/input-data-files). " +
               s"Default: ${defaultParams.dateRangeDaysAgoOpt}.")
           .action((x, c) => c.copy(dateRangeDaysAgoOpt = Some(x)))
-        .action((x, c) => c.copy(dateRangeOpt = Some(x)))
```
XianXing: Here is a bug introduced in PR #97.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
```diff
@@ -82,10 +86,10 @@ object Params {
...
     val parser = new OptionParser[Params]("GLMix-Scoring-Params") {
       opt[String]("input-data-dirs")
-        .required()
-        .text("Input directories of data to be scored. Multiple input directories are also accepted if they are " +
-          "separated by commas, e.g., inputDir1,inputDir2,inputDir3.")
-        .action((x, c) => c.copy(inputDirs = x.split(",")))
+          .required()
+          .text("Input directories of data to be scored. Multiple input directories are also accepted if they are " +
+              "separated by commas, e.g., inputDir1,inputDir2,inputDir3.")
+          .action((x, c) => c.copy(inputDirs = x.split(",")))
```
joshvfleming: @XianXing I can't find anything about this in the style guide, but I believe the previous indentation style (2-space) is the generally accepted way when you have a multi-line method chain like this.

XianXing: Thanks for pointing it out. 

The continuation indent could be confusing sometimes, for example, for method/class's parameters the continuation indent is recommended to be 4-space (https://github.com/databricks/scala-style-guide#indent), while it looks like for most other cases it should be 2-space, as you suggested. But unfortunately Intellij doesn't allow us to set the continuation indent case by case, so we need to pay extra attention to it each time.

joshvfleming:
LGTM

PR #103, opened 2016-06-02
Fix the bugs identified through Photon model training logs

XianXing:
When investigating some of the issues that our Photon ML customers encountered recently, I found the following logs from the model training job:

number of training data points: 25250560, number of features in **each** training example including intercept (if any) 3479.
Convergence reason: objective is not improving
      Iter   Time(s)                    Value     |Gradient|
         0     1.607        37800195.11474618       1.18e+07
         1    21.362        35780206.30930489       1.52e+07
         2    28.717        35490290.19306723       1.95e+06
         **3    50.755        35224190.91374312       1.94e+06
         3    69.871        35224190.91374312       1.94e+06**

Highlighted are two problems I found:
The first is the misleading statement on the feature size, what we really want to inform the user is that "**total** number of **unique** features found in training data including intercept", not "number of features in **each** training example including intercept".

The second is that, a duplicated optimization state is always tracked at the last iteration, if the convergence reason is "objective is not improving".

And this PR addresses the above two issues.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/optimization/AbstractOptimizer.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/optimization/AbstractOptimizer.scala
@@ -98,15 +98,21 @@ abstract class AbstractOptimizer[Datum <: DataPoint, -Function <: DiffFunction[D
   def getCurrentState: Option[OptimizerState] = currentState
 
   protected def setCurrentState(state: Option[OptimizerState]): Unit = {
-    currentState = state
-    state match {
-      case Some(x) =>
-        statesTracker match {
-          case Some(y) =>
-            y.track(x)
-            y.convergenceReason = convergenceReason
-          case None =>
+    (state, statesTracker, currentState) match {
+      case (Some(s), Some(st), Some(ct)) =>
+        // Only tracks the state if it is different from the previous state (e.g., different objects)
+        if (s != ct) {
+          st.track(s)
         }
+      case _ =>
+    }
+    currentState = state
+  }
+
+  protected def setConvergenceReason(): Unit = {
+    // Set the convergenceReason when the optimization is done
+    statesTracker match {
+      case Some(y) => y.convergenceReason = convergenceReason
       case None =>
     }
   }
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/Driver.scala
@@ -188,8 +188,8 @@ protected[ml] class Driver(
     featureNum = trainingData.first().features.size
 
     // Print out the basic statistics of the training data
-    logger.info(s"number of training data points: $trainingDataNum, " +
-      s"number of features in each training example including intercept (if any) $featureNum.")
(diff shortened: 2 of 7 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/optimization/AbstractOptimizer.scala
```diff
@@ -44,7 +44,17 @@ abstract class AbstractOptimizer[Datum <: DataPoint, -Function <: DiffFunction[D
...
+    val isDone = convergenceReason.nonEmpty
+    if (isDone) {
+      // Set the convergenceReason when the optimization is done
+      statesTracker match {
+        case Some(y) => y.convergenceReason = convergenceReason
+        case None =>
+      }
+    }
+    isDone
+  }
```
joshvfleming: This is semantically a little weird. The name ```isDone``` suggests a simple predicate function with no side-effects, but this change introduces the side-effect of changing the state of ```stateTracker```. Maybe the mutation should be in ```setCurrentState```, but with a check for ```isDone``` as above?

photon-ml/src/main/scala/com/linkedin/photon/ml/optimization/AbstractOptimizer.scala
```diff
@@ -98,17 +108,19 @@ abstract class AbstractOptimizer[Datum <: DataPoint, -Function <: DiffFunction[D
...
-            y.convergenceReason = convergenceReason
+            // Only tracks the state if it is different from the previous state (e.g., different objects)
+            if (state != currentState) {
+              y.track(x)
+            }
           case None =>
         }
       case None =>
     }
+    currentState = state
```
joshvfleming: Consider collapsing this into one pattern match, e.g.:

```scala
(state, statesTracker) match {
  case (Some(s), Some(st)) =>
    // Only tracks the state if it is different from the previous state (e.g., different objects)
    if (s != currentState) {
      st.track(s)
    }
  case _ =>
}
```

ashelkovnykov:
LGTM (aside from Josh's comments)

photon-ml/src/main/scala/com/linkedin/photon/ml/optimization/AbstractOptimizer.scala
XianXing: Good call!

photon-ml/src/main/scala/com/linkedin/photon/ml/optimization/AbstractOptimizer.scala
XianXing: Good suggestion.

I put this piece of code outside of ```setCurrentState``` because ```isDone``` depends on the state of ```previousState```, and coupling ```previousState``` with ```setCurrentState``` might be tricky because the ```Optimizer``` class can change the state of ```previousState``` without informing ```currentState```.

Let me see what I can do to simply the logic a little bit without introducing additional side effect.

XianXing: Separated set convergence reason from set current/previous states logic.

For now we can't really make too much code change for ```AbstractOptimizer``` because the unit tests are not there, so I opened #104 to remind us this issue.

Issue #76, opened 2016-05-16
Getting started documentation


sbourke:
Hi Guys - nice job. 

I think the project would benefit from a simple walk through on some public data. I'm just digging into photon now and it takes a little digging to understand the flow correctly.

XianXing:
Thanks for the feedback, we agree and will work on that. Please stay tuned.

XianXing:
Photon-ML tutorial added to the wiki page.

PR #105, opened 2016-06-03
Add training walkthough on public dataset to README

joshvfleming:
This addresses issue https://github.com/linkedin/photon-ml/issues/76

I've tested this flow on an OS X machine and a Linux machine (Ubuntu 16.04).

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/diagnostics/hl/PredictedProbabilityVersusObservedFrequencyHistogramBin.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/diagnostics/hl/PredictedProbabilityVersusObservedFrequencyHistogramBin.scala
@@ -14,8 +14,6 @@
  */
 package com.linkedin.photon.ml.diagnostics.hl
 
-import com.google.common.base.Preconditions
-
 /**
  * Model an individual histogram bin for purposes of HL tests. Probably want to think about several things:
  * <ul>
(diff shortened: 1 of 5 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/diagnostics/hl/PredictedProbabilityVersusObservedFrequencyHistogramBin.scala
```diff
@@ -44,19 +42,19 @@ class PredictedProbabilityVersusObservedFrequencyHistogramBin(
...
-  Preconditions.checkArgument(lowerBound >= 0 && lowerBound <= 1.0)
-  Preconditions.checkArgument(upperBound >= 0 && upperBound <= 1.0)
-  Preconditions.checkArgument(lowerBound < upperBound)
-  Preconditions.checkArgument(observedNegCount >= 0L)
-  Preconditions.checkArgument(observedPosCount >= 0L)
+  require(lowerBound >= 0 && lowerBound <= 1.0)
+  require(upperBound >= 0 && upperBound <= 1.0)
+  require(lowerBound < upperBound)
+  require(observedNegCount >= 0L)
+  require(observedPosCount >= 0L)
```
ashelkovnykov: I recall that earlier you described the difference between assert and require as (paraphrasing) "require() when arguments are provided by the user, assert() when they're provided from elsewhere in the code". With that in mind, shouldn't these be asserts?

joshvfleming: Nope, I don't think so. A "user" in this sense is a more general user or consumer of the API this class exposes. This is true for almost any case where you do bounds checking for arguments -- you'd never want to elide these checks during build, because it would allow a consumer to call the function (or constructor in this case) with invalid arguments.

ashelkovnykov:
LGTM

XianXing:
LGTM

PR #107, opened 2016-06-06
Improve PhotonLogger stability when jobs are running for hours

timyitong:
Currently PhotonLogger is maintaining a remote output stream with HDFS directly,
the connection could be easily interrupted due to cluster/network instability if the job
is running for a long time. e.g. hours. In my env, this could happen occasionally. Though not a bit issue, but when it happens that means a heavy job just failed and I would waste hours waiting.

Therefore, I'm suggesting change the the behavior by firstly writing the logs onto a local tmp file and then finally copy it from local to HDFS when the logger is closed.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/util/PhotonLogger.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/util/PhotonLogger.scala
@@ -14,6 +14,8 @@
  */
 package com.linkedin.photon.ml.util
 
+import org.apache.commons.io.FileUtils
+import org.apache.hadoop.conf.Configuration
 import org.apache.hadoop.fs.{FileSystem, Path}
 import org.apache.spark.SparkContext
 import org.slf4j.helpers.{MarkerIgnoringBase, MessageFormatter}
@@ -22,6 +24,9 @@ import org.slf4j.spi.LocationAwareLogger
 import java.io.{BufferedWriter, OutputStreamWriter, PrintWriter}
 import java.text.SimpleDateFormat
 import java.util.Date
+import java.io._
+import java.net.URI
+import java.util.UUID
 
 import scala.Predef.{println => sprintln}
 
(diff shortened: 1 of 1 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/util/PhotonLogger.scala
```diff
@@ -63,6 +73,23 @@ protected[ml] class PhotonLogger(logPath: Path, sc: SparkContext) extends Marker
...
+
+    // Overwrite any existing file
+    // TODO: make this configurable?
+    if (fs.exists(logPath)) {
+      fs.delete(logPath, false)
+    }
+
+    fs.copyFromLocalFile(tmpLocalPath, logPath)
+
+    new File(tmpLocalPath.toString()).delete()
```
joshvfleming: Hi @timyitong, good to run into you again!

If ```fs.copyFromLocalFile``` fails with ```IOException```, we could end up with a dangling temporary file here. We should probably have this in a ```finally``` block.

timyitong: Hi @joshvfleming , how are things going?

This is a very good catch, I'll address it : )

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PhotonLogger.scala
```diff
@@ -62,7 +72,28 @@ protected[ml] class PhotonLogger(logPath: Path, sc: SparkContext) extends Marker
...
+      // Create the parent directory if it doesn't exist already
+      if (!fs.exists(parent)) {
+        fs.mkdirs(parent)
+      }
+
+      // Overwrite any existing file
+      // TODO: make this configurable?
+      if (fs.exists(logPath)) {
+        fs.delete(logPath, false)
+      }
```
XianXing: As in #89, currently the logic in Photon is that if the output directory of Photon already exists, an error will be thrown. Since the ```logPath``` is always a child of Photon's output directory, in order to be consistent with the rest of code I guess we may not need the check and overwriting here.

Actually I forgot to remove this logic in PR#89 :)

XianXing:
LGTM, I agree this is a very good idea to have a local tmp file to keep the logs.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PhotonLogger.scala
XianXing: Let me fix this problem later.

PR #108, opened 2016-06-06
Photon's GLM should not filter out data point with empty feature bag.

XianXing:
A quick fix for the issue reported by Photon users at LinkedIn. 

Problem description:
When the input data are represented in the Avro format, data point with empty feature bag will not be used for model training.

And such logic is problematic, for example, if a data set used for model training only contains one binary feature, with each data point that contains such feature being treated as feature with value 1 and data point that doesn't contain such feature (missing feature) being treated as feature with value 0, the logic described above will throw away all the data points with missing feature and only data points with feature value 1 will kept for model training. As a result, the learned model will be wrong.


XianXing:
@ltangt FYI.

joshvfleming:
LGTM. I think we should also have a unit test that verifies this behavior.

PR #110, opened 2016-06-07
Fix the bug where data point with empty feature bag being thrown away in Photon's GLMSuite

XianXing:
A quick fix for the issue reported by Photon users at LinkedIn.

Problem description:
When the input data are represented in the Avro format, data point with empty feature bag will not be used for model training.

And such logic is problematic, for example, if a data set used for model training only contains one binary feature, with each data point that contains such feature being treated as feature with value 1 and data point that doesn't contain such feature (missing feature) being treated as feature with value 0, the logic described above will throw away all the data points with missing feature and only data points with feature value 1 will kept for model training. As a result, the learned model will be wrong.

Integration tests updated to cover the fixed bug.


joshvfleming:
LGTM

PR #121, opened 2016-06-18
Allow the user to specify modelId with GAME's scoring module

XianXing:
```ScoringResultAvro```, which is the output schema of the computed scores of GAME's scoring module, contains a required field ```modelId```.

This PR allows the user to specify modelId that is used to populate the ```modelId``` field in ```ScoringResultAvro```. This PR relates to issue #71 as well.

### CODE REVIEW

File: photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala [hunk 1]
```diff
@@ -178,12 +178,12 @@ class Driver(val params: Params, val sparkContext: SparkContext, val logger: Pho
     logger.info(s"Number of scored items: $numScoredItems (s)\n")
     val scoredItemsToBeSaved =
       if (numOutputFilesForScores > 0 && numOutputFilesForScores != scoredItems.partitions.length) {
-        scoredItems.coalesce(numOutputFilesForScores)
+        scoredItems.repartition(numOutputFilesForScores)
```
ashelkovnykov: Why do we want repartition instead of coalesce in this case? Should coalesce in the training Driver also be changed to repartition?

photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/scoring/DriverTest.scala
```diff
@@ -70,24 +100,42 @@ object DriverTest {
...
-    val featureShardIdToFeatureSectionKeysMap =
-      "globalShard:features,songFeatures,userFeatures|userShard:features,songFeatures|songShard:features,userFeatures"
-    val randomEffectIdSet = "userId,songId"
-    val modelDir = new Path(inputRoot, "gameModel").toString
-    val numExecutors = "1"
+
+    val (featureShardIdToFeatureSectionKeysMap, randomEffectIdSet, modelDir) =
+      if (fixedEffectOnly) {
+        val featureMap = "globalShard:features,songFeatures,userFeatures"
+        val idSet = ","
```
ashelkovnykov: Does an empty random effect ID set require a comma?

ashelkovnykov:
LGTM, aside from above comments

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
XianXing: Because ```repartition``` will give us exactly the number of partitions specified, while coalesce may not. At the same time, ```repartition``` is easier to test, but has may have larger overhead (extra shuffling) compared to ```coalesce```. 

Since the size of scores here is much smaller than the size of data that need to be "coalesced" in the training driver, and given the tradeoff above, we may want to have ```repartition``` in the scoring driver but ```coalesce``` in the training driver.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/scoring/DriverTest.scala
XianXing: Based on the current implementation in ```scoring.Params```, yes.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
ashelkovnykov: Right, I'm curious as to whether there is a case in which we want to increase the number of output files produced by the scorer.

XianXing: Currently I don't see there would be a case like that for neither training nor scoring driver. And as you may already see from above comments, choosing ```repartition``` here for scoring driver is just because of that it gives us exactly the number of partitions specified by the user and thus is easier to test and understand by the user, and at the same time without introducing too much overhead compared to ```coalesce``` (which doesn't always give us the specified number of partitions even if the user wants to decrease the number of output files).

PR #125, opened 2016-06-22
Remove extra calculation of feature number

joshvfleming:
The same line appears 5 lines below here. This duplicate was causing a
bug where we call "trainingData.first()" before checking whether
"trainingData" is empty.


namitk:
LGTM. Will merge after travis checks

PR #126, opened 2016-06-24
Allow PhotonML to accept LibSVM file format in addition to Avro

timyitong:
Introducing ```InputDataFormat``` trait to ease the data format assumption of PhotonML.

After this PR, supposedly we should be able to load LibSVM for
training/validation and also bring in the potential of supporting any other
arbitrary format.

I'm introducing this core trait to allow different file formats:
```
trait InputDataFormat {

  /**
    * Load an RDD of LabeledPoints
    *
    * @param sc
    * @param inputPath
    * @param selectedFeaturesPath
    * @param minPartitions
    * @return an RDD of LabeledPoints
    */
  def loadLabeledPoints(sc: SparkContext,
                        inputPath: String,
                        selectedFeaturesPath: Option[String],
                        minPartitions: Int): RDD[LabeledPoint]

  /**
    * Provide an IndexMapLoader for the current format
    *
    * @return IndexMapLoader
    */
  def indexMapLoader(): IndexMapLoader

  def constrainFeatureMap(): Option[Map[Int, (Double, Double)]] = None
}
```

Reasoning here is that given an input dataset, the indexMap choice supposedly should also be closely coupled with the format (because we might either need to precompute a map (default ) or load an already-computed one (offheap) or totally don't need any (libsvm). Thus, I feel it's okay to bind ```indexMapLoader``` method and ```loadLabledPoints``` together.

However, I'm a bit concerned with the ```constrainFeatureMap``` mention here as the only reason I added it is I figure right now it's the easily way to integrate everything together without hugely rewrite ```GLMSuite```. Maybe @namitk could suggest how we could improve the binding here. Potentially something like a Feature context map, providing all meta info needed for one feature?

Issues related:
I'm doing a minimal implementation, currently it only supports binary classification case for simple LibSVM data (not considering id or weight or offset). Added integTest in DriverIntegTest and verified that the new settings work as expected and also the old setting won't have any issue.

I tried to do a minimal refactoring possible for Driver. The main goal is to
decouple the linkage between GLMSuite and core Driver functions; so that now
Driver only cares about an InputDataFormat implementation which will provide
loaded label points and index map if asked, but it won't really bother asking
the details of the implementation.

Ideally, some functionalities in GLMSuite should be further decoupled and
refactored to better locations. But to limit the scope, we are not addressing
those issues here.

@XianXing @joshvfleming @namitk


timyitong:
Looks like test JVM got OOM:
```
    2016-06-24T22:35:18.867+0000 [DEBUG] Input records paths:
    /home/travis/build/linkedin/photon-ml/build/photon-ml_2.10/resources/integTest/GameIntegTest/input/test-with-uid
    2016-06-24T22:35:19.390+0000 [DEBUG] Summary for the GAME data set
    2016-06-24T22:35:21.642+0000 [DEBUG] numSamples: 9195
    2016-06-24T22:35:21.932+0000 [DEBUG] numSamples for userId: (count: 9175, mean: 1.002180, stdev: 0.046638, max: 2.000000, min: 1.000000)
    2016-06-24T22:35:22.164+0000 [DEBUG] numSamples for songId: (count: 6987, mean: 1.316015, stdev: 0.880610, max: 14.000000, min: 1.000000)
    2016-06-24T22:35:22.165+0000 [INFO] Time elapsed after game data set preparation: 3.2973623770000002 (s)

Java HotSpot(TM) 64-Bit Server VM warning: INFO: os::commit_memory(0x000000064d080000, 1118306304, 0) failed; error='Cannot allocate memory' (errno=12)
#
# There is insufficient memory for the Java Runtime Environment to continue.
# Native memory allocation (mmap) failed to map 1118306304 bytes for committing reserved memory.
# An error report file with more information is saved as:
# /home/travis/build/linkedin/photon-ml/photon-ml/hs_err_pid4708.log
:photon-ml_2.10 -- Executed 0 tests: 0 succeeded, 0 failed, 0 skipped
:photon-ml_2.10 -- Executed 0 tests: 0 succeeded, 0 failed, 0 skipped
:photon-ml_2.10:integTest FAILED

FAILURE: Build failed with an exception.
```

XianXing:
The overall logic of this PR looks good to me.

@ashelkovnykov also has a plan separate the Photon core code from Photon client code, which seems to be in line with this PR. @ashelkovnykov, please let us know if you have any suggestions on the issues @timyitong discussed above.

One more thing that might not be related to this PR, but do you have a working example/script on how to run Photon's driver with data in LibSVM data format, like this [one](https://github.com/linkedin/photon-ml/blob/master/examples/run_photon_ml_driver.sh) you provided for Avro?

After this PR is merged, we may want to update ```README.md``` accordingly.

But overall this is very nice work!

photon-ml/src/integTest/scala/com/linkedin/photon/ml/DriverIntegTest.scala
```diff
@@ -73,6 +73,83 @@ class DriverIntegTest extends SparkTestUtils with TestTemplateWithTmpDir {
...
+    val models = loadAllModels(new Path(outputDir, Driver.LEARNED_MODELS_TEXT).toString)
+    assertEquals(models.length, defaultParams.regularizationWeights.length)
+    // Verify lambdas
+    assertEquals(models.map(_._1), defaultParams.regularizationWeights.toArray)
+
+    // The selected best model is supposed to be of lambda 100.0 with features scaling with standard deviation
+    val bestModel = loadAllModels(new Path(outputDir, Driver.BEST_MODEL_TEXT).toString)
+    assertEquals(bestModel.length, 1)
+    // Verify lambda
+    assertEquals(bestModel(0)._1, 10, MathConst.HIGH_PRECISION_TOLERANCE_THRESHOLD)
```
ashelkovnykov: Comment says best model has lambda 100, but test checks for lambda=10

photon-ml/src/main/scala/com/linkedin/photon/ml/Params.scala
```diff
@@ -158,6 +160,9 @@ class Params {
    */
   var deleteOutputDirsIfExist: Boolean = false
 
+  var inputFormatType: InputFormatType = InputFormatType.AVRO
+  var featureDimension: Int = -1
```
ashelkovnykov: Please add some short documentation for these two

photon-ml/src/main/scala/com/linkedin/photon/ml/PhotonMLCmdLineParser.scala
```diff
@@ -224,6 +224,12 @@ object PhotonMLCmdLineParser {
         .text(s"Delete the output directories (including the model and summarization output directories) if exist." +
             s"Default: ${defaultParams.deleteOutputDirsIfExist}")
         .foreach(x => params.deleteOutputDirsIfExist = x)
+      opt[String](INPUT_FILE_FORMAT)
+        .text("Indicating the input data format for PhotonML")
+        .foreach(x => params.inputFormatType = InputFormatType.withName(x.toUpperCase()))
+      opt[Int](FEATURE_DIMENSION)
+        .text("A preliminary indicator of how many features the input data contains (just an upper bound, not including intercept).")
+        .foreach(x => params.featureDimension = x)
```
ashelkovnykov: Should make note that feature dimension is only used for LibSVM input

photon-ml/src/main/scala/com/linkedin/photon/ml/io/InputFormatType.scala
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.io
+
+/**
+  * @author yitong_zhou
+  */
```
ashelkovnykov: Please add some short documentation about this Enum (I know it's pretty obvious from the name but we should be consistent)
We've also decided to get rid of author tags (it's happening slowly as files that still contain them get touched)

photon-ml/src/main/scala/com/linkedin/photon/ml/io/LibSVMInputDataFormat.scala
```diff
@@ -0,0 +1,75 @@
...
+  *
+  * By default, libsvm feature indices start from 1, but if zeroBased is true, we will assume it
+  * starts from 0.
+  */
+class LibSVMInputDataFormat(val featureDimension: Int,
+                            val useIntercept: Boolean = true,
+                            val zeroBased: Boolean = false,
+                            val delim: String = " ",
+                            val idxValueDelim: String = ":"
+                           ) extends InputDataFormat {
```
ashelkovnykov: Please make sure that changes follow the Scala style guide:
http://docs.scala-lang.org/style/declarations.html#classes

photon-ml/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -185,4 +196,134 @@ protected[ml] object IOUtils {
...
+  }
+
+  private case class BasicSummaryItems(
+                                        max: Double,
+                                        min: Double,
+                                        mean: Double,
+                                        normL1: Double,
+                                        normL2: Double,
+                                        numNonzeros: Double,
+                                        variance: Double)
```
ashelkovnykov: Scala style

photon-ml/src/main/scala/com/linkedin/photon/ml/util/IdentityIndexMapLoader.scala
```diff
@@ -0,0 +1,66 @@
...
+
+import com.linkedin.photon.ml.Params
+import com.linkedin.photon.ml.io.GLMSuite
+import org.apache.spark.SparkContext
+
+/**
+  * A wrapper that loads a pseudo index map that always directly passes through indices other than
+  * intercept.
+  *
+  * @author yitong_zhou
```
ashelkovnykov: Same as above

photon-ml/src/main/scala/com/linkedin/photon/ml/io/AvroInputDataFormat.scala
```diff
@@ -0,0 +1,37 @@
...
+  override def loadLabeledPoints(sc: SparkContext,
+                                 inputPath: String,
+                                 selectedFeaturesPath: Option[String],
+                                 minPartitions: Int): RDD[LabeledPoint] = {
+    suite.readLabeledPointsFromAvro(sc, inputPath, selectedFeaturesPath, minPartitions)
+  }
+
+  override def indexMapLoader(): IndexMapLoader = suite.indexMapLoader()
+
+  override def constrainFeatureMap(): Option[Map[Int, (Double, Double)]] = suite.constraintFeatureMap
```
ashelkovnykov: Should this be `constraintFeatureMap`?

ashelkovnykov:
@XianXing That hasn't gotten off the ground yet, we'll need to discuss if and how we want to do that as part of the Q3 tasks

@timyitong Looks good to me aside from what nitpicks I could find

photon-ml/src/main/scala/com/linkedin/photon/ml/Params.scala
XianXing: We need ```featureDimension``` here because it is a required parameter to construct a sparse vector of type ```breeze.linalg.SparseVector```.

However, I would prefer us not exposing such implementation details to the users. Is it possible to let the index map loaders to figure out the total number of features instead of asking the users to specify it explicitly? 

For example, instead of having ```featureDimension``` to be a constructor parameter of ```IdentityIndexMapLoader```, maybe we can have factory method similar to ```createDefaultIndexMapLoader``` that figures out the number of features in the data set automatically.

timyitong: It's one part of the reason that we need to indicate feature dimension when constructing `SparseVector`. But it's also because presumably LibSVM format is already a pre-indexed format that I feel it's a bit redundant to do second round indexing just to learn the feature dimension. (Although it might make some sense in some cases when the original indexing is not dense enough). So I'm inclined to keep it simple for now but later add in other options more properly if needed. But we do need to be specific that `featureDimension` is a parameter only used by LibSVM format. What do you think?

XianXing: Okay, sounds reasonable. 

I was suggesting to count the number of unique features in ```IdentityIndexMapLoader``` when creating the map loader, which should be a computationally light job, but *not* to actually re-index the features as other map loaders do, because it's redundant for most LibSVM data set as you suggested. And I just want to be very cautious when introducing any user-facing API changes 

Please also note that ```featureDimension``` the user specified will also be used to decide the dimension of the dense vectors used in optimization (e.g., coefficients vector, gradient, hessian vector), so if the user specify an upper bound that is (much) larger than the actual number of unique features, Photon's performance may suffer.

XianXing:
@timyitong Can you take a look at those minor issues Alex suggested? I will merge this PR afterwards.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/DriverIntegTest.scala
```diff
@@ -865,13 +942,17 @@ object DriverIntegTest {
 
   val TEST_DIR = ClassLoader.getSystemResource("DriverIntegTest").getPath
 
-  def appendCommonJobArgs(args: mutable.ArrayBuffer[String], outputDir: String, isValidating: Boolean = false): Unit = {
+  def appendCommonJobArgs(args: mutable.ArrayBuffer[String],
+                          outputDir: String,
+                          isValidating: Boolean = false,
+                          fileSuffix: String = ".avro"
+                         ): Unit = {
```
ashelkovnykov: Scala style; though not official Scala standard, we would prefer:
```Scala
def appendCommonJobArgs(
    args: mutable.ArrayBuffer[String],
    outputDir: String,
    isValidating: Boolean = false,
    fileSuffix: String = ".avro"): Unit = {

  ...
}
```

photon-ml/src/main/scala/com/linkedin/photon/ml/io/AvroInputDataFormat.scala
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+  * A general input format supporting reading Avro data
+  */
+class AvroInputDataFormat(val suite: GLMSuite) extends InputDataFormat {
+
+  override def loadLabeledPoints(sc: SparkContext,
+                                 inputPath: String,
+                                 selectedFeaturesPath: Option[String],
+                                 minPartitions: Int): RDD[LabeledPoint] = {
```
ashelkovnykov: Scala style

photon-ml/src/main/scala/com/linkedin/photon/ml/io/InputDataFormat.scala
```diff
@@ -0,0 +1,49 @@
...
+    * @param sc
+    * @param inputPath
+    * @param selectedFeaturesPath
+    * @param minPartitions
+    * @return an RDD of LabeledPoints
+    */
+  def loadLabeledPoints(sc: SparkContext,
+                        inputPath: String,
+                        selectedFeaturesPath: Option[String],
+                        minPartitions: Int): RDD[LabeledPoint]
```
ashelkovnykov: Scala style

photon-ml/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -185,4 +196,134 @@ protected[ml] object IOUtils {
...
+    *
+    * @param sc       The Spark context
+    * @param models   The map of (Model Id -> [[GeneralizedLinearModel]])
+    * @param modelDir The directory for the output text files
+    */
+  def writeModelsInText(sc: SparkContext,
+                        models: Iterable[(Double, GeneralizedLinearModel)],
+                        modelDir: String,
+                        indexMapLoader: IndexMapLoader
+                       ): Unit = {
```
ashelkovnykov: Scala style

photon-ml/src/main/scala/com/linkedin/photon/ml/Params.scala
timyitong: That's a very reasonable argument. How about that I introduce another PR right after this one to address the second-pass feature dimension counting problem? I'm thinking that we could also sort of write out a small ```_meta``` file in the input path if not presented, so that we could only do the extra pass once for the same input data. This could take a few more lines which might be a distraction to review here.

XianXing: Sure, sounds like a plan to me.

photon-ml/src/main/scala/com/linkedin/photon/ml/io/InputFormatType.scala
timyitong: Totally makes sense. I also prefer not putting the author tags, just somehow missed this one : )

timyitong:
@XianXing @ashelkovnykov sorry I did not follow up with thread earlier. I just addressed all the style related issues. Also I'm trying to check in some unit tests for IdentityIndexMapLoader very soon.

ashelkovnykov:
Thanks @timyitong, looks good. We just merged a big PR, hopefully it doesn't cause you too many headaches during merge

timyitong:
@ashelkovnykov : minutes ago I just resolved the rebase conflicts, there aren't many actually; but I'd wait until the integ tests finished.... I only did unit tests locally, which turns out to be fine.

photon-ml/src/test/scala/com/linkedin/photon/ml/util/IdentityIndexMapLoaderTest.scala
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    assertEquals(nameSet.size, 10)
+    assertEquals(idxSet.size, 10)
+    for (i <- 0 until 10) {
+      assertTrue(nameSet.contains(i.toString()))
+      assertTrue(idxSet.contains(i))
+    }
+  }
+}
```
ashelkovnykov: Please add a newline after the last bracket. Also, LGTM, squash the PR into one commit (the changes are small enough that I think it would be best that way) and I'll merge

timyitong: Just did these things. Let's wait till tests finished.

Issue #135, opened 2016-07-01
GAME Params cmd parser is replacing outputDir ':' characters


timyitong:
Things like ```hdfs://foo/bar/tar``` will be replaced into: ```hdfs_/foo/bar/tar```, basically hdfs becomes a local dir.

XianXing:
I will fix it.

PR #136, opened 2016-07-01
Fix the bug where colon is replaced by underscore.

XianXing:
More details can be find in issue #135.

Unit test also added to cover the bug.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Params.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Params.scala
@@ -216,7 +216,7 @@ object Params {
       opt[String]("output-dir")
         .required()
         .text(s"Output directory for logs and learned models.")
-        .foreach(x => params.outputDir = x.replace(',', '_').replace(':', '_'))
+        .foreach(x => params.outputDir = x.replace(',', '_'))
       opt[String]("feature-name-and-term-set-path")
         .required()
         .text(s"Input path to the features name-and-term lists.")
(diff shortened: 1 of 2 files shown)


XianXing:
@timyitong

photon-ml/src/test/scala/com/linkedin/photon/ml/cli/game/training/ParamsTest.scala
```diff
@@ -334,44 +345,48 @@ object ParamsTest {
...
-    args ++ requiredArgs()
+  def setOneMoreArg(argName: String, argValue: String): Map[String, String] = {
+    val name = CommonTestUtils.fromOptionNameToArg(argName)
+    val value = argValue
+    requiredArgs().updated(name, value)
+  }
+
+  implicit def mapToArray(args: Map[String, String]): Array[String] = {
+    args.toArray.flatMap { case (name, value) => Seq(name, value) }
   }
```
XianXing: I usually prefer not to use implicit function in Scala for readability and performance reasons, but the implicit conversion comes pretty handy here for unit tests.

ashelkovnykov:
LGTM

photon-ml/src/test/scala/com/linkedin/photon/ml/cli/game/training/ParamsTest.scala
joshvfleming: Yeah. I agree that this kind of thing is okay for making tests more concise, but could lead to readability and comprehensibility problems in the production code.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Params.scala
```diff
@@ -216,7 +216,7 @@ object Params {
       opt[String]("output-dir")
         .required()
         .text(s"Output directory for logs and learned models.")
-        .foreach(x => params.outputDir = x.replace(',', '_').replace(':', '_'))
+        .foreach(x => params.outputDir = x.replace(',', '_'))
```
joshvfleming: We could make this a little more robust by having it parse URIs first, and then do the replacement -- if the intention is to support URIs as an input here (and if it's important to do the colon replacement).

I don't think we need to do this now though.

timyitong:
@joshvfleming: Thanks so much for fixing this so quickly!

PR #145, opened 2016-07-29
Add evaluators for performing Poisson regression and Smoothed Hinge L…

ashelkovnykov:
…oss Linear SVM training in GAME

This change is necessary to enable Poisson Regression and Smoothed Hinge Loss for GAME.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
@@ -207,6 +207,10 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
         new LogisticLossEvaluator(labelAndOffsetAndWeights)
       case LINEAR_REGRESSION =>
         new SquaredLossEvaluator(labelAndOffsetAndWeights)
+      case POISSON_REGRESSION =>
+        new PoissonLossEvaluator(labelAndOffsetAndWeights)
+      case SMOOTHED_HINGE_LOSS_LINEAR_SVM =>
+        new SmoothedHingeLossEvaluator(labelAndOffsetAndWeights)
       case _ =>
         throw new UnsupportedOperationException(s"Task type: $taskType is not supported to create training evaluator")
     }
@@ -281,19 +285,19 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
       logger.debug(s"responseSumStats for $randomEffectId: $responseSumStats")
     }
 
-    val validatingLabelAndOffsets = gameDataSet.mapValues(gameData => (gameData.response, gameData.offset))
-        .setName(s"Validating labels and offsets").persist(StorageLevel.FREQUENT_REUSE_RDD_STORAGE_LEVEL)
-    validatingLabelAndOffsets.count()
+    val validatingLabelsAndOffsetsAndWeights = gameDataSet
+      .mapValues(gameData => (gameData.response, gameData.offset, gameData.weight))
+      .setName(s"Validating labels and offsets").persist(StorageLevel.FREQUENT_REUSE_RDD_STORAGE_LEVEL)
+    validatingLabelsAndOffsetsAndWeights.count()
 
     val evaluator =
       taskType match {
-        case LOGISTIC_REGRESSION =>
-          new BinaryClassificationEvaluator(validatingLabelAndOffsets)
+        case LOGISTIC_REGRESSION | SMOOTHED_HINGE_LOSS_LINEAR_SVM =>
+          new BinaryClassificationEvaluator(validatingLabelsAndOffsetsAndWeights)
         case LINEAR_REGRESSION =>
-          val validatingLabelAndOffsetAndWeights = validatingLabelAndOffsets.mapValues { case (label, offset) =>
-            (label, offset, 1.0)
-          }
-          new RMSEEvaluator(validatingLabelAndOffsetAndWeights)
+          new RMSEEvaluator(validatingLabelsAndOffsetsAndWeights)
+        case POISSON_REGRESSION =>
+          new PoissonLossEvaluator(validatingLabelsAndOffsetsAndWeights)
         case _ =>
           throw new UnsupportedOperationException(s"Task type: $taskType is not supported to create validating " +
               s"evaluator")
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/LogisticLossEvaluator.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/LogisticLossEvaluator.scala
@@ -16,51 +16,48 @@ package com.linkedin.photon.ml.evaluation
 
 import org.apache.spark.rdd.RDD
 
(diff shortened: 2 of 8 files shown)


XianXing:
Can you update README.md to reflect these changes as well? For example, just mention that we have functionalities like model training with smoothed hinge loss and model training without regularization.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -287,9 +291,9 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
 
     val evaluator =
       taskType match {
-        case LOGISTIC_REGRESSION =>
+        case LOGISTIC_REGRESSION | SMOOTHED_HINGE_LOSS_LINEAR_SVM =>
           new BinaryClassificationEvaluator(validatingLabelAndOffsets)
-        case LINEAR_REGRESSION =>
+        case LINEAR_REGRESSION | POISSON_REGRESSION =>
```
ankans: After a discussion last time, we had decided to use the log likelihood as the best evaluator for Poisson Regression, given that it should always increase on the training data and generally increase on the test data as well. RMSE is clearly a bad metric, since Poisson regression does not try to fit the actual value of the predictor but a complicated function involving exp of the predictor.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/LogisticLossEvaluator.scala
```diff
@@ -19,48 +19,49 @@ import org.apache.spark.rdd.RDD
...
     scoreAndLabelAndWeights.map { case (score, (label, weight)) =>
-      if (label > MathConst.POSITIVE_RESPONSE_THRESHOLD) {
-        weight * Utils.log1pExp(-score)
-      } else {
-        weight * Utils.log1pExp(score)
+        if (label > MathConst.POSITIVE_RESPONSE_THRESHOLD) {
+          weight * Utils.log1pExp(-score)
+        } else {
+          weight * Utils.log1pExp(score)
+        }
```
XianXing: Can we call ```PointwiseLogisticLossFunction.loss``` directly to get the loss function value, instead of re-defining and computing the loss function value here?

The same goes for ```PoissonLossEvaluator```, ```SmoothedHingeLossEvaluator``` and ```SquaredLossEvaluator```.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PoissonLossEvaluator.scala
```diff
@@ -0,0 +1,60 @@
...
+
+    val scoreAndLabelAndWeights = scores
+      .rightOuterJoin(labelAndOffsetAndWeights)
+      .mapValues { case (scoreOption, (label, offset, weight)) =>
+        (scoreOption.getOrElse(defaultScore) + offset, (label, weight))
+      }
+      .values
+
+    scoreAndLabelAndWeights.map { case (score, (label, weight)) =>
+        weight * (math.exp(score) - score * label)
```
XianXing: See comment above.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/SmoothedHingeLossEvaluator.scala
```diff
@@ -0,0 +1,70 @@
...
+
+    scoreAndLabelAndWeights.map { case (score, (label, weight)) =>
+        val z = if (label < MathConst.POSITIVE_RESPONSE_THRESHOLD) -score else score
+
+        if (z <= 0) {
+          weight * (0.5 - z)
+        } else if (z < 1) {
+          weight * 0.5 * (1.0 - z) * (1.0 - z)
+        } else {
+          0
```
XianXing: See comment above, but we may need some special treatment for ```SmoothedHingeLossFunction```.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/SquaredLossEvaluator.scala
```diff
@@ -17,43 +17,45 @@ package com.linkedin.photon.ml.evaluation
...
+        }
+      .values
 
     scoreAndLabelAndWeights.map { case (score, (label, weight)) =>
-      val diff = score - label
-      weight * diff * diff
-    }.reduce(_ + _)
+        val diff = score - label
+        weight * diff * diff
+      }
```
XianXing: Ditto.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ashelkovnykov: Yes, I had asked @ankans but forgotten to make the change, sorry about that

ashelkovnykov:
While making the suggested changes, I've also enabled using validation record weights during evaluation (previously they were being ignored).

photon-ml/src/main/scala/com/linkedin/photon/ml/function/SmoothedHingeLossFunction.scala
```diff
@@ -15,22 +15,42 @@
...
-    val z = actualLabel * margin
+    val (loss, deriv) = SmoothedHingeLossFunction.lossAndDerivative(margin, datum.label)
+
+    // Eq. 5, page 2 (derivative multiplied by label in lossAndDerivative method)
+    breeze.linalg.axpy(datum.weight * deriv, datum.features, cumGradient)
+    datum.weight * loss
+  }
+}
+
+object SmoothedHingeLossFunction {
```
XianXing: Are these changes already covered by existing unit tests?

ashelkovnykov: Yes, these changes just move the calculation out to an object so that it is available for the `SmoothedHingeLossEvaluator`, they shouldn't modify the behaviour. Even so, the behaviour is tested by the `ObjectiveFunctionTest` and `ObjectiveFunctionIntegTest` tests.

XianXing:
LGTM. Will merge if Travis is happy with this PR too.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PoissonLossEvaluator.scala
```diff
@@ -0,0 +1,62 @@
...
+
+    val scoreAndLabelAndWeights = scores
+      .rightOuterJoin(labelAndOffsetAndWeights)
+      .mapValues { case (scoreOption, (label, offset, weight)) =>
+        (scoreOption.getOrElse(defaultScore) + offset, (label, weight))
+      }
+      .values
+
+    scoreAndLabelAndWeights.map { case (score, (label, weight)) =>
+        weight * PointwisePoissonLossFunction.loss(score, label)._1
```
ankans: Are you using the same unit test for PoissonGradient evaluation that Brendan was using? Also, it is mandatory that the users only run Poisson regression with normalized features. Should this be specified as a comment in this file as well?

PR #154, opened 2016-08-22
Fix scalability issue with DefaultIndexMap.size

joshvfleming:
Previously there was no implementation for "DefaultIndexMap.size", so
the runtime fell back to the "size" implementation in TraversableOnce,
which iterates over the collections of entries each time. Since the
backing map "featureNameToIdMap" is immutable, we can assume that its
size never changes, and cache the size when it's passed in.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
@@ -24,6 +24,12 @@ class DefaultIndexMap(@transient val featureNameToIdMap: Map[String, Int]) exten
   @transient
   private var _idToNameMap: Map[Int, String] = null
 
+  private val _size: Int = featureNameToIdMap.size
+
+  override def size(): Int = _size
+
+  override def isEmpty(): Boolean = size == 0
+
   override def getFeatureName(idx: Int): Option[String] = {
     if (_idToNameMap == null) {
       _idToNameMap = featureNameToIdMap.map{case (k, v) => (v, k)}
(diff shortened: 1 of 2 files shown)


convexquad:
Josh - thanks for posting this PR. It looks good to me. Also, glad to see you included a couple unit tests.

XianXing:
+1.

Will merge after the Travis CI checks finish.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
```diff
@@ -20,10 +20,16 @@ package com.linkedin.photon.ml.util
   *
   * @author yizhou
   */
-class DefaultIndexMap(@transient val featureNameToIdMap: Map[String, Int]) extends IndexMap {
+class DefaultIndexMap(@transient private val featureNameToIdMap: Map[String, Int]) extends IndexMap {
```
XianXing: This is a backward incompatible change and might break things, for example:
GLMSuiteIntegTest.scala:300: value featureNameToIdMap in class DefaultIndexMap cannot be accessed in com.linkedin.photon.ml.util.DefaultIndexMap.

Why do we want to change featureNameToIdMap to private?

joshvfleming: You're right @XianXing, I fixed it.

Generally it's best to keep members private by default, unless there's a really good reason to expose them, because otherwise you give up control over who can change the internal state of the object. In this case it doesn't really matter, since the member is immutable, but it's a good general rule to follow IMO.

PR #157, opened 2016-08-24
Fix bug where Timer is never stopped

joshvfleming:
We weren't stopping the timer here if no validating dir was
specified. Also, add test to validate this code path.


XianXing:
LGTM. Will merge after the checks completed.

ashelkovnykov:
LGTM

Surprising that a code path without validation data wasn't being run already. Perhaps we should split `defaultArgs` in DriverTest into 2 sets of arguments, one with only required arguments and another with optional arguments as well?

joshvfleming:
@ashelkovnykov That sounds like a fine idea.

On an unrelated note, it seems like Travis takes a lot longer to kick off our CI jobs these days...

XianXing:
Yes, I agree, Travis is currently blocking our model deployment 😂.

Issue #156, opened 2016-08-24
Add an option in GAME avro that it looks into metadataMap if an randomEffectId is not found in 1st layer field 


timyitong:
I'm wondering whether it's an okay to add an option here:
https://github.com/linkedin/photon-ml/blob/3b6bcb8c0211b4bd6fe4e339b14597fde1730283/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala#L120

So the suggested behavior is that we firstly look into a particular field for id, if not found, we would try a second time looking into metadataMap.

The major benefit is that then if I want to try out multiple ids, I could easily modify metadaMap field without changing the schema (schema change means either a copy-and-paste-and-mod or some hacky programmed dynamic generation) to add/delete/mod group ids in data preparation jobs.

@XianXing @joshvfleming @ashelkovnykov @TDDFT : would this be a reasonable feature to add? I'd be happy to help check in the codes.

XianXing:
Sounds reasonable to me, at least we are not making the code backward incompatible.

Since @joshvfleming is working on separating the Photon client code (e.g., Avro related stuff) from the Photon core code, he may have some insights on this issue as well.

joshvfleming:
Sounds alright to me. It could be confusing behavior to someone new to the system, though, so we need to make sure it's documented.

XianXing:
So let's do it @timyitong :)

ashelkovnykov:
Sounds reasonable to me - sorry for the late reply, misread some of the code and got myself very confused

rrohan:
+1 This is super useful.

timyitong:
@XianXing @joshvfleming @ashelkovnykov: thanks a lot for all the useful feedback! I'll start checking in the codes. Will pay attention to documentation that we just do this dark magic without explaining.

PR #158, opened 2016-08-25
Issue#156: allowing randomEffectId fetching in metadatMap as an option

timyitong:
This change will allow randomEffectId fetching to fall back to metadataMap if
a key is not present in the Avro field directly. This will make the Avro schema
a bit more flexible in the sense that users won't necessarily to change Avro
schema definition when changing the set of ids provided for training.

Addressing #156 , I modified `Utils` and added corresponding tests for reading a map. Could anyone help point me to the doc location for explaining the Avro input format, I didn't find it's explained in param parser? ...

to: @joshvfleming @XianXing @ashelkovnykov 

**Another thing:**
I'd like to introduce @jq (Julian Qian) to everyone. He is my current colleague at Airbnb Pricing team. We are starting on some experiments using Photon GAME in our models; things are still at an early stage and we have not laid out a clear plan yet. But generally speaking, we are foreseeing some changes might be needed to carry out in:
  * data formatting/conversion;
  * parameter configuration;
  * diff function and optimizer interfacing (cz. our problem is a regression problem and most likely we need to develop customized objective function implementations) and etc.

Although the internal customized modeling details might not be a good fit for PhotonML's purpose. We'd like to contribute generally useful functions to Photon ML as much as possible. Please bare with us that we might frequently file issues, start discussions or sending out PRs. 

Ps: [here](https://github.com/airbnb/aerosolve/pull/236) is a recent data conversion job we added into aerosolve library allowing us to converting aerosolve style of inputs (with transformations) into Photon GAME format.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
@@ -84,6 +84,39 @@ object DataProcessingUtils {
     }
   }
 
+  /**
+    * Given a GenericRecord, build the random effect id map:
+    *     (random effect name -> random effect id value)
+    *
+    * @note Exposed for testing purpose.
+    * @param record the avro generic record
+    * @param randomEffectIdSet a set of random effect id names
+    * @return the random effect id map of (name -> value)
+    */
+  protected[avro] def makeRandomEffectIdMap(record: GenericRecord,
+      randomEffectIdSet: Set[String]): Map[String, String] = {
+    val metaMap = Utils.getMapAvro(record, AvroFieldNames.META_DATA_MAP, isNullOK = true)
+
+    randomEffectIdSet.map { randomEffectId =>
+      val idValue = Utils.getStringAvro(record, randomEffectId, isNullOK = true)
+
+      val finalIdValue = if (idValue.isEmpty()) {
+        val mapIdValue = if (metaMap != null) metaMap.get(randomEffectId) else null
+        if (mapIdValue == null) {
+          throw new IllegalArgumentException(s"Cannot find id in either record" +
+            s"field: ${randomEffectId} or in metadataMap with key: #${randomEffectId}")
+        }
+        mapIdValue
+      } else {
+        idValue
+      }
+
+      // random effect group name -> random effect group id value
+      // random effect ids are assumed to be strings
+      (randomEffectId, finalIdValue.toString())
+    }.toMap
+  }
+
   private def getGameDatumFromGenericRecord(
       record: GenericRecord,
       featureShardSectionKeys: Map[String, Set[String]],
@@ -117,13 +150,9 @@ object DataProcessingUtils {
     } else {
       1.0
     }
-    val ids = randomEffectIdSet
-      .map { randomEffectId =>
-        (randomEffectId, Utils.getStringAvro(record, randomEffectId))
-      }
-      .toMap
 
-    new GameDatum(response, offset, weight, featureShardContainer, ids)
+    new GameDatum(response, offset, weight, featureShardContainer,
+        makeRandomEffectIdMap(record, randomEffectIdSet))
   }
 
   private def getFeaturesFromGenericRecord(
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
@@ -127,6 +127,24 @@ protected[ml] object Utils {
   }
 
   /**
+    * Fetch the java map from an Avro map field.
+    *
+    * @param record The Avro generic record
+    * @param key The field key
+    * @return A java map of String -> Object
+    */
+  def getMapAvro(
+      record: GenericRecord,
+      key: String,
+      isNullOK: Boolean = false): java.util.Map[String, JObject] = {
+    record.get(key) match {
+      case map: java.util.Map[String, JObject] => map
+      case obj: JObject => throw new IllegalArgumentException(s"$obj is not map type.")
+      case _ => if (isNullOK) null else throw new IllegalArgumentException(s"field $key is null")
+    }
+  }
+
+  /**
     * Parse String to Double
     */
   private def atod(string: String): Double = {
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroFieldNames.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroFieldNames.scala
@@ -25,4 +25,5 @@ object AvroFieldNames {
   protected[avro] val OFFSET: String = "offset"
   protected[avro] val WEIGHT: String = "weight"
   protected[avro] val UID: String = "uid"
+  protected[avro] val META_DATA_MAP: String = "metadataMap"
 }
--- a/photon-ml/src/test/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtilsTest.scala
+++ b/photon-ml/src/test/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtilsTest.scala
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2016 LinkedIn Corp. All rights reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.avro.data
+
+import com.linkedin.photon.ml.avro.AvroFieldNames
+import org.apache.avro.SchemaBuilder
+import org.apache.avro.generic.GenericData
+import org.testng.annotations.Test
+import org.testng.Assert._
+
+/**
+  * This class tests the DataProcessingUtils.
+  *
+  * @see [[com.linkedin.photon.ml.avro.data.DataProcessingUtils]]
+  */
+class DataProcessingUtilsTest {
+
+  @Test
+  def testMakeRandomEffectIdMapWithIdField(): Unit = {
+    val record = new GenericData.Record(SchemaBuilder
+      .record("testRecordForRandomIdFetch1")
+      .namespace("com.linkedin.photon.ml.avro.data")
+      .fields()
+      .name("userId").`type`().stringType().noDefault()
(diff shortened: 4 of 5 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -117,9 +117,24 @@ object DataProcessingUtils {
...
     val ids = randomEffectIdSet
       .map { randomEffectId =>
-        (randomEffectId, Utils.getStringAvro(record, randomEffectId))
+        val idValue = Utils.getStringAvro(record, randomEffectId, isNullOK = true)
+        val finalIdValue = if (idValue == null) {
+          val mapIdValue = metaMap.get(randomEffectId)
+          if (mapIdValue == null) {
+            throw new IllegalArgumentException(s"Cannot find id in either record" +
+              s"field: ${randomEffectId} or in metadataMap with key: #${randomEffectId}")
+          }
```
XianXing: Shall we return ```mapIdValue.toString()``` here?

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -117,9 +117,24 @@ object DataProcessingUtils {
     } else {
       1.0
     }
+
+    val metaMap = Utils.getMapAvro(record, AvroFieldNames.META_DATA_MAP)
```
XianXing: Looks like if ```META_DATA_MAP``` is null in the record, an ```IllegalArgumentException``` will be thrown.

One suggestion is to return an empty ```Map[String, JObject]``` instead of a ```Null```.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -117,9 +117,24 @@ object DataProcessingUtils {
...
     val ids = randomEffectIdSet
       .map { randomEffectId =>
-        (randomEffectId, Utils.getStringAvro(record, randomEffectId))
+        val idValue = Utils.getStringAvro(record, randomEffectId, isNullOK = true)
+        val finalIdValue = if (idValue == null) {
```
XianXing: Shall we add a test to cover this new code path here?

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
timyitong: I can add the get method to be nullable ok. That's probably more general.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
timyitong: We called an `toString` in the end actually.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
timyitong: I think there is an integTest related with these part of changes. The previous non-nullable change actually failed it. Sorry about that, I'll see whether I could add a branch there to make it covered.

XianXing:
For the Avro input format, you are right, we currently don't have any documentation for it yet. Partly because we are asking Photon to detect the schema of the input in runtime, for example, the ids fields present in the data can be different from data sets to data sets, as well as those feature bags.

For example, the input records can contain ids such as "houseId", "userId", "city", "zipcode", and during different runs of the GAME model we can specify different types of random effect models, like the first model is a per-house + per-user model, the second is a per-zipcode + per-user model. We only need to change the configs of the model, without the need to change the data.


However, I think you raised a good question, and we should better document the input schema so that the users can better leverage it.

XianXing:
Hi @jq! Nice to know you here. I just looked into [aerosolve](https://github.com/airbnb/aerosolve) and it's awesome!

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
XianXing: Yes I know we call ```toString``` in the end, but I was wondering don't we need to return anything after the ```if (mapIdValue == null) {...}``` branch, or did I miss anything?

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
XianXing: Some folks may argue that ```Null``` is evil and we should not see any ```Null``` in Scala (e.g., an empty ```Map``` here instead). But this is a minor issue and I am fine with the nullable way of handling things.

jq:
Thanks for @timyitong intro, and looking forward to meet @XianXing @ashelkovnykov, @joshvfleming in the near future!

ashelkovnykov:
@XianXing @joshvfleming @timyitong I spoke with Xianxing offline - we agreed that we should have documentation about the Avro input for GAME. Xianxing suggested adding it to the Avro input section of the README and I agree that currently that is the most logical place to put it. After the API changes that Josh is working on are added, the documentation should be moved to the class responsible for reading Avro data as input.

@jq Nice to meet you

@timyitong I'll take a look at this PR tomorrow morning

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -117,9 +117,24 @@ object DataProcessingUtils {
...
     }
+
+    val metaMap = Utils.getMapAvro(record, AvroFieldNames.META_DATA_MAP, isNullOK = true)
     val ids = randomEffectIdSet
       .map { randomEffectId =>
-        (randomEffectId, Utils.getStringAvro(record, randomEffectId))
+        val idValue = Utils.getStringAvro(record, randomEffectId, isNullOK = true)
+        val finalIdValue = if (idValue == null) {
+          val mapIdValue = if (metaMap != null) metaMap.get(randomEffectId) else null
+          if ( mapIdValue == null) {
```
joshvfleming: Minor: Extra space before ```mapIdValue```

photon-ml/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
```diff
@@ -127,6 +127,22 @@ protected[ml] object Utils {
...
 
   /**
+    * Fetch the java map from an Avro map field.
+    *
+    * @param record The Avro generic record
+    * @param key The field key
+    * @return A java map of String -> Object
+    */
+  def getMapAvro(record: GenericRecord,key: String, isNullOK: Boolean = false):
+      java.util.Map[String, JObject] = {
```
joshvfleming: Minor: Couple scala style issues here. There's a space missing after the first comma, and each parameter should have its own newline with 4-space indentation if you need a linebreak, e.g.:

```scala
def getMapAvro(
    record: GenericRecord,
    key: String,
    isNullOK: Boolean = false): java.util.Map[String, JObject] = {
```

joshvfleming:
Nice to meet you, @jq !

One word of warning: there are some big changes coming soon to the Photon / GAME API, particularly w.r.t. data formats and handling. I'll try to keep you guys in the loop with all the changes.

timyitong:
@joshvfleming : really sorry for all the style issues, I'm messing it up with other auto style correction config. Mind sharing an Idea style config file in the repo? I tried to import `scalastyle-config.xml` directly, but it doesn't seem to be working my Idea. (maybe the config is universal but somehow my Idea version is problematic) And thanks a lot for reminding us the potential changes in the APIs. Actually I am quite looking forward to it as it seems that we could possibly loosen up the data format assumption here and allow other format adapters : )

I extracted the id fetching logics into a `protected[avro]` method and then added unit tests for it. Figured it might be too crazy to set it up in `integTest` as I need to mock a new version of testing avro data which I don't think make too much sense since it's just a minor schema difference. Hope this assumption is reasonable?

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -84,6 +84,39 @@ object DataProcessingUtils {
...
+    * Given a GenericRecord, build the random effect id map:
+    *     (random effect name -> random effect id value)
+    *
+    * @note Exposed for testing purpose.
+    * @param record the avro generic record
+    * @param randomEffectIdSet a set of random effect id names
+    * @return the random effect id map of (name -> value)
+    */
+  protected[avro] def makeRandomEffectIdMap(record: GenericRecord,
+      randomEffectIdSet: Set[String]): Map[String, String] = {
```
joshvfleming: Sorry, another minor one @timyitong: ```record: GenericRecord``` should be on its own line, aligned with ```randomEffectIdSet```.

I don't believe we have a scalaStyle rule that checks for this currently. I'll look into that today if I have time.

joshvfleming:
LGTM

I don't have an Idea style config, but @XianXing or @ashelkovnykov might. I don't believe we have rules for some of these issues also -- I'll look into that today if I have a chance.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
XianXing: I will take this one and fix it in the next PR.

XianXing:
LGTM. I am going to merge this PR, but please feel free to continue the discussion on Scala style issues in this thread.

ashelkovnykov:
I'm late to the party but LGTM.

One thing: I believe that the test values in the unit tests should be stored and tested as constants, ex:

```
val jobIdVal = 113L
val jobIdValStr = "113"

record.put("jobId", jobIdVal)
...
assertEquals(res("jobId"), jobIdValStr)
```

@timyitong @joshvfleming I kind of have an IDEA style config - it does most of the style changes on its own but I still need to manually do some of the changes. I'll see if I can make it stick to the style as strictly as possible and then I can try to share it with @timyitong.

If it works well, there's no harm in adding it to the repo. Could lead to some good starter issues for people who want to contribute to Photon - make a style config file for Eclipse, or some other IDE.

XianXing:
Good point @ashelkovnykov, the IDEA style config would be *very* helpful, as I always believe the code style issues are very important, but they should be handled by the machine (ideally) but not human beings.

But let's please stick with our high standards on Scala code style issues for all PRs before coming up with a better solution!

P.S., I will fix those test constants issues together with a few other Scala style issues in a PR that will be submitted shortly.

ashelkovnykov:
Just a follow up: I took a look at the IDEA style config and I can't quite get the Spark style locked in. I don't want to invest my time into it right now, so I'll add it as a starter issue, in case some generous soul decides they would like to do it. Otherwise, I'll take a look at it during the next hack-day.

PR #160, opened 2016-08-26
Integrate PalDB off-heap index map into GAME

joshvfleming:
## Performance Comparison Results

The following shows a performance comparison between the master branch before this PR, after this PR, and after this PR with PalDB index map enabled. A real-world proprietary dataset was used to give realistic figures.

### Feature Set / Index Map Generation
NameAndTermFeatureSetContainer: **19m 40s**
FeatureIndexingJob: **27m 20s**

### Learned Coefficients, Before PR
Model name: global, summary:
Count: 22468
Mean: 1.971609503055699E-4
Variance: 0.0015829544632712039
L2 Norm: 5.963649160732309

Model name: per-x1-factor, summary:

| Measure | Count | Mean | Std. | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Length | 2370903 | 347.456437 | 316.488609 | 1949 | 5 |
| Mean | 2370903 | 0.318371 | 0.202498 | 3.130833 | 0.000045 |

Model name: per-x2-factor, summary:

| Measure | Count | Mean | Std. | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Length | 967254 | 572.243904 | 521.408027 | 2038 | 4 |
| Mean | 967254 | 0.136786 | 0.154131 | 2.728295 | 0.000027 |

Evaluation result is : **0.7795401439192408**
Overall time elapsed **27.20245588006667** minutes

### Learned Coefficients, After PR
Model name: global, summary:
Count: 22468
Mean: 1.9693874274906873E-4
Variance: 0.0015806326238098128
L2 Norm: 5.9592738297176835

Model name: per-x1-factor, summary:

| Measure | Count | Mean | Std. | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Length | 2370903 | 347.456437 | 316.488609 | 1949 | 5 |
| Mean | 2370903 | 0.318372 | 0.202496 | 3.130997 | 0.000045 |

Model name: per-x2-factor, summary:

| Measure | Count | Mean | Std. | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Length | 967254 | 572.243904 | 521.408027 | 2038 | 4 |
| Mean | 967254 | 0.136786 | 0.154137 | 2.728520 | 0.000027 |

Evaluation result is : **0.7795494384777749**
Overall time elapsed **26.4413001128** minutes

### Learned Coefficients, After PR (with PalDB index map enabled)
Model name: global, summary:
Count: 22468
Mean: 1.7330011905692327E-4
Variance: 0.001537551943564667
L2 Norm: 5.877487073228393

Model name: per-x1-factor, summary:

| Measure | Count | Mean | Std. | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Length | 2370903 | 347.456437 | 316.488609 | 1949 | 5 |
| Mean | 2370903 | 0.318041 | 0.202916 | 3.132190 | 0.000043 |

Model name: per-x2-factor, summary:

| Measure | Count | Mean | Std. | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Length | 967254 | 572.243904 | 521.408027 | 2038 | 4 |
| Mean | 967254 | 0.136706 | 0.154347 | 2.727148 | 0.000028 |

Evaluation result is : **0.7794371544608726**
Overall time elapsed **29.7215908014** minutes


photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -19,40 +19,59 @@ import org.apache.hadoop.conf.Configuration
...
+    val featureShardIdToFeatureSectionKeysMap: Option[Map[String, Set[String]]] = None) {
+
+  private val logger: PhotonLogger = new PhotonLogger(new Path(outputPath, "_log"), sc)
+
+  /**
+   * Ensures that the output path exists.
+   */
+  private def ensureOutputPath(outputPath: String): Path = {
+    val path = new Path(outputPath)
     val fs = FileSystem.get(new Configuration())
```
timyitong: Use `path.getFileSystem` instead? There is slight different that `path.getFileSystem` would pick the right cluster host while `FileSystem.get(new Configuration())` will use the host where the spark job is residing. It's almost the same but `path.getFileSystem` should be more general (equivalent to `FileSystem.get(new Configuration(), path.toUri())`)

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Params.scala
```diff
@@ -14,23 +14,24 @@
...
-import scala.collection.{Map, Set}
-
 /**
  * A bean class for GAME training parameters to replace the original case class for input parameters.
  * @note Note that examples of how to configure GAME parameters can be found in the integration tests for the GAME
  *       driver.
  * @todo Making the way GAME being configured more user friendly
  */
-class Params {
+class Params extends FeatureParams with PalDBIndexMapParams {
```
timyitong: Just want to understand this more, so we are now saying a `Param` could be a collection of multiple `XXXParam`? It seems to be a nice way of framing it.

joshvfleming: That's right, @timyitong. This way we can group the related Params together, and pass in an object of type ```IndexMapParams``` object into the ```IndexMapLoader``` instead of the app-global ```Params``` type.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
joshvfleming: Ah, nice. I'll make that change.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -19,40 +19,59 @@ import org.apache.hadoop.conf.Configuration
...
-  * The job expected three required arguments and one optional one:
-  * [input_data_path]: The input path of data
-  * [partition_num]: The number of partitions to break the storage into. This is merely introduced as an optimization at
-  *   build stage that we don't necessary have to shuffle all features into the same partition but rather we could
-  *   just hold them into multiple ones. A heuristically good number could be 1-10, or even as large as 20; but please
+  * The job expects three required arguments and three optional ones:
+  *
+  * input-paths: The input paths of the data
+  * partition-num: The number of partitions to break the storage into. This is merely introduced as an optimization at
+  *   the building stage where we don't necessary have to shuffle all features into the same partition, but rather we
```
ashelkovnykov: Typo: `necessary -> necessarily`, poor grammar to use 'have to' in place of 'must'

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Params.scala
ashelkovnykov: Very nice, +1

ashelkovnykov:
LGTM

timyitong:
LGTM + 1

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -167,8 +184,9 @@ object DataProcessingUtils {
           recordList.asScala.flatMap {
             case record: GenericRecord =>
               val nameAndTerm = AvroUtils.readNameAndTermFromGenericRecord(record)
-              if (nameAndTermToIndexMap.contains(nameAndTerm)) {
-                Some(nameAndTermToIndexMap(nameAndTerm) -> Utils.getDoubleAvro(record, AvroFieldNames.VALUE))
+              val featureKey = Utils.getFeatureKey(nameAndTerm.name, nameAndTerm.term)
+              if (featureMap.contains(featureKey)) {
+                Some(featureMap(featureKey) -> Utils.getDoubleAvro(record, AvroFieldNames.VALUE))
```
XianXing: Nit: How about ```featureMap(featureKey)``` -> ```featureMap.getIndex(featureKey)``` for readability?

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
```diff
@@ -177,10 +195,10 @@ object DataProcessingUtils {
         case _ => throw new IllegalArgumentException(s"$fieldName is not a list (or is null).")
       }
     ).foldLeft(Array[(Int, Double)]())(_ ++ _)
-    val isAddingInterceptToFeatureMap = nameAndTermToIndexMap.contains(NameAndTerm.INTERCEPT_NAME_AND_TERM)
+    val isAddingInterceptToFeatureMap = featureMap.contains(GLMSuite.INTERCEPT_NAME_TERM)
     if (isAddingInterceptToFeatureMap) {
       VectorUtils.convertIndexAndValuePairArrayToSparseVector(featuresAsIndexValueArray ++
-        Array(nameAndTermToIndexMap(NameAndTerm.INTERCEPT_NAME_AND_TERM) -> 1.0), featureDimension)
+        Array(featureMap(GLMSuite.INTERCEPT_NAME_TERM) -> 1.0), featureDimension)
```
XianXing: Same as above.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/model/ModelProcessingUtils.scala
```diff
@@ -195,24 +189,27 @@ object ModelProcessingUtils {
...
   private def saveModelsRDDToHDFS(
       modelsRDD: RDD[(String, GeneralizedLinearModel)],
-      featureIndexToNameAndTermMapBroadcast: Broadcast[Map[Int, NameAndTerm]],
+      featureMapLoader: IndexMapLoader,
       outputDir: String): Unit = {
 
-    val linearModelAvro = modelsRDD.map { case (modelId, model) =>
-      AvroUtils.convertGLMModelToBayesianLinearModelAvro(model, modelId, featureIndexToNameAndTermMapBroadcast.value)
+    val linearModelAvro = modelsRDD.mapPartitions { iter =>
+      val featureMap = featureMapLoader.indexMapForRDD
```
XianXing: This is great, ```indexMapForRDD``` should always be called inside the RDD operation closures.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
```diff
@@ -70,11 +70,10 @@ protected[ml] object Utils {
     * @return The feature name
     */
   def getFeatureNameFromKey(key: String, delimiter: String = GLMSuite.DELIMITER): String = {
-    val tokens = key.split(delimiter)
-    if (tokens.size != 2) {
-      throw new IllegalArgumentException(s"Provided input [$key] is not a valid feature key")
-    }
-    tokens(0)
+    require(delimiter.r.findAllIn(key).length == 1, s"Provided input [$key] is not a valid feature key")
```
XianXing: Can we add a test case which fails such ```require``` check?

photon-ml/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
```diff
@@ -70,11 +70,10 @@ protected[ml] object Utils {
...
+    key.split(delimiter)
+      .lift(0)
```
XianXing: +1 for ```lift```.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -19,79 +19,104 @@ import org.apache.hadoop.conf.Configuration
...
   *   avoid setting it into an arbitrarily large number. Use just 1 or 2 if you find the indexing job is already fast
   *   enough.
-  * [output_dir]: The output directory
-  * [if_add_intercept] (optional, default=true): whether include INTERCEPT into the map.
-  *
+  * output-path: The output path
+  * date-range: Date range for the input data represented in the form start.date-end.date, e.g. 20150501-20150631
+  * date-range-days-ago: Date range for the input data represented in the form start.daysAgo-end.daysAgo, e.g. 90-1
+  * add-intercept: whether include INTERCEPT in the map (optional, default: true).
+  * data-format: Input data format. (optional, default: TRAINING_EXAMPLE, options:
```
XianXing: What is ```data-format``` doing here?

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -19,79 +19,104 @@ import org.apache.hadoop.conf.Configuration
...
+  *   [TRAINING_EXAMPLE|RESPONSE_PREDICTION])
+  * feature-shard-id-to-feature-section-keys-map: A map between the feature shard id and it's corresponding feature
+  *   section keys, in the following format: shardId1:sectionKey1,sectionKey2|shardId2:sectionKey2,sectionKey3.
+  * feature-shard-id-to-intercept-map: A map between the feature shard id and a boolean variable that decides whether a
```
XianXing: Using such a map to determine whether an intercept term should be added for each feature shard is desired, but the following uses a single Boolean variables ```addIntercept``` for this purpose.

Please either update the docs, or let's really use ```featureShardIdToInterceptMap``` here instead.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -138,39 +171,159 @@ class FeatureIndexingJob(val sc: SparkContext,
...
+      case Some(shardToSectionsMap) =>
+        // If a shard to feature section set was specified, build an index for each shard
+        shardToSectionsMap.map { case (shardId, featureSections) =>
+          val featuresRdd = partitionedUniqueFeatures(inputRdd, Some(featureSections))
+          buildIndexMap(featuresRdd, outputPath, shardId)
+        }
+
+      case _ => {
+        // Otherwise, build a global index
+        val featuresRdd = partitionedUniqueFeatures(inputRdd)
```
XianXing: The more I think about such design choice, the more I feel like that sharded feature maps might not be necessary at all, and a global index map might be suffice, and is much cleaner.

This is because the feature indexing job should be only care about mapping the feature keys from Strings to Ints, and such mapping shouldn't be shard-dependent.

But I am not suggesting that we should make a decision in this PR, just think aloud here.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -107,16 +132,24 @@ class FeatureIndexingJob(val sc: SparkContext,
     keyedFeaturesUnionedRDD.distinct().groupByKey(new HashPartitioner(partitionNum))
```
XianXing: @timyitong Can you remind me why the fact that "integer's hashcode is still itself" would save the amount of data to be shuffled?

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/GAMEDriver.scala
```diff
@@ -0,0 +1,89 @@
...
+ */
+package com.linkedin.photon.ml.cli.game
+
+import com.linkedin.photon.ml.util._
+import com.linkedin.photon.ml.avro.data.NameAndTermFeatureSetContainer
+
+import org.apache.spark.SparkContext
+
+/**
+  * Contains common functions for GAME training and scoring drivers.
```
XianXing: +1, nice refactoring.

XianXing:
A great PR with loads of craftsmanship @joshvfleming!

Please let me know when we finished adding additional tests and done with performance benchmarking.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
timyitong: Memory is a bit blurred here. I think previously I was worried that because I'm using `HashPartitioner` in group by key, and inside of HashPartitioner we also we do the hashing for a second time, so it's like `hash(hash(feature_str))`; because `hash(int) = int` I don't need to worry about in the reading stage, I cannot reproduce the hash code exactly; otherwise if `hash(int) != int` I might need to pass in some unique string key instead to make sure the hash code is reproducible and then string keys might be worse compared to int keys. (Yeah, the derivation is long here....)

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
joshvfleming: Yes, sorry about that. I originally had an intercept option for each shard, but then I reasoned as follows: there's really not much cost to having an "extra" entry in the feature map in cases where we globally add an intercept, but don't necessarily use it in certain shards. And the interface is a bit simpler if we just have a global setting.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
joshvfleming: I think the idea is that we wouldn't hard-code the "features" field name. Looks like in both TrainingExample and ResponsePrediction, it's called "features", so we probably don't need this at present.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
joshvfleming: The original code had a bug: when the term is an empty string "", the size of the split array is 1, e.g.:

```scala
"name,".split(",")
// => res1: Array[String] = Array(name)
```

This causes the ```tokens.size != 2``` check to fail.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/data/DataProcessingUtils.scala
joshvfleming: Yes, I got lazy there. Will make that change.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
joshvfleming: Interesting. This would break the nice conceptual mapping we have between feature shards and the individual feature vectors in the GAME model though, wouldn't it?

ashelkovnykov: Currently each shard has unique feature indices, yes? Because they're stored in the smallest possible vector, so indices are reused between shards. We'd still need a mapping of global index -> shard index.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
```diff
@@ -19,79 +19,100 @@ import org.apache.hadoop.conf.Configuration
...
-                         val fieldNames: FieldNames) {
-  private val outputPath: Path = {
-    val path = new Path(outputDir)
-    val fs = FileSystem.get(new Configuration())
+class FeatureIndexingJob(
+    val sc: SparkContext,
+    val inputPaths: Seq[String],
+    val partitionNum: Int,
+    val outputPath: String,
+    val addIntercept: Boolean,
```
XianXing: Can you double check the way we deal with intercepts and the impact of using the global flag ```addIntercept``` instead of a per-shard one?

Line 182 (before this PR) in ```DataProcessingUtils.scala```:
```val isAddingInterceptToFeatureMap = nameAndTermToIndexMap.contains(NameAndTerm.INTERCEPT_NAME_AND_TERM)```
So it loos like we are deciding whether to add the intercept based on whether the intercept term is added to the feature map or not.

Are we keeping the same logic as above after this PR?

File: photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala (hunk 11 above)
Comment (XianXing) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala]: @ashelkovnykov Yes. But we still have a mapping available from ```IndexMapProjector```, so potentially we can get rid of the sharded feature maps.

Let's move the discussion offline or open an issue to formally discuss it. For now, let's still adopt the current way of using sharded feature maps.

File: photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala (hunk 14 above)
Comment (joshvfleming) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala]: I think we're also using the feature map to calculate feature dimension in a couple places. I'll check into it.

Comment (joshvfleming) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala]: @XianXing I went ahead and changed this to a per-shard configuration.

File: photon-ml/src/test/scala/com/linkedin/photon/ml/util/PalDBIndexMapLoaderTest.scala [hunk 15]
```diff
@@ -0,0 +1,49 @@
...
+      Array("s3a://bucket", "/path/to/resource", "s3a://bucket/path/to/resource"))
+  }
+
+  @Test(dataProvider = "pathDataProvider")
+  def testGetPath(defaultName: String, inputPath: String, expectedResult: String): Unit = {
+    val loader = new PalDBIndexMapLoader
+    val path = new Path(inputPath)
+
+    val sc = mock(classOf[SparkContext])
+    val conf = mock(classOf[Configuration])
```
XianXing: Nice, does this mean that we don't need to put all tests into integTests that needs ```sc```?

joshvfleming: Yep. If you know exactly which parts of ```sc``` will be used, this is a lighter option.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
XianXing: Thanks. Then we don't need ```addIntercept``` in line 63 anymore, right?

photon-ml/src/test/scala/com/linkedin/photon/ml/util/PalDBIndexMapLoaderTest.scala
joshvfleming: One word of warning, though: it doesn't work very well with the RDD interface.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/GAMEDriver.scala
```diff
@@ -0,0 +1,89 @@
...
+  protected[game] def prepareFeatureMapsDefault(): Map[String, IndexMapLoader] = {
+    val allFeatureSectionKeys = featureShardIdToFeatureSectionKeysMap.values.reduce(_ ++ _)
+    val nameAndTermFeatureSetContainer = NameAndTermFeatureSetContainer.readNameAndTermFeatureSetContainerFromTextFiles(
+      featureNameAndTermSetInputPath, allFeatureSectionKeys, hadoopConfiguration)
+
+    val featureShardIdToFeatureMapLoader =
+      featureShardIdToFeatureSectionKeysMap.map { case (shardId, featureSectionKeys) =>
+        val featureMap = nameAndTermFeatureSetContainer
+          .getFeatureNameAndTermToIndexMap(featureSectionKeys, featureShardIdToInterceptMap.getOrElse(shardId, true))
+          .map { case (k, v) => Utils.getFeatureKey(k.name, k.term) -> v }
```
XianXing: It looks like that ```featureShardIdToInterceptMap``` now is only needed to prepare default feature maps.

Now we have some API level inconsistency in adding intercept to different feature shard between default index map and PalDB based index map.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMapLoader.scala
```diff
@@ -27,21 +28,52 @@ import org.apache.spark.SparkContext
...
+      _numPartitions = palDBParams.offHeapIndexMapNumPartitions
+      _namespace = namespace
 
-  override def prepare(sc: SparkContext, params: Params): Unit = {
-    if (!params.offHeapIndexMapDir.isEmpty && params.offHeapIndexMapNumPartitions != 0) {
-      _storeDir = params.offHeapIndexMapDir.get
-      _numPartitions = params.offHeapIndexMapNumPartitions
       (0 until _numPartitions).foreach(i =>
-        sc.addFile(new Path(_storeDir, PalDBIndexMap.partitionFilename(i)).toUri().toString())
+        sc.addFile(getPath(sc, new Path(_storeDir, PalDBIndexMap.partitionFilename(i, namespace))))
```
XianXing: This is really nice.

@timyitong Have we faced any issues previously when the input paths were not well taken care of?

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
joshvfleming: Actually I was just thinking about this. We might have consumers of this job who don't use shards (e.g. old-school Photon). In that case, the "addIntercept" global option still makes more sense. I'm not too happy about having both methods available at once, though.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/GAMEDriver.scala
joshvfleming: Yeah. This is an artifact of the difference of when the index map is created. In the offheap case, it's created during the build pre-job. In the default case, it's created during the main training job.

My thinking here was that the purpose of the "default" code path is backwards compatibility, and we'd deprecate it in favor of the "offheap" path once we're happy with that path and have everyone upgraded. Not saying that necessarily correct, though.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
XianXing: First of all I think we can better take care of the user-facing APIs through the client code refactoring and the Photon templates. 

For now, I guess we can make ```addIntercept``` as the default value for ```featureShardIdToInterceptMap```. For example, if the user only specifies ```addIntercept```but not ```featureShardIdToInterceptMap```, or specifies it only for a subset of shard Ids, then the missing shard id's value inn ```featureShardIdToInterceptMap``` will be initialized to ```addIntercept```, instead of ```true```.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/GAMEDriver.scala
XianXing: We may still want to keep the default path for a while, because it's more efficient and lightweight when the number of features is not large (e.g., number features < tens of thousands).

It bothers me a little bit, however, I am not asking that we need to make them consistent in this PR, as this is a minor point compared to what we've achieved in this PR.

joshvfleming: Yep, makes sense.

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
joshvfleming: Agreed. I've made this change.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMapLoader.scala
```diff
@@ -54,4 +55,25 @@ class PalDBIndexMapLoader extends IndexMapLoader {
...
+    var uri = path.toUri
+
+    Option(uri.getScheme) match {
+      case Some(scheme) => uri.toString
+      case _ =>
+        // If the path specifies no scheme, use the current default
+        val default = new Path(sc.hadoopConfiguration.get("fs.default.name")).toUri
+        new URI(
+          default.getScheme, default.getUserInfo, default.getHost, default.getPort,
+          uri.getPath, uri.getQuery, uri.getFragment).toString
```
ashelkovnykov: Scala style: If all of the arguments cannot fit on one line, they should each be on a separate line

photon-ml/src/integTest/scala/com/linkedin/photon/ml/FeatureIndexingJobTest.scala
```diff
@@ -68,16 +73,16 @@ class FeatureIndexingJobTest {
...
-        featureShardIdToExpectedDimension)
+        featureShardIdToInterceptMap, featureShardIdToExpectedDimension)
       testShardedJob(
         tempFile.toString, partitionNum, addIntercept = true, featureShardIdToFeatureSectionKeysMap,
-        featureShardIdToExpectedDimension)
+        featureShardIdToInterceptMap, featureShardIdToExpectedDimension)
       testShardedJob(
         tempFile.toString, partitionNum, addIntercept = false, featureShardIdToFeatureSectionKeysMap,
-        featureShardIdToExpectedDimension)
+        featureShardIdToInterceptMap, featureShardIdToExpectedDimension)
```
ashelkovnykov: Scala style: If all of the arguments cannot fit on one line, they should each be on a separate line

photon-ml/src/main/scala/com/linkedin/photon/ml/FeatureIndexingJob.scala
XianXing: But I don't see ```data-format``` as one parameter in ```FeatureIndexingJob```.

XianXing: It might be better if we use "@param" annotation to document all parameters, and IntelliJ would highlight all those mismatches like "data-format".

ashelkovnykov: @joshvfleming Join us Josh - abandon emacs forever

joshvfleming: It looks like this is documenting the command line api rather than the class. I don't feel strongly about either keeping it that way or changing it.

@ashelkovnykov never!

Issue #151, opened 2016-08-17
Precision at n evaluation & optimization


Mschneider93:
Hi guys,

I use recommendation systems only with respect to a certain rank. In particular, on the website we can only display 4 recommendations. Would it be possible optimize and to benchmark on "precision @ n"?  

Best,
Michael

XianXing:
Good point, I will look into it and see how can we add precision@n as a new evaluator to GAME.

pogil:
@Mschneider93, do you mean precision@n for an entire dataset, or do you mean the average across context for precision@n on items sharing some context id?

XianXing:
Precision @ k as an evaluator is added to Photon, the next step is to refactor Photon so that different algorithms can call it for evaluation purpose.

Mschneider93:
@XianXing that is awesome I check that out

@pogil good point, having both information can be extremely helpful. precision@n for an entire dataset can be used to justify the overall expect performance. And the precisio@n depending on a specific context helps certainly in communicating with project/context owners.  

Thanks a lot - this is a really awesome project

XianXing:
Precision@K is introduced to Photon-ML through #166.

PR #162, opened 2016-08-30
Add precision@K evaluator with tests.

XianXing:
Part of the effort to address issue #151.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PrecisionAtKEvaluator.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PrecisionAtKEvaluator.scala
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2016 LinkedIn Corp. All rights reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.evaluation
+
+import org.apache.spark.rdd.RDD
+
+import com.linkedin.photon.ml.constants.MathConst
+
+/**
+ * Evaluator for Precision@k. Definition of this evaluator along with terminologies used here (documentId) can be
+ * found at [[https://en.wikipedia.org/wiki/Information_retrieval#Precision_at_K]]. No special tiebreaker is used if
+ * there are ties in scores, and one of them will be picked randomly.
+ * @param k The cut-off rank based on which the precision is computed (precision @ k)
+ * @param labelAndOffsetAndWeights A [[RDD]] of (id, (label, offset, weight)) tuples
+ * @param documentIds Document ids based on which the labels and scores are grouped to form documents in order to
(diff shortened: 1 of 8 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PrecisionAtKEvaluator.scala
```diff
@@ -0,0 +1,55 @@
...
+      .mapValues { case (scoreOption, (label, offset, _)) =>
+        (scoreOption.getOrElse(defaultScore) + offset, label)
+      }
+    val posThreshold = MathConst.POSITIVE_RESPONSE_THRESHOLD
+    documentIds
+      .join(scoreAndLabels)
+      .values
+      .groupByKey()
+      .values
+      .map(_.toArray.sortBy(_._1)(Ordering[Double].reverse).take(k).count(_._2 > posThreshold) * 1.0 / k)
```
ankans: This looks really concise in Scala! Thanks for providing the Precision@k metric!

ashelkovnykov:
LGTM

I hope `groupByKey` and sorting per documentID isn't too slow

XianXing:
In most cases, we are expecting tens of items per key to compute precision@k, and the distribution of items for each key won't be skewed. As a result, using ```groupByKey``` here is a safe choice.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PrecisionAtKEvaluator.scala
```diff
@@ -0,0 +1,55 @@
...
+
+import org.apache.spark.rdd.RDD
+
+import com.linkedin.photon.ml.constants.MathConst
+
+/**
+ * Evaluator for binary classification problems
+ * @param k The cut-off rank based on which the precision is computed (precision @ k)
+ * @param labelAndOffsetAndWeights A [[RDD]] of (id, (label, offset, weight)) tuples
+ * @param documentIds Document ids based on which the labels and scores are grouped to form documents in order to
```
pogil: Is it fair to think of the documentIds as a recommendation context? E.g. in evaluating the job recommendations for members - one would use the member id as a documentId?

XianXing: Yes.

I chose the term "documentId" because this is how precision@k is introduced in the IR field (https://en.wikipedia.org/wiki/Information_retrieval#Precision_at_K).

But you made a good point, let me better document this.

pogil: In the IR field I would personally tend to think of it as a query rather than a documentId. Adding a little documentation here would go a long way. Also line 22 - references "binary classification", when this is useful as a general metric for ranking problems. It would also be good to document behavior when there are ties in scores.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/PrecisionAtKEvaluator.scala
```diff
@@ -0,0 +1,59 @@
...
+ *                    compute precision @ K. Such document ids can be thought as a recommendation context, e.g. in
+ *                    evaluating the relevance of search results of given a query - one would use the query id as a
+ *                    documentId.
+ * @param defaultScore The default score used to compute the metric
+ */
+class PrecisionAtKEvaluator(
+  k: Int,
+  labelAndOffsetAndWeights: RDD[(Long, (Double, Double, Double))],
+  documentIds: RDD[(Long, String)],
+  defaultScore: Double = 0.0) extends Evaluator {
```
joshvfleming: Minor: these need to be indented 2 more spaces.

joshvfleming:
LGTM

PR #169, opened 2016-09-22
Bug fix: ids used to compute precision@k should be considered explici…

XianXing:
…tly when generating GAMEDatum from Avro generic records.

Otherwise, fixed effect model will not be able to use precision@k, even full GLMix model cannot use precision@k if the id used to compute the metric is not random effect id.

But unit test and integTest are updated to catch such a bug.

### DIFF
--- a/photon-ml/src/test/scala/com/linkedin/photon/ml/cli/game/EvaluatorParamsTest.scala
+++ b/photon-ml/src/test/scala/com/linkedin/photon/ml/cli/game/EvaluatorParamsTest.scala
@@ -14,24 +14,21 @@
  */
 package com.linkedin.photon.ml.cli.game
 
-import com.linkedin.photon.ml.evaluation.{EvaluatorType, PrecisionAtK}
+import org.testng.Assert._
+import org.testng.annotations.Test
 
-/**
- * Evaluator params common to GAME training and scoring.
- */
-trait EvaluatorParams {
-  /**
-   * A list of evaluators separated by comma. E.g, AUC,Precision@1:documentId,Precision@3:documentId,Logistic_Loss
-   */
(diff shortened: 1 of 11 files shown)


photon-ml/src/test/scala/com/linkedin/photon/ml/cli/game/EvaluatorParamsTest.scala
```diff
@@ -0,0 +1,34 @@
...
+
+class EvaluatorParamsTest {
+
+  @Test
+  def precisionAtKIdTypeSetTest(): Unit = {
+    val expectedIdTypeSet = Set("documentId", "queryId", "foo", "bar")
+    val precisionAtKEvaluators = expectedIdTypeSet.toSeq.flatMap(t => Seq(1, 3, 5, 10).map(PrecisionAtK(_, t)))
+    val allEvaluators = Seq(AUC, RMSE, LogisticLoss) ++ precisionAtKEvaluators
+
+    val evaluatorParamsMocker = new EvaluatorParams {}
```
XianXing: Since ```EvaluatorParams``` doesn't contain any functions that need to be implemented, here an anonymous class is created for unit test. However, if we add unimplemented functions to trait ```EvaluatorParams``` in the future, the creation of the anonymous class here will break.

@joshvfleming, do you have any suggestions on how to deal with such scenario, can we use some tools like Mockito to better handle such cases?

joshvfleming: Looks fine to me. You don't really want to mock this object, because you're testing one of its methods. Also, a spy would have the same issue, since you'd need to create the object first.

PR #166, opened 2016-09-09
Add precision@K to GAME, and address some related issues.

XianXing:
A few things to watch out for in this PR:

0. Precision@K is now ready to be used as an evaluator in GAME, similar to other metrics such as AUC and RMSE.

1. ```GameDatum``` now hold all types of Ids other than just random effect ids. For example, ```uid``` used for some scoring downstream applications and ```documentId``` or ```queryId``` used to compute precision@K metric (or replay metric in the future). This change simplifies the data processing logic for GAME's scoring and evaluation modules.

2. GAME now supports running multiple evaluators at the same time, for example, the users can specify precision@1 + precision@3 + precision@10 + AUC as evaluators at the same time, and GAME will spit out four evaluation metrics per iteration - given the validation data is specified by the user. Motivation for this feature is stated in #164.

3. Minor issues like #165 is addressed, along with some minor Scala code style fixes - sorry, I just can't help fixing them when seeing those warning in my Intellij.

Issue #144, opened 2016-07-28
Clean up GAME Evaluators


ashelkovnykov:
The `Evaluator` trait should be split into two traits, and it's descendants in the `com.linkedin.photon.ml.evaluation` package need to be separated into two groups: one group for calculating loss during coordinate descent and one group for selecting the best model. Currently there is an overlap.

Ideally, the `Evaluator` used for choosing between models should be configurable.

The `LocalEvaluator` and `AreaUnderROCCurveLocalEvaluator` are part of an upcoming enhancement to Photon-ML, but are not currently used anywhere.

XianXing:
Looks good.
Just want to add that ```LocalEvaluator``` and ```AreaUnderROCCurveLocalEvaluator``` are used in a separate LinkedIn MP, although they are not used anywhere in Photon-ML.

PR #173, opened 2016-10-05
Add sharded evaluators to GAME's evaluation package

XianXing:
This PR introduces the new concept of "sharded evaluators". Similar to the sharded data sets and sharded optimization problems, the sharded evaluators consist of a RDD of local evaluators, with each evaluator responsible for evaluating data grouped by a certain type of id. 

For example, in news feed recommendation problem, sometimes we'd like to evaluate the quality of news feed recommendations per session, where each session can be defined as, for example, a user's browsing and click history within a certain time period. In this case, we can partition (shard) the data by session id and compute some metrics for data from each session, then take the average of metrics computed for each session as the final evaluation metric. Sometimes such "session" based evaluation metric is more useful than a metric computed with all data mixed together, as it removes certain biases in the data (e.g., heavy users vs. light users). However, such "session" based metrics suffer from the assumption that each session is equally important, which may not be true and would potentially lead to also biased conclusions.

The following two concrete examples of sharded evaluators are added in this PR:
1. Sharded precision@K. This one is same as the precision@K added in PR #166, and in this PR we refactored the old precision@K based on the newly designed sharded evaluators. As a recap: sharded precision@K is defined as the mean of precision@Ks computed for each sharded data set (e.g., precision@K computed for each session in the news feed example above).
2. Sharded AUC. where is defined as the mean of AUCs computed for each sharded data set (e.g., AUC computed for each session in the news feed example above). Note that sharded AUC may suffer from numerical instability problem when each sharded data set is dominated by one type of labels (e.g., all negative or all positive labels, see integration tests for sharded AUC for more details), so please use it with caution.

This PR also relates to issue #41 and #144.


photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/ShardedPrecisionAtKEvaluator.scala
```diff
@@ -0,0 +1,41 @@
...
+/**
+ * Sharded version of the precision@k evaluator
+ *
+ * @param k The cut-off rank based on which the precision is computed (precision @ k)
+ * @param idType Type of the id, e.g., documentId or queryId.
+ * @param ids Ids based on which the labels and scores are grouped (sharded) to compute the evaluation metric for each
+ *            shard/group. Such ids can be thought as a recommendation context, e.g. in evaluating the relevance of
+ *            search results of given a query, the id can be the query itself.
+ * @param labelAndOffsetAndWeights a [[RDD]] of (id, (labels, offsets, weights)) pairs
+ * Interface for evaluation implementations at the [[RDD]] level
```
ashelkovnykov: Extra comment on line 29, looks like leftover from Evaluator trait comments.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/EvaluatorType.scala
```diff
@@ -72,9 +67,12 @@ object EvaluatorType {
     case PoissonLoss.name | "POISSONLOSS" => PoissonLoss
     case SmoothedHingeLoss.name | "SMOOTHEDHINGELOSS" => SmoothedHingeLoss
     case SquaredLoss.name | "SQUAREDLOSS" => SquaredLoss
```
XianXing: Yes I agree, and I think the list will grow.

As a side note: here it looks redundant (e.g., ```PoissonLoss.name``` and ```"POISSONLOSS"```) partially because we want to keep it backward incompatible.

photon-ml/src/main/scala/com/linkedin/photon/ml/evaluation/ShardedAreaUnderROCCurveEvaluator.scala
```diff
@@ -0,0 +1,38 @@
...
+
+
+/**
+ * Sharded version of the AUC evaluator
+ * @param idType Type of the id, e.g., documentId or queryId.
+ * @param ids Ids based on which the labels and scores are grouped (sharded) to compute the evaluation metric for each
+ *            shard/group. Such ids can be thought as a recommendation context, e.g. in evaluating the relevance of
+ *            search results of given a query, the id can be the query itself.
+ * @param labelAndOffsetAndWeights a [[RDD]] of (id, (labels, offsets, weights)) pairs
+ * Interface for evaluation implementations at the [[RDD]] level
```
XianXing: These are good comments.

ashelkovnykov:
@joshvfleming Any comments? Otherwise I'm ready to merge

joshvfleming:
LGTM

Issue #124, opened 2016-06-21
Throw an error if the input record contains features with duplicated name.


XianXing:
In function `getFeaturesFromGenericRecord` in `DataProcessingUtils`, we are reading and combining features from different fields for each record.

In some cases, it is possible that two features with same name may appear in different fields of the same record, for example, in an item-to-item recommendation scenario, the same feature (e.g., item category) may appear in both source and destination feature fields, even though they mean different things (e.g., source item's category and destination item's category).

However, currently `getFeaturesFromGenericRecord` converts features with same name but from different fields into the same feature index that is used in both model training and scoring, which may lead to some unexpected model behavior.

In general, it might be useful to throw an error (or warning?) message to let the user aware of such potential problem if more than one features with same name are found while processing the record in `getFeaturesFromGenericRecord`.

XianXing:
@TDDFT suggested one such potential case in job-to-jobs recommendation problem to me.

fastier-li:
That has been fixed for sure. We got a lot of feedback when we turned it on :-) But in the end, we prevailed and people had to clean up their pipelines :-)

PR #179, opened 2016-10-10
Add DataReader, AvroDataReader, and integrate with GAME drivers

joshvfleming:
This is an incremental step towards the new client API. This change introduces a new interface -- **DataReader**, whose concrete implementations are responsible for reading data files of various formats and producing a DataFrame (and optionally an IndexMap). 

| DataReader |
| --- |
| `readMerged(path: String, featureColumnMap: Map[String, Set[String]]): (DataFrame, Map[String, IndexMapLoader])` |
| `readMerged(path: String, indexMapLoaders: Map[String, IndexMapLoader], featureColumnMap: Map[String, Set[String]]): DataFrame` |
| `read(path: String): (DataFrame, IndexMapLoader)` |
| `read(path: String, indexMapLoaders: Map[String, IndexMapLoader]): DataFrame` |
| `readMerged(paths: Seq[String], featureColumnMap: Map[String, Set[String]]): (DataFrame, Map[String, IndexMapLoader])` |
| `readMerged(paths: Seq[String], indexMapLoaders: Map[String, IndexMapLoader], featureColumnMap: Map[String, Set[String]]): DataFrame` |
| `read(paths: Seq[String]): (DataFrame, IndexMapLoader)` |
| `read(paths: Seq[String], indexMapLoaders: Map[String, IndexMapLoader]): DataFrame` |

The first such implementation is **AvroDataReader**, which has been integrated here into the GAME training and scoring drivers to replace the legacy Avro code. The current approach is to provide a converter that transforms a DataFrame into a GAMEDataSet, but eventually the GAME and Photon api will consume DataFrame / Dataset directly.

**Note:** This change requires Spark 1.6+, and has been tested on the grid with 1.6.1.
### Performance Comparison

The following documents a performance comparison on an identical proprietary training set before and after this PR.

| Code | AUC | Runtime |
| --- | --- | --- |
| Before | 0.7796065687206302 | 39.55 m |
| After | 0.7801638770709362 | 40.38 m |
### Fixed effect model

| Code | Count | Mean | Variance | L2 norm |
| --- | --- | --- | --- | --- |
| Before | 22468 | 1.985378978052645E-4 | 0.001589526313337537 | 5.976016508476237 |
| After | 22468 | -1.82578704936172E-4 | 0.001650688565248512 | 6.089890718711416 |
### Random effect model 1

| Code | Count | Mean | Std Dev | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Before | 2370903 | 0.317850 | 0.203167 | 3.132743 | 0.000040 |
| After | 2370903 | 0.317734 | 0.203154 | 3.137519 | 0.000037 |
### Random effect model 2

| Code | Count | Mean | Std Dev | Max | Min |
| --- | --- | --- | --- | --- | --- |
| Before | 967254 | 0.136670 | 0.154361 | 2.722700 | 0.000027 |
| After | 967254 | 0.136611 | 0.154208 | 2.719635 | 0.000027 |


photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+          case Row(featureKey: String, value: Double) =>
+            if (featureMap.contains(featureKey)) {
+              Some(featureMap.getIndex(featureKey) -> value)
+            } else {
+              None
+            }
+        }
+
+        // Add intercept if it exists in the feature map
+        val featuresWithIntercept = if (featureMap.contains(AvroDataReader.intercept._1)) {
```
XianXing: Here we implicitly say that if the intercept exists in the feature map, then it will be added to the features vector for model training/scoring later on.

We may want to make this explicit to the users, and we may want to make sure that the users at least are able to config whether the intercept should be added to the index map or not.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+          features :+ (featureMap.getIndex(AvroDataReader.intercept._1) -> AvroDataReader.intercept._2)
+        } else {
+          features
+        }
+
+        val (indices, values) = featuresWithIntercept.distinct.sorted.unzip
```
XianXing: What if we got features with duplicated names? (#124).

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+        val vector = new SparseVector(featureMap.featureDimension, indices.toArray, values.toArray)
+
+        // Add the column
+        Row.fromSeq(row.toSeq :+ vector)
```
XianXing: Can you briefly explain why we combine the original row (I suppose it's (key,value) pairs?) with the vector (sparse vector from Breeze) together?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
```diff
@@ -0,0 +1,164 @@
...
+   */
+  protected[data] def getGameDatumFromRow(
+      row: Row,
+      featureShards: Set[String],
+      idTypeSet: Set[String],
+      isResponseRequired: Boolean): GameDatum = {
+
+    val featureShardContainer = featureShards.map { shardId =>
+      val features = row.getAs[SparkSparseVector](shardId)
+      (shardId, new SparseVector(features.indices, features.values, features.size))
```
XianXing: Nitpick, let's be explicit about the type of ```SparseVector``` for readability:
 
```new SparseVector[Double](features.indices, features.values, features.size)```

File: photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala [hunk 5]
```diff
@@ -23,7 +23,6 @@ import org.apache.spark.rdd.RDD
 import com.linkedin.photon.ml.cli.game.GAMEDriver
 import com.linkedin.photon.ml.algorithm._
 import com.linkedin.photon.ml.avro.AvroUtils
```
XianXing: Unused import statement.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -234,17 +238,24 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
...
-    val partitioner = new LongHashPartitioner(records.partitions.length)
+    val randomEffectIdTypeSet = randomEffectDataConfigurations.values.map(_.randomEffectType).toSet
+
+    val dataReader = new AvroDataReader(sparkContext)
+    val data = dataReader.read(
+      validatingRecordsPath,
+      featureShardIdToFeatureMapLoader.toMap,
+      featureShardIdToFeatureSectionKeysMap)
+
+    val numPartitions = Math.max(data.rdd.partitions.size, minPartitionsForValidation)
```
XianXing: Nitpick: ```size``` -> ```length``` for arrays.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
```diff
@@ -17,10 +17,10 @@ package com.linkedin.photon.ml.cli.game.scoring
 import com.linkedin.photon.ml.SparkContextConfiguration
 import com.linkedin.photon.ml.cli.game.GAMEDriver
 import com.linkedin.photon.ml.avro.AvroUtils
```
XianXing: Unused import statement.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+
+    val initTempColMappings = Map.empty[String, String]
+
+    // Merge the columns into new temporary columns. We need to create intermediate temporary columns to correctly
+    // handle cases where destination columns are named the same as source columns. Replacing these in-situ could cause
+    // unexpected results when merging subsequent columns.
+    val (merged, tempColMappings) = featureColumnMap.foldLeft((df, initTempColMappings)) {
+      case ((merged, tempColMappings), (destColumn, sourceColumns)) =>
+        val sourceCols = sourceColumns.map(col(_)).toSeq
+        val newCol = mergeFeatures(array(sourceCols:_*))
```
joshvfleming: It's the second one. Basically the column is materialized according to the udf when we call `merged.withColumn`.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+  protected def mergeColumns(
+      df: DataFrame,
+      featureColumnMap: Map[String, Set[String]]): DataFrame = {
+
+    val initTempColMappings = Map.empty[String, String]
+
+    // Merge the columns into new temporary columns. We need to create intermediate temporary columns to correctly
+    // handle cases where destination columns are named the same as source columns. Replacing these in-situ could cause
+    // unexpected results when merging subsequent columns.
+    val (merged, tempColMappings) = featureColumnMap.foldLeft((df, initTempColMappings)) {
```
joshvfleming: That's right!

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+   */
+  protected def tempColumnName(name: String): String = s"__${name}__${System.currentTimeMillis}__"
+
+  /**
+   * Read a feature tuple from the name, term, and value DataFrame row
+   *
+   * @param row the row from which to read the name, term, value tuple
+   * @return the feature tuple
+   */
+  protected def readFeature = (feature: Row) => {
```
joshvfleming: I've done it this way so I can include the function in the closure on line 133 rather than serializing the entire object. As I write this, though, I realize that I could accomplish the same thing by making this function a member of the companion object. I think I'll do that instead.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/DataReader.scala
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  /**
+   * Reads the files at the given paths into a DataFrame, using the given index map for feature names
+   *
+   * @param paths the paths to the files or folders
+   * @param indexMapLoaders a map of index map loaders, containing one loader for each feature column
+   * @return the loaded and transformed DataFrame
+   */
+  def read(paths: Seq[String], indexMapLoaders: Map[String, IndexMapLoader]): DataFrame =
```
joshvfleming: There should be only one -- the other one accepts only a single `path: String`.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/DataReader.scala
```diff
@@ -0,0 +1,158 @@
...
+abstract class DataReader {
+
+  val defaultFeatureColumn = "features"
+  val defaultFeatureColumnMap = Map(defaultFeatureColumn -> Set(defaultFeatureColumn))
+
+  /**
+   * Reads the file at the given path into a DataFrame, generating a default index map for feature names
+   *
+   * @param path the path to the file or folder
+   * @param featureColumnMap a map that specifies how the feature columns should be merged. The keys specify the name
```
joshvfleming: Yep, I'll do this.

photon-ml/build.gradle
```diff
@@ -32,11 +32,11 @@ dependencies {
...
 
-  compile("org.apache.spark:spark-core$scalaSuffix:1.4.0")
-  compile("org.apache.spark:spark-mllib$scalaSuffix:1.4.0")
+  compile("org.apache.spark:spark-core$scalaSuffix:1.6.1")
+  compile("org.apache.spark:spark-mllib$scalaSuffix:1.6.1")
   compile("com.databricks:spark-avro$scalaSuffix:2.0.1")
-  compile("org.apache.spark:spark-sql$scalaSuffix:1.4.0")
-  compile("org.apache.spark:spark-catalyst$scalaSuffix:1.4.0")
+  compile("org.apache.spark:spark-sql$scalaSuffix:1.6.1")
+  compile("org.apache.spark:spark-catalyst$scalaSuffix:1.6.1")
```
joshvfleming: I haven't tested it for 2.0 yet, but this follows roughly the same progression that spark made from DataFrame in 1.6 toward Dataset in 2.0. This is important to keep in mind, though, and we have a couple other small 2.0 incompatibilities in the code also.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+   * @param indexMapLoader the index map loader for this column
+   * @return the transformed DataFrame
+   */
+  protected def vectorize(
+      df: DataFrame,
+      columnName: String,
+      indexMapLoader: IndexMapLoader): DataFrame = {
+
+    val vectorized = df.mapPartitions { rows =>
+      val featureMap = indexMapLoader.indexMapForRDD
```
joshvfleming: Yup.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
joshvfleming: If they have identical values, they're condensed into one. Otherwise, they remain as distinct features. Maybe this is the right time to disambiguate duplicate feature names from merged sections?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
joshvfleming: The original row contains columns with information about the data point (e.g. offset, weight, response, (key,value) pairs for the features, etc.). This line adds a new column that contains the feature vector transformed from the (key, value) pairs. So originally the DataFrame looks like this:

| response | weight | offset | features |
| --- | --- | --- | --- |
| 1.0 | 1.0 | 0.0 | [("blah1", "", 1.0), ("blah2", "", 1.0)] |

And after this line, it will look like this:

| response | weight | offset | features | features-tmp-col-12345 |
| --- | --- | --- | --- | --- |
| 1.0 | 1.0 | 0.0 | [("blah1", "", 1.0), ("blah2", "", 2.0)] | \<SparseVector 0=1.0 1=2.0\> |

Backing up a bit. In spark.ml, the convention is to have a single column in the DataFrame that contains feature vectors, rather than having columns for each feature. I haven't dug into this design choice in detail, but I presume it's for reasons of sparsity. I'm following their pattern here because it seems like a sane choice, and it will allow easily switching between spark.ml and Photon-ml.

Thanks for bringing this up, I'll add it to the documentation in the code.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+  /**
+   * Generates a default index map loader by scanning the dataframe
+   *
+   * @param df the source DataFrame
+   * @param columnName the feature column from which to build the index
+   * @return the index map loader
+   */
+  protected def generateIndexMapLoader(df: DataFrame, columnName: String): IndexMapLoader = {
+    val features = df.select(columnName).flatMap { row: Row =>
+      row.getSeq[Row](0).map {
```
joshvfleming: Yes, these columns contain only one value -- the seq of row structs that represent the (key, value) pairs for features.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/DataReader.scala
XianXing: How about functions defined in line 97:
```
  def read(
      paths: Seq[String],
      featureColumnMap: Map[String, Set[String]]): (DataFrame, Map[String, IndexMapLoader]) = {
```
and line 155:
```
def read(paths: Seq[String], indexMapLoaders: Map[String, IndexMapLoader]): DataFrame =
```,

And function defined in line 40 vs. line 81?

File: photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala (hunk 2 above)
Comment (XianXing) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala]: Yes, I feel like it's about time to do something about the duplicate features.

File: photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala (hunk 3 above)
Comment (XianXing) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala]: It makes sense to me, thanks a lot for the detailed explanation!

File: photon-ml/src/main/scala/com/linkedin/photon/ml/data/DataReader.scala (hunk 11 above)
Comment (joshvfleming) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/data/DataReader.scala]: Ah, you're right. It's using the return type to disambiguate the call, but that is a little awkward. I think I'll rename the functions that merge columns to `readMerged`. This makes the combined transformation a little more explicit anyway. Sound reasonable?

Comment (XianXing) [inline photon-ml/src/main/scala/com/linkedin/photon/ml/data/DataReader.scala]: Sounds good.

File: photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala [hunk 16]
```diff
@@ -234,17 +238,24 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
     }
     logger.debug(s"Validating records paths:\n${validatingRecordsPath.mkString("\n")}")
 
-    val records = AvroUtils.readAvroFiles(sparkContext, validatingRecordsPath, minPartitionsForValidation)
-    val recordsWithUniqueId = records.zipWithUniqueId().map(_.swap)
-    val partitioner = new LongHashPartitioner(records.partitions.length)
+    val randomEffectIdTypeSet = randomEffectDataConfigurations.values.map(_.randomEffectType).toSet
```
ashelkovnykov: This is never used

photon-ml/src/integTest/scala/com/linkedin/photon/ml/data/GameConvertersTest.scala
```diff
@@ -0,0 +1,181 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.data
+
+import com.linkedin.photon.ml.test.SparkTestUtils
+
+import org.apache.spark.sql.{DataFrame, Row, SQLContext}
+import org.apache.spark.sql.types.{LongType, MapType, StructField, StructType, StringType}
+import org.testng.annotations.{DataProvider, Test}
```
ashelkovnykov: DataFrame and DataProvider imports not used

photon-ml/src/integTest/scala/com/linkedin/photon/ml/data/GameConvertersTest.scala
```diff
@@ -0,0 +1,181 @@
...
+      row = dataFrame.head,
+      featureShards = Set(),
+      idTypeSet = Set(),
+      isResponseRequired = false)
+
+    assertEquals(gameDatumWithoutResponse.idTypeToValueMap.get(GameConverters.FieldNames.UID), Some(uid))
+  }
+
+  @Test(expectedExceptions = Array(classOf[IllegalArgumentException]))
+  def testGetGameDatumFromGenericRecordWithNoResponse(): Unit =
```
ashelkovnykov: Test name doesn't match now that we're not reading from GenericRecord

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
ashelkovnykov: If the features vector (the one stored as an actual Vector) is given a temp column name, how is it found later by GameConverters (which uses the shard name but expects a Vector)?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,307 @@
...
+          tempColMappings ++ Map(destColumn -> tempColName))
+    }
+
+    // Remove the source feature columns
+    val scrubbed = dropColumns(merged, featureColumnMap.values.flatten.toSeq.distinct)
+
+    // Rename the temporary columns
+    tempColMappings.foldLeft(scrubbed) { case (data, (destColumn, tempColName)) =>
+      data.drop(destColumn)
+        .withColumnRenamed(tempColName, destColumn)
```
joshvfleming: This drops a source column with the same name as the destination column. Then it renames the destination column with that name.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
joshvfleming: We rename it to the "real" name after processing. See line 185/186.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
ashelkovnykov: Yes - sorry, I misread that

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
joshvfleming: Spark vector representations don't take type arguments.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
joshvfleming: Spark vector representations don't take type arguments.

PR #186, opened 2016-10-20
Remove VectorsWrapper around private Spark api

joshvfleming:
This replaces the wrapper with equivalent code.


photon-ml/src/main/scala/com/linkedin/photon/ml/util/VectorUtils.scala
```diff
@@ -16,6 +16,7 @@ package com.linkedin.photon.ml.util
 
 
 import breeze.linalg.{DenseVector, SparseVector, Vector}
+import org.apache.spark.mllib.linalg.{DenseVector => SDV, SparseVector => SSV, Vector => SV}
```
ferd36: "SV"? In casual reading without seeing this import line, I would have thought it meant "SparseVector" already.

joshvfleming:
@pogil No, but this is an emergency fix so that we have a workaround. I'll add tests next week.

joshvfleming:
@XianXing They added a new rule to our internal byte-ray tool that blocked Azkaban uploads for any jar containing Spark package names. This PR was an emergency workaround to unblock our users.

Also, it's generally not a good practice to have dependencies on private APIs, because they can change at any time. The wrapper was essentially exposing a private Spark API. This change introduces a small amount of copy-paste, but it has the benefit of using only public APIs.

XianXing:
Sounds reasonable @joshvfleming.

+ @TDDFT for more context.

TDDFT:
LGTM

Issue #181, opened 2016-10-12
Track best model in GAME


ferd36:
In the training loop for CoordinateDescent, GAME currently returns the model trained at the last iteration. However, sometimes models at intermediate iterations actually perform better. CoordinateDescent should return the best model that appeared in any iteration, rather than simply returning the final one.

PR #191, opened 2016-10-24
Fix for issue #181: track best model in CoordinateDescent

ferd36:
Let me know what you think. I've added a simple unit test that passes.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
@@ -14,14 +14,13 @@
  */
 package com.linkedin.photon.ml.algorithm
 
-import org.apache.spark.rdd.RDD
-
 import com.linkedin.photon.ml.constants.{MathConst, StorageLevel}
 import com.linkedin.photon.ml.data.{DataSet, GameDatum}
 import com.linkedin.photon.ml.evaluation.Evaluator
 import com.linkedin.photon.ml.model.GAMEModel
 import com.linkedin.photon.ml.util.{ObjectiveFunctionValue, PhotonLogger, Timer}
 import com.linkedin.photon.ml.{BroadcastLike, RDDLike}
+import org.apache.spark.rdd.RDD
 
 /**
  * Coordinate descent implementation
@@ -49,7 +48,9 @@ class CoordinateDescent(
    * @return A trained GAME model
    */
   def run(numIterations: Int, seed: Long = MathConst.RANDOM_SEED): GAMEModel = {
+
     val initializedModelContainer = coordinates.map { case (coordinateId, coordinate) =>
+
       val initializedModel = coordinate.initializeModel(seed)
       initializedModel match {
         case rddLike: RDDLike =>
@@ -68,11 +69,15 @@ class CoordinateDescent(
   }
 
   /**
-   * Run coordinate descent
+   * Run coordinate descent. This function optimizes the model w.r.t. the objective
+   * function. Optionally, it also evaluates the model on the validation data set using one or
+   * more validation functions. In that case, the output is the model which yielded the
+   * best evaluation on the validation data set w.r.t. the primary evaluation function.
+   * Otherwise, it's simply the trained model.
    *
    * @param numIterations Number of iterations
    * @param gameModel The initial GAME model
-   * @return Trained GAME model
+   * @return The best GAME model (see above for exact meaning of "best")
    */
   def run(numIterations: Int, gameModel: GAMEModel): GAMEModel = {
 
@@ -80,55 +85,68 @@ class CoordinateDescent(
       require(gameModel.getModel(coordinateId).isDefined,
         s"Model with coordinateId $coordinateId is expected but not found from the initial GAME model!")
     }
+
     var updatedGAMEModel = gameModel
 
     // Initialize the training scores
     var updatedScoresContainer = coordinates.map { case (coordinateId, coordinate) =>
       val updatedScores = coordinate.score(updatedGAMEModel.getModel(coordinateId).get)
-      updatedScores.setName(s"Initialized training scores with coordinateId $coordinateId")
-          .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+      updatedScores
+        .setName(s"Initialized training scores with coordinateId $coordinateId")
+        .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+
       (coordinateId, updatedScores)
     }.toMap
 
     // Initialize the validating scores
     var validatingScoresContainerOption = validatingDataAndEvaluatorsOption.map { case (validatingData, _) =>
-      val validatingScoresContainer = coordinates.map { case (coordinateId, _) =>
+      coordinates.map { case (coordinateId, _) =>
         val updatedModel = updatedGAMEModel.getModel(coordinateId).get
         val validatingScores = updatedModel.score(validatingData)
-        validatingScores.setName(s"Initialized validating scores with coordinateId $coordinateId")
-            .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+        validatingScores
+          .setName(s"Initialized validating scores with coordinateId $coordinateId")
+          .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+
         (coordinateId, validatingScores)
       }.toMap
-      validatingScoresContainer
     }
 
     // Initialize the regularization term value
-    var regularizationTermValueContainer = coordinates
-      .map { case (coordinateId, coordinate) =>
+    var regularizationTermValueContainer = coordinates.map { case (coordinateId, coordinate) =>
         val updatedModel = updatedGAMEModel.getModel(coordinateId).get
+
         (coordinateId, coordinate.computeRegularizationTermValue(updatedModel))
-      }
-      .toMap
+      }.toMap
+
+    // This will track the "best" model according to the first evaluation function chosen by the user.
+    // If the user did not specify any evaluation function, this var will be None.
+    // NOTE: these two variables are updated by comparing *FULL* models, including random effects, i.e.
+    // outside the loop on coordinates.
+    // If we allowed the "best" model to be selected inside the loop over coordinates, the "best"
+    // model could be one that doesn't contain some random effects.
+    var bestModel: Option[GAMEModel] = None
+    var bestEval: Option[Double] = None
 
     for (iteration <- 0 until numIterations) {
+
       val iterationTimer = Timer.start()
       logger.debug(s"Iteration $iteration of coordinate descent starts...\n")
 
-      coordinates.foreach { case (coordinateId, coordinate) =>
+      val currentEvaluation = coordinates.map { case (coordinateId, coordinate) =>
+
         val coordinateTimer = Timer.start()
         logger.debug(s"Start to update coordinate with ID $coordinateId (${coordinate.getClass})")
 
         // Update the model
         val modelUpdatingTimer = Timer.start()
-
         val oldModel = updatedGAMEModel.getModel(coordinateId).get
         val (updatedModel, optimizationTrackerOption) = if (updatedScoresContainer.keys.size > 1) {
-          // If there are other coordinates, collect their scores into a partial score and optimize
+          // If there are multiple coordinates,
+          // collect scores for previously optimized coordinates into a partial score and update
           val partialScore = updatedScoresContainer.filterKeys(_ != coordinateId).values.reduce(_ + _)
           coordinate.updateModel(oldModel, partialScore)
-
         } else {
-          // Otherwise, just optimize
+          // Otherwise, just update (very first coordinate)
           coordinate.updateModel(oldModel)
         }
 
@@ -138,9 +156,9 @@ class CoordinateDescent(
               .setName(s"Updated model with coordinateId $coordinateId at iteration $iteration")
               .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
               .materialize()
-
           case _ =>
         }
+        // We need to split out the following 2 match expressions: [[FactoredRandomEffectModel]] matches both
         oldModel match {
           case broadcastLike: BroadcastLike => broadcastLike.unpersistBroadcast()
           case _ =>
@@ -149,32 +167,35 @@ class CoordinateDescent(
           case rddLike: RDDLike => rddLike.unpersistRDD()
           case _ =>
         }
+
         updatedGAMEModel = updatedGAMEModel.updateModel(coordinateId, updatedModel)
 
         // Summarize the current progress
         logger.info(s"Finished training the model in coordinate $coordinateId, " +
             s"time elapsed: ${modelUpdatingTimer.stop().durationSeconds} (s).")
-        optimizationTrackerOption.foreach { optimizationTracker =>
-          logger.debug(s"OptimizationTracker:\n${optimizationTracker.toSummaryString}")
-          optimizationTracker match {
-            case rddLike: RDDLike => rddLike.unpersistRDD()
-            case _ =>
-          }
-        }
         logger.debug(s"Summary of the learned model:\n${updatedModel.toSummaryString}")
+
+        optimizationTrackerOption.foreach {
+          optimizationTracker => logger.debug(s"OptimizationTracker:\n${optimizationTracker.toSummaryString}")
+            optimizationTrackerOption match {
+              case rddLike: RDDLike => rddLike.unpersistRDD()
+              case _ =>
+            }
+        }
 
         // Update the training score
         val updatedScores = coordinate.score(updatedModel)
-        updatedScores.setName(s"Updated training scores with key $coordinateId at iteration $iteration")
-            .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
-            .materialize()
+        updatedScores
+          .setName(s"Updated training scores with key $coordinateId at iteration $iteration")
+          .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+          .materialize()
         updatedScoresContainer(coordinateId).unpersistRDD()
         updatedScoresContainer = updatedScoresContainer.updated(coordinateId, updatedScores)
 
         // Update the regularization term value
         val updatedRegularizationTermValue = coordinate.computeRegularizationTermValue(updatedModel)
         regularizationTermValueContainer =
-            regularizationTermValueContainer.updated(coordinateId, updatedRegularizationTermValue)
+          regularizationTermValueContainer.updated(coordinateId, updatedRegularizationTermValue)
         // Compute the training objective function value
         val fullScore = updatedScoresContainer.values.reduce(_ + _)
         val lossFunctionValue = trainingLossFunctionEvaluator.evaluate(fullScore.scores)
@@ -184,35 +205,60 @@ class CoordinateDescent(
             s"iteration $iteration is:\n$objectiveFunctionValue")
 
         // Update the validating score and evaluate the updated model on the validating data
-        validatingScoresContainerOption = validatingDataAndEvaluatorsOption.map { case (validatingData, evaluators) =>
-          var validatingScoresContainer = validatingScoresContainerOption.get
-          val validatingScores = updatedModel
-            .score(validatingData)
-            .setName(s"Updated validating scores with coordinateId $coordinateId")
-            .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
-            .materialize()
+        val currentEvaluation =
+        validatingDataAndEvaluatorsOption.map { case (validatingData, evaluators) =>
+
+          val validationTimer = Timer.start()
+          val validatingScoresContainer = validatingScoresContainerOption.get
+          validatingScoresContainer.updated(coordinateId,
+            updatedModel
+              .score(validatingData)
+              .setName(s"Updated validating scores with coordinateId $coordinateId")
+              .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+              .materialize()
+          )
           validatingScoresContainer(coordinateId).unpersistRDD()
-          validatingScoresContainer = validatingScoresContainer.updated(coordinateId, validatingScores)
+          validatingScoresContainerOption = Some(validatingScoresContainer)
           val fullScore = validatingScoresContainer.values.reduce(_ + _)
-          evaluators.foreach { evaluator =>
-            val validationTimer = Timer.start()
-            val evaluationMetric = evaluator.evaluate(fullScore.scores)
-            logger.debug(s"Finished validating the model, time elapsed: ${validationTimer.stop().durationSeconds} (s).")
+
+          val (firstEvaluator, currentEvaluation) = evaluators.map { evaluator =>
+            val evaluationTimer = Timer.start()
+            val evaluation = evaluator.evaluate(fullScore.scores)
+            logger.debug(s"Finished score evaluation, time elapsed: ${evaluationTimer.stop().durationSeconds} (s).")
             logger.info(s"Evaluation metric computed with ${evaluator.getEvaluatorName} after updating " +
-                s"coordinateId $coordinateId at iteration $iteration is $evaluationMetric")
+              s"coordinateId $coordinateId at iteration $iteration is $evaluation")
+
+            (evaluator, evaluation)
+          }.head // we use only the first evaluator to select the "best" model
+
+          logger.debug(s"Finished validating model, time elapsed: ${validationTimer.stop().durationSeconds} (s).")
+
+          (firstEvaluator, currentEvaluation)
+        } // validatingDataAndEvaluatorsOption.map
+
+        logger.info(s"Updating coordinate $coordinateId finished, time elapsed: " +
+            s"${coordinateTimer.stop().durationSeconds} (s)\n")
+
+        currentEvaluation // the first evaluator of the last coordinate evaluates the full model
+      }.last // end each coordinate
+
+      currentEvaluation match {
+        case Some((firstEvaluator: Evaluator, currentEval: Double)) =>
+          if (bestEval.isEmpty || firstEvaluator.betterThan(currentEval, bestEval.get)) {
+            bestEval = Some(currentEval)
+            bestModel = Some(updatedGAMEModel)
+            logger.debug(s"Found better GAME model, with evaluation: $currentEval")
           }
-          validatingScoresContainer
+
+        case _ =>
+          logger.debug("No evaulator specified to select best model")
         }
 
-        logger.info(
-          s"Updating coordinate $coordinateId finished, time elapsed: ${coordinateTimer.stop().durationSeconds} (s)\n")
-      }
+      logger.info(s"Iteration $iteration of coordinate descent finished, time elapsed: " +
+          s"${iterationTimer.stop().durationSeconds} (s)\n\n")
 
-      logger.info(
-        s"Iteration $iteration of coordinate descent finished, time elapsed: " +
-        s"${iterationTimer.stop().durationSeconds} (s)\n\n")
-    }
+    } // end iterations
 
-    updatedGAMEModel
-  }
+    bestModel.getOrElse(updatedGAMEModel)
+  } // end run
 }
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
@@ -20,9 +20,9 @@ import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 
-import com.linkedin.photon.ml.cli.game.GAMEDriver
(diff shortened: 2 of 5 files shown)


ashelkovnykov:
I have comments, will post tomorrow morning

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -178,35 +189,47 @@ class CoordinateDescent(
...
+        // if there are evaluators defined
+        val currentEval = validatingDataAndEvaluatorsOption.map {
+          case (validatingData, evaluators) =>
+            validatingScoresContainerOption.get.updated(coordinateId,
+              updatedModel
+                .score(validatingData)
+                .setName(s"Updated validating scores for coordinateId $coordinateId")
+                .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+                .materialize()
+            )
```
joshvfleming: This looks like we're not really updating `validatingScoresContainerOption` here. We call `updated`, but then drop the updated value completely. This will cause subsequent coordinates to compute an incorrect ```fullScore```.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -415,36 +416,56 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
...
+
+    validationDataAndEvaluatorsOption.map {
+      case ((data, evaluators)) =>
+
+        val evaluator = evaluators.head
+
+        gameModels
+          .map { case (modelConfig, model) => (modelConfig, model, model.score(data).scores) }
+          .map { case (modelConfig, model, scores) => (modelConfig, model, evaluator.evaluate(scores))}
+          .reduce { (tuple1, tuple2) => if (evaluator.betterThan(tuple1._3, tuple2._3)) tuple1 else tuple2 }
```
joshvfleming: Need more descriptive var names.

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
```diff
@@ -76,19 +99,33 @@ class CoordinateDescentTest {
...
       (coordinates zip models).map { case ((coordinateId, coordinate), model) =>
-        verify(coordinate, times(numIterations)).updateModel(model, score)
+        verify(coordinate, times(iterationCount)).updateModel(model, score)
+      }
+    }
+
+    // Verify the calls to the validation evaluator(s), if any
+    validationDataAndEvaluators.map { case (data, evaluators) =>
+      evaluators.map { case (evaluator) =>
+        verify(evaluator, times(iterationCount * coordinateCount)).evaluate(modelScores)
```
joshvfleming: Any way to verify that we select a "previous" model rather than just the last one in the case that it evaluates better?

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: Proposing new names for tuple1 and tuple2. Let me know what you think.

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
ferd36: I'll try to make that happen by controlling the mocks around "betterThan".

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -415,36 +416,59 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
...
+  : Option[(String, GAMEModel, Double)] = {
+
+    validationDataAndEvaluatorsOption.map {
+      case ((data, evaluators)) =>
+
+        val evaluator = evaluators.head
+
+        gameModels
+          .map { case (modelConfig, model) => (modelConfig, model, model.score(data).scores) }
+          .map { case (modelConfig, model, scores) => (modelConfig, model, evaluator.evaluate(scores))}
```
ashelkovnykov: Why not collapse it into `evaluator.evaluate(model.score(data).scores)`?

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -514,18 +538,22 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
 
     trainingDataSet.foreach { case (_, rddLike: RDDLike) => rddLike.unpersistRDD() }
 
+    timer.start()
+    val bestModel = selectBestModel(gameModelsMap, validatingDataAndEvaluatorsOption)
+    timer.stop()
+    logger.info(s"Time spent in model evaluation: ${timer.durationSeconds} (s)\n")
```
ashelkovnykov: Doesn't fit in with the other log statements. Should say something like:
```
Time elapsed after selecting best model:
```

Also, this block should be within the `if (modelOutputMode != ...)` statement.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -415,36 +416,59 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
...
+    gameModelsMap: Map[String, GAMEModel],
+    bestModel: Option[(String, GAMEModel, Double)]) {
 
     // Write the best model to HDFS
-    validatingDataAndEvaluatorOption match {
-      case Some((validatingData, evaluator)) =>
-        val (bestModelConfig, evaluationResult) = gameModelsMap.mapValues(_.score(validatingData).scores)
-          .mapValues(evaluator.evaluate)
-          .reduce((result1, result2) => if (evaluator.betterThan(result1._2, result2._2)) result1 else result2)
+    if (bestModel.isDefined) {
```
ashelkovnykov: The preferred way to do this is:
```
bestModel match {
  case (modelConfig, model, metric) =>
    ...

  case None =>
    logger.info("")
}
```

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
```diff
@@ -14,31 +14,43 @@
...
-
+import org.apache.spark.rdd.RDD
 import org.mockito.Matchers
 import org.mockito.Mockito._
 import org.testng.annotations.{DataProvider, Test}
 
 /**
-  * Tests for the CoordinateDescent implementation
-  */
+ * Unit tests for the CoordinateDescent
```
ashelkovnykov: Extra word "the"

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
ashelkovnykov: I think that this test should be mostly reverted and the new changes tested in a new test method. The new test method would mock `Evaluator.evaluate` to return different metrics depending on the input mocked KeyValueScore objects.

I'm thinking 3 iterations, and 4 mocked KeyValueScores objects (base + one per iteration) that each point to the next in the chain. In addition, 2 mocked Evaluators: the first ranks the second model the highest, and the second ranks the first model the highest. We should check that we get back the second model, as either the first or third would be incorrect (first means we're using wrong Evaluator, third means we're not using either Evaluator).

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -67,11 +67,14 @@ class CoordinateDescent(
   }
 
   /**
-   * Run coordinate descent
+   * Run coordinate descent. This function optimizes the model w.r.t. to the loss function, and also, optionally,
+   * evaluates the model on the validation data set, using one or more validation functions. This function returns
+   * the model that yielded the best evaluation on the validation data set, using the first evaluation function
+   * specified by the user, or the trained model.
```
ferd36: Thanks! I'll take your version.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -67,11 +67,14 @@ class CoordinateDescent(
...
-   * Run coordinate descent
+   * Run coordinate descent. This function optimizes the model w.r.t. to the loss function, and also, optionally,
+   * evaluates the model on the validation data set, using one or more validation functions. This function returns
+   * the model that yielded the best evaluation on the validation data set, using the first evaluation function
+   * specified by the user, if the user specified evaluation functions, or the trained model.
    *
    * @param numIterations Number of iterations
    * @param gameModel The initial GAME model
-   * @return Trained GAME model
+   * @return The best GAME model for first evaluation function, if present, or the trained GAME model
```
ferd36: OK, will shorten.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -84,34 +87,40 @@ class CoordinateDescent(
     // Initialize the training scores
     var updatedScoresContainer = coordinates.map { case (coordinateId, coordinate) =>
       val updatedScores = coordinate.score(updatedGAMEModel.getModel(coordinateId).get)
-      updatedScores.setName(s"Initialized training scores with coordinateId $coordinateId")
-          .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+      updatedScores
+        .setName(s"Initialized training scores with coordinateId $coordinateId")
+        .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
```
ferd36: OK - the photon style is on my radar screen, but I haven't started on it yet.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -177,36 +188,48 @@ class CoordinateDescent(
...
+          )
           validatingScoresContainer(coordinateId).unpersistRDD()
-          validatingScoresContainer = validatingScoresContainer.updated(coordinateId, validatingScores)
           val fullScore = validatingScoresContainer.values.reduce(_ + _)
-          evaluators.foreach { evaluator =>
+          evaluators.map { evaluator =>
             val validationTimer = Timer.start()
             val evaluationMetric = evaluator.evaluate(fullScore.scores)
-            logger.debug(s"Finished validating the model, time elapsed: ${validationTimer.stop().durationSeconds} (s).")
+            logger.debug(s"Finished model validation, time elapsed: ${validationTimer.stop().durationSeconds} (s).")
```
ferd36: More descriptive is usually good. But you seem to mention "more detailed" (with new timers) rather than "more descriptive"?

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -177,36 +188,48 @@ class CoordinateDescent(
...
-      logger.info(
-        s"Iteration $iteration of coordinate descent finished, time elapsed: " +
-        s"${iterationTimer.stop().durationSeconds} (s)\n\n")
-    }
+        // The "best" model is whichever one optimizes the value of the first evaluation function
+        if (bestEval.isEmpty || currentEval.get > bestEval.get) {
+          bestEval = currentEval
+          bestModel = Some(updatedGAMEModel)
+          logger.debug(s"Found better GAME model, with evaluation: $currentEval")
+        }
```
ferd36: Looks good - I'll incorporate.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -286,8 +284,7 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
             case POISSON_REGRESSION =>
               new PoissonLossEvaluator(validatingLabelsAndOffsetsAndWeights)
             case _ =>
-              throw new UnsupportedOperationException(s"Task type: $taskType is not supported to create validating " +
-                  s"evaluator")
+              throw new UnsupportedOperationException(s"Task type: $taskType does not have a validation evaluator")
```
ferd36: Will fix.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -400,8 +397,12 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
         }
         Pair[String, Coordinate[_ <: DataSet[_], _ <: Coordinate[_, _]]](coordinateId, coordinate)
       }
-      val coordinateDescent = new CoordinateDescent(coordinates, trainingEvaluator, validatingDataAndEvaluatorsOption,
-        logger)
+
+      val coordinateDescent =
+        new CoordinateDescent(coordinates, trainingEvaluator, validatingDataAndEvaluatorsOption, logger)
```
ferd36: Will fix.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: Not sure I follow?

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: Will fix.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: Will fix.

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
ferd36: Will fix.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
joshvfleming: This change would make me a little nervous, because it gives the validating evaluator a very active role in training optimization. If we wait until after all the coordinates are updated, and then evaluate the model, it would be more of a model selection optimization. If we do it in the middle of an iteration like this, we're greedily selecting coordinate models based on the the validating metric -- it's not obvious to me that this is correct. At the very least it's building some amount of validation information into model training.

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
ferd36: We need a way to identify that the result is correct though - i.e. make sure that the returned **mock** model corresponds to the best evaluation.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
ferd36: OK - tried to add a newline consistently throughout. We can take that up later when we iron out the photon style guide too.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -84,34 +87,40 @@ class CoordinateDescent(
...
+      updatedScores
+        .setName(s"Initialized training scores with coordinateId $coordinateId")
+        .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
       (coordinateId, updatedScores)
     }.toMap
 
     // Initialize the validating scores
-    var validatingScoresContainerOption = validatingDataAndEvaluatorsOption.map { case (validatingData, _) =>
-      val validatingScoresContainer = coordinates.map { case (coordinateId, _) =>
+    val validatingScoresContainerOption = validatingDataAndEvaluatorsOption.map { case (validatingData, _) =>
```
ferd36: Fixed.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
ashelkovnykov: Hmm - that's a good point. I definitely lost sight of that. It shouldn't be selecting coordinate models in the middle of an iteration, but do you object to recording mid-iteration GAME models as the best one as well? I'm asking because you bring up waiting until all coordinates in an outer-iteration have been updated.

joshvfleming: Yeah, I think it's okay if we're not leaking information from the validation set into the training iteration.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -84,34 +87,40 @@ class CoordinateDescent(
...
         val updatedModel = updatedGAMEModel.getModel(coordinateId).get
         (coordinateId, coordinate.computeRegularizationTermValue(updatedModel))
-      }
-      .toMap
+      }.toMap
+
+    // This will track the "best" model according to the first evaluation function chosen by the user.
+    // If the user did not specify any evaluation function, this var will be None.
+    var bestModel: Option[GAMEModel] = None
+    var bestEval: Option[Double] = None
```
ferd36: Either way is fine with me. I am not sure whether it makes a big difference in terms of readability, correctness or performance. We are probably not taking up the proposed change below though, as it allows information about the validation to percolate into the training. So we still need a bestModel var, and this proposed change feels less attractive to me. I'll punt on it for now.

ashelkovnykov: Yes, my mistake, we still need bestModel

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
ferd36: Refined and added a Timer.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: That style is not at all my preference, but done.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: Got it now! Done. Good idea, it simplifies. Thanks.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ferd36: I am not sure that this should go inside the block - regardless of whether the users want to see an output model, selecting the best model is now part of the training/validation algorithm, IMHO.

photon-ml/src/test/scala/com/linkedin/photon/ml/algorithm/CoordinateDescentTest.scala
```diff
@@ -62,23 +59,24 @@ class CoordinateDescentTest {
...
       doReturn((model, tracker)).when(coordinate).updateModel(model)
       doReturn((model, tracker)).when(coordinate).updateModel(model, score)
+
+      // GAMEModel mock setup
+      doReturn(gameModel).when(gameModel).updateModel(coordinateId, model)
+      doReturn(Some(model)).when(gameModel).getModel(coordinateId)
     }
 
-    // Run coordinate descent
+    // Run coordinate descent - None = no validation
```
ashelkovnykov: Scala has a built in way to include this info without comments. You can explicitly name which parameter you're setting:
`new CoordinateDescent(..., validatingDataAndEvaluatorsOption = None, ...)`
It's recommended when calling methods/initializing classes with hard values.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -84,59 +90,68 @@ class CoordinateDescent(
...
         val coordinateTimer = Timer.start()
         logger.debug(s"Start to update coordinate with ID $coordinateId (${coordinate.getClass})")
 
         // Update the model
         val modelUpdatingTimer = Timer.start()
-
         val oldModel = updatedGAMEModel.getModel(coordinateId).get
         val (updatedModel, optimizationTracker) = if (updatedScoresContainer.keys.size > 1) {
-          // If there are other coordinates, collect their scores into a partial score and optimize
+          // Collect scores for previously optimized coordinates into a partial score and optimize
```
ferd36: Done.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -84,59 +90,68 @@ class CoordinateDescent(
...
-          // Otherwise, just optimize
+          // Otherwise, just optimize (very first coordinate)
           coordinate.updateModel(oldModel)
         }
 
         updatedModel match {
           case rddLike: RDDLike =>
             rddLike.setName(s"Updated model with coordinateId $coordinateId at iteration $iteration")
-                .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL).materialize()
+              .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL).materialize()
```
ferd36: Done.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -84,59 +90,68 @@ class CoordinateDescent(
...
           case _ =>
         }
+        // We need to split out the following 2 match expressions, because e.g. a factored random effect model
```
ferd36: Done.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -178,35 +196,49 @@ class CoordinateDescent(
...
+              .materialize()
+          )
           validatingScoresContainer(coordinateId).unpersistRDD()
-          validatingScoresContainer = validatingScoresContainer.updated(coordinateId, validatingScores)
+          validatingScoresContainerOption = Some(validatingScoresContainer)
           val fullScore = validatingScoresContainer.values.reduce(_ + _)
-          evaluators.foreach { evaluator =>
-            val validationTimer = Timer.start()
+
+          val eval = evaluators.map { evaluator =>
```
ferd36: Much better! Thanks for suggesting this - makes the code much more readable.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -514,18 +551,22 @@ final class Driver(val params: Params, val sparkContext: SparkContext, val logge
 
     trainingDataSet.foreach { case (_, rddLike: RDDLike) => rddLike.unpersistRDD() }
 
+    timer.start()
+    val bestModel = selectBestModel(gameModelsMap, validatingDataAndEvaluatorsOption)
+    timer.stop()
+    logger.info(s"Time elapsed after selecting best model: ${timer.durationSeconds} (s)\n")
+
```
ferd36: IMHO, selecting the best model is akin to tuning hyper-parameters on a validation data set, which logically does not belong with "saving to HDFS". I would like to keep those 2 features/concerns separate. Furthermore, if running the algorithm without "saving to HDFS" at the end is useful to e.g. track what the algo is doing via logs, or to look at an evaluation of the model, by that same token, users could be interested in seeing what happens during model selection. It could also be a diagnosis tool for us.

ashelkovnykov: Fair enough, in that case we should move the output to log of the best model from `saveModelToHDFS` to `selectBestModel`.

ferd36: You got it.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -138,9 +156,9 @@ class CoordinateDescent(
               .setName(s"Updated model with coordinateId $coordinateId at iteration $iteration")
               .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
               .materialize()
-
           case _ =>
         }
+        // We need to split out the following 2 match expressions: [[FactoredRandomEffectModel]] matches both
         oldModel match {
           case broadcastLike: BroadcastLike => broadcastLike.unpersistBroadcast()
           case _ =>
```
ankans: I am not that familiar with Scala, but can you please explain why the match statement is done twice? Couldn't you put broadcastLike and rddLike in the same match statement? Sorry if it is a stupid question, just trying to understand here.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -149,32 +167,35 @@ class CoordinateDescent(
...
-          optimizationTracker match {
-            case rddLike: RDDLike => rddLike.unpersistRDD()
-            case _ =>
-          }
-        }
         logger.debug(s"Summary of the learned model:\n${updatedModel.toSummaryString}")
 
+        optimizationTrackerOption.foreach {
+          optimizationTracker => logger.debug(s"OptimizationTracker:\n${optimizationTracker.toSummaryString}")
+            optimizationTrackerOption match {
```
ankans: Should this be optimizationTracker instead?

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
joshvfleming: @ankans I believe it's done this way because we have some objects that are both `BroadcastLike` and `RDDLike` (i.e., they're not mutually exclusive). If they were in the same match statement, only one of them would be matched. That sounds like a strange mixture of traits to me, though, so we should probably evaluate it and make sure it's right.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
ashelkovnykov: Yes, good catch. Also `optimizationTracker =>` should be on the preceding line. Why was the order of the debug statement and the optimizationTracker printout changed?

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
ferd36: Now I remember that Alex S. mentioned it should be split that way - but I didn't investigate further.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
ferd36: Will fix.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
XianXing: Yep, some models are both ```BroadcastLike``` and ```RDDLike```, for example, the GAME model may contain the fixed-effect model that is ```BroadcastLike```, and the random effect model that is ```RDDLike```.

But I'd encourage everyone to think about whether there is a better design for such traits.

ferd36: XianXing - could you give more details about the rationale for using BroadcastLike sometimes: what does it try to achieve? Is it always the fixed effect model (and only that model) that can be BroadcastLike? Can it be not-BroadcastLinke? When/why would that happen (if it happens)?

joshvfleming: @ferd36 The fixed effect model is Broadcast-like because it's a single model that gets distributed to the executors to be aggregated in parallel. Random effect models are RDD-like because each random-effect problem is solved individually on a single executor, so we map over them like an RDD.

ferd36: Thanks!

PR #194, opened 2016-11-12
Adding generation of default index maps when reading Game model

ferd36:
This is for model conversion - I need an API to load Game models from HDFS without having a feature index. In that case, the feature index is generated on the fly when the model is loaded.

### DIFF
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
@@ -14,10 +14,10 @@
  */
 package com.linkedin.photon.ml.avro
 
-import java.util.Random
-
 import scala.collection.Map
+import scala.util.Random
 
+import breeze.linalg.DenseVector
 import org.apache.hadoop.fs.Path
 import org.testng.Assert._
 import org.testng.annotations.{DataProvider, Test}
@@ -34,64 +34,132 @@ import com.linkedin.photon.ml.util._
 
 class ModelProcessingUtilsTest extends SparkTestUtils with TestTemplateWithTmpDir {
 
+  /**
+   * Test that we can load a simple Game model with fixed and random effects, given a feature index
+   */
   @Test
   def testLoadAndSaveGameModels(): Unit = sparkTest("loadAndSaveRandomEffectModelToHDFS") {
-    // Coefficients parameter
-    val coefficientDimension = 1
-    val coefficients = Coefficients.initializeZeroCoefficients(coefficientDimension)
 
-    // Features
-    val featureShardId = "featureShardId"
-    val featureNameAndTermToIndexMap = Array.tabulate(coefficientDimension)( i => (NameAndTerm("n", "t"), i) ).toMap
-    val indexMapLoader = new DefaultIndexMapLoader(
-      featureNameAndTermToIndexMap.map { case (nat, i) => (Utils.getFeatureKey(nat.name, nat.term), i) }.toMap)
-    indexMapLoader.prepare(sc, null)
-
-    val featureShardIdToFeatureNameAndTermToIndexMapMap = Map(featureShardId -> indexMapLoader)
+    // Features: we have two feature spaces, one for the fix model, and one for the random model
+    // Each model has its own, separate feature space, but feature values can be shared between spaces.
+    // Features shared between spaces have a unique name, but possibly different indices.
+    val numFeaturesPerModel = Map("fixed" -> 10, "random" -> 15)
+    val featureIndexLoaders = numFeaturesPerModel.mapValues { numFeatures =>
+      DefaultIndexMapLoader(sc, (0 until numFeatures).map(i => Utils.getFeatureKey("n" + i, "t")))
+    }
 
     // Fixed effect model
-    val glm: GeneralizedLinearModel = new LogisticRegressionModel(coefficients)
-
-    val coefficientsBC = sc.broadcast(glm)
-    val fixedEffectModel = new FixedEffectModel(coefficientsBC, featureShardId)
+    val fixedEffectCoefficients = Coefficients(numFeaturesPerModel("fixed"))(1,2,5)(1,2,5)
+    val glm: GeneralizedLinearModel = new LogisticRegressionModel(fixedEffectCoefficients)
+    val fixedEffectModel = new FixedEffectModel(sc.broadcast(glm), "fixed")
 
     // Random effect model
-    val numCoefficients = 5
-    val randomEffectType = "randomEffectType"
-    val glmRE: GeneralizedLinearModel = new LogisticRegressionModel(coefficients)
-
-    val glmReRDD = sc.parallelize(Seq.tabulate(numCoefficients)(i => (i.toString, glmRE)))
-    val randomEffectModel = new RandomEffectModel(glmReRDD, randomEffectType, featureShardId)
+    val randomEffectCoefficients = Coefficients(numFeaturesPerModel("random"))(1,5,9)(1,5,9)
+    val glmRE: GeneralizedLinearModel = new LogisticRegressionModel(randomEffectCoefficients)
+    val glmReRDD = sc.parallelize((0 until 2).map(i => (i.toString, glmRE)))
+    val randomEffectModel = new RandomEffectModel(glmReRDD, "randomEffectType", "random")
 
     // GAME model
-    val fixedEffectModelName = "fixed"
-    val randomEffectModelName = "random"
-    val gameModel =
-      new GAMEModel(Map(fixedEffectModelName -> fixedEffectModel, randomEffectModelName -> randomEffectModel))
+    val gameModel = new GAMEModel(Map("fixed" -> fixedEffectModel, "random" -> randomEffectModel))
 
     // Default number of output files
     val numberOfOutputFilesForRandomEffectModel = 2
     val outputDir = getTmpDir
     val outputDirAsPath = new Path(outputDir)
-    ModelProcessingUtils.saveGameModelsToHDFS(gameModel, featureShardIdToFeatureNameAndTermToIndexMapMap, outputDir,
-      numberOfOutputFilesForRandomEffectModel, sc)
+
+    ModelProcessingUtils.saveGameModelsToHDFS(
+      gameModel,
+      featureIndexLoaders,
+      outputDir,
+      numberOfOutputFilesForRandomEffectModel,
+      sc)
 
     val fs = outputDirAsPath.getFileSystem(sc.hadoopConfiguration)
     assertTrue(fs.exists(outputDirAsPath))
 
     // Check if the numberOfOutputFilesForRandomEffectModel parameter is working or not
-    val randomEffectModelCoefficientsDir = new Path(outputDirAsPath,
-      s"$RANDOM_EFFECT/$randomEffectModelName/$COEFFICIENTS")
+    val randomEffectModelCoefficientsDir = new Path(outputDirAsPath, s"$RANDOM_EFFECT/random/$COEFFICIENTS")
     val numRandomEffectModelFiles = fs.listStatus(randomEffectModelCoefficientsDir)
       .count(_.getPath.toString.contains("part"))
     assertEquals(numRandomEffectModelFiles, numberOfOutputFilesForRandomEffectModel,
       s"Expected number of random effect model files: $numberOfOutputFilesForRandomEffectModel, " +
-          s"found: $numRandomEffectModelFiles")
+        s"found: $numRandomEffectModelFiles")
 
     // Check if the models loaded correctly and they are the same as the models saved previously
-    val loadedGameModel = ModelProcessingUtils.loadGameModelFromHDFS(
-      featureShardIdToFeatureNameAndTermToIndexMapMap, outputDir, sc)
-    assertEquals(loadedGameModel, gameModel)
+    // The first value returned is the feature index, which we don't need here
+    val (loadedGameModel, _) = ModelProcessingUtils.loadGameModelFromHDFS(Some(featureIndexLoaders), outputDir, sc)
+    assertEquals(gameModel, loadedGameModel)
+  }
+
+  /**
+   * Test that we can load a model even if we don't have a feature index. In that case, the feature index is
+   * generated as part of loading the model, and it will not directly match the feature index before the save.
+   */
+  @Test
+  def testLoadGameModelsWithoutFeatureIndex(): Unit = sparkTest("loadAndSaveRandomEffectModelToHDFS2") {
+
+    // Features: we have two feature spaces, one for the fix model, and one for the random model
+    // Each model has its own, separate feature space, but feature values can be shared between spaces.
+    // Features shared between spaces have a unique name, but possibly different indices.
+    val numFeaturesPerModel = Map("fixed" -> 10, "random" -> 15)
+    val featureIndexLoaders = numFeaturesPerModel.mapValues { numFeatures =>
+      DefaultIndexMapLoader(sc, (0 until numFeatures).map(i => Utils.getFeatureKey("n" + i, "t")))
+    }.map(identity) // .map(identity) needed because of: https://issues.scala-lang.org/browse/SI-7005
+
+    // Fixed effect model
+    val fixedEffectCoefficients = Coefficients(numFeaturesPerModel("fixed"))(1,2,5)(11,21,51)
+    val glm: GeneralizedLinearModel = new LogisticRegressionModel(fixedEffectCoefficients)
+    val fixedEffectModel = new FixedEffectModel(sc.broadcast(glm), "fixed")
+
+    // Random effect model
+    val randomEffectCoefficients = Coefficients(numFeaturesPerModel("random"))(1,5,7)(101,501,901)
+    val glmRE: GeneralizedLinearModel = new LogisticRegressionModel(randomEffectCoefficients)
+    val glmReRDD = sc.parallelize((0 until 2).map(i => (i.toString, glmRE)))
+    val randomEffectModel = new RandomEffectModel(glmReRDD, "randomEffectType", "random")
+
+    // GAME model
+    val modelDir = getTmpDir
+    val gameModel = new GAMEModel(Map("fixed" -> fixedEffectModel, "random" -> randomEffectModel))
+
+    ModelProcessingUtils.saveGameModelsToHDFS(gameModel, featureIndexLoaders, modelDir, 2, sc)
+
+    // Check if the models loaded correctly and they are the same as the models saved previously
+    // The first value returned is the feature index, which we don't need here
+    val (loadedGameModel, newFeatureIndexLoaders) = ModelProcessingUtils.loadGameModelFromHDFS(None, modelDir, sc)
+
+    // Since the feature index after the load is not the same as before the save, we need to calculate an
+    // invariant that survives the save/load. That invariant is Map[feature value, feature name].
+    def extractNamedFeatures(glm: GeneralizedLinearModel, featureIndexLoader: IndexMapLoader) = {
+      val featureIndex = featureIndexLoader.indexMapForRDD()
+      glm.coefficients.means match {
+        case (vector: breeze.linalg.DenseVector[Double]) =>
+          vector.iterator.map { case (index, value) => (value, featureIndex.getFeatureName(index)) }
+        case (vector: breeze.linalg.SparseVector[Double]) =>
+          // activeIterator to iterate over the non-zeros
+          vector.activeIterator.map { case (index, value) => (value, featureIndex.getFeatureName(index)) }
+      }
+    }
+
+    (gameModel.toMap zip loadedGameModel.toMap) foreach {
+
+      case ((n1: String, m1: FixedEffectModel), (n2: String, m2: FixedEffectModel)) =>
+
+        assertEquals(n1, n2)
+        val fixedMap1 = extractNamedFeatures(m1.model, featureIndexLoaders(n1))
+        val fixedMap2 = extractNamedFeatures(m2.model, newFeatureIndexLoaders(n2))
+        assert(fixedMap1 sameElements fixedMap2)
+
+      case ((n1: String, m1: RandomEffectModel), (n2: String, m2: RandomEffectModel)) =>
+
+        assertEquals(n1, n2)
+        m1.modelsRDD.join(m2.modelsRDD).mapValues{ case (glm1, glm2) =>
+          val fixedMap1 = extractNamedFeatures(glm1, featureIndexLoaders(n1))
+          val fixedMap2 = extractNamedFeatures(glm2, newFeatureIndexLoaders(n2))
+          assert(fixedMap1 sameElements fixedMap2)
+        }.collect
+
+      case _ =>
+    }
   }
 
   @DataProvider
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
@@ -18,21 +18,21 @@ import java.lang.{Double => JDouble}
 import java.util.{List => JList}
 
 import breeze.linalg.{DenseVector, SparseVector, Vector}
+
 import com.linkedin.photon.ml.avro.data.{NameAndTerm, NameAndTermFeatureSetContainer}
-import com.linkedin.photon.ml.avro.generated.{BayesianLinearModelAvro, NameTermValueAvro, LatentFactorAvro}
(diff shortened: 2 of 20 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
```diff
@@ -50,13 +50,18 @@ protected[ml] case class Coefficients(means: Vector[Double], variancesOption: Op
     stringBuilder.toString()
   }
 
+  // TODO: what should we do about the epsilon used in this comparison?
   override def equals(that: Any): Boolean = {
+
+    def cmp(a1: Vector[Double], a2: Vector[Double]): Boolean =
+      (a1.toArray zip a2.toArray).foldLeft(true) { case (v, (m1, m2)) => v & math.abs(m2 - m1) < 1e-6}
```
joshvfleming: More concise, and better to use the tolerance constant than a magic number:

```
(a1.toArray zip a2.toArray).forall {
  case ((m1, m2)) => math.abs(m2 - m1) < MathConst.HIGH_PRECISION_TOLERANCE_THRESHOLD
}
```

Also, we should give this a more descriptive name and move it to VectorUtils.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
```diff
@@ -22,29 +22,37 @@ import org.apache.spark.storage.StorageLevel
...
 /**
- * Representation of the generalized additive mixed effect (GAME) model
+ * Generalized additive mixed effect (GAME) model
  *
- * @param gameModels A (modelName -> model) map representation of the models that consist of the GAME model
+ * @param gameModels A (modelName -> model) map containing the models that make up the complete GAME model
  */
 class GAMEModel(gameModels: Map[String, DatumScoringModel]) extends DatumScoringModel {
 
+  require(gameModels != null)
```
joshvfleming: This isn't needed.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
```diff
@@ -22,29 +22,37 @@ import org.apache.spark.storage.StorageLevel
...
    *
    * @param name The model name
-   * @return An option value containing the value associated with model name `name` in the GAME model, or `None`
-   *         if none exists.
+   * @return An option value containing the value associated with model `name` in the GAME model,
+   *         or `None` if none exists.
    */
-  def getModel(name: String): Option[DatumScoringModel] = gameModels.get(name)
+  def getModel(name: String): Option[DatumScoringModel] = {
+    require(name != null && !name.isEmpty, "Passed null or empty model name")
```
joshvfleming: Don't need the null check. Same comment for the null checks below.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -191,32 +192,39 @@ object AvroUtils {
...
         val value = feature.getValue
-        val index = featureMap.getOrElse(featureKey,
+        val index = featureMap.get.getOrElse(featureKey,
           throw new NoSuchElementException(s"nameAndTerm $featureKey not found in the feature map"))
         indexAndValueArrayBuffer += ((index, value))
+      } else if (featureMap.isEmpty) {
+        val value = feature.getValue
+        val index = counter
+        counter += 1
+        indexAndValueArrayBuffer += ((index, value))
```
joshvfleming: I think you'll need to build and return an actual IndexMap here. Otherwise, you won't be able to reverse-lookup the feature names after loading this model.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
```diff
@@ -22,29 +22,37 @@ import org.apache.spark.storage.StorageLevel
...
    *
    * @param name The name of the model to be updated
-   * @param model The model used to update the previous model
-   * @return The GAME model with updated model
+   * @param model The new model
+   * @return The GAME model with the updated model
    */
   def updateModel(name: String, model: DatumScoringModel): GAMEModel = {
+    require(name != null && !name.isEmpty, "Passed null or empty model name")
+    require(model != null, s"Can't update $name to null")
```
joshvfleming: Don't need to check null.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
ferd36: Done. For the record, persisting models and reloading them does introduce small variations, which I have observed and which is why I needed an epsilon. Previously written tests used "0" for all coefficients, which hid this issue (there was no error on "0").

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
ferd36: OK - I like to spell out pre-conditions, and I tend to think they can be needed later and can help in development, but it's case by case, and I agree that this one might be overkill. I'll retract.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
ferd36: Removed.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
ferd36: Removed.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
ferd36: OK... the difficulty probably being that I'm going to have to pass it back up the call chain.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
joshvfleming: Yep, I'm with you on spelling out preconditions, but use of "null" is highly discouraged in Scala because it's supposed to be the compiler's job to guarantee that references are right. If we have cases where a null is passed here, that's definitely a bug in the calling code and should be removed.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
```diff
@@ -73,7 +73,68 @@ class ModelProcessingUtilsTest extends SparkTestUtils with TestTemplateWithTmpDi
     val numberOfOutputFilesForRandomEffectModel = 2
     val outputDir = getTmpDir
     val outputDirAsPath = new Path(outputDir)
-    ModelProcessingUtils.saveGameModelsToHDFS(gameModel, featureShardIdToFeatureNameAndTermToIndexMapMap, outputDir,
+    ModelProcessingUtils.saveGameModelsToHDFS(gameModel, indexMapLoaders, outputDir,
+      numberOfOutputFilesForRandomEffectModel, sc)
```
ashelkovnykov: If all of the parameters can't fit on one line, they should each be on a separate line.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
```diff
@@ -116,7 +178,7 @@ class ModelProcessingUtilsTest extends SparkTestUtils with TestTemplateWithTmpDi
       val colEffectType = "colEffectType"
 
       // Generate the random matrix
-      val random = new Random(MathConst.RANDOM_SEED)
+      val random = new java.util.Random(MathConst.RANDOM_SEED)
```
ashelkovnykov: I think your IDE replaced the Java `Random` with the Scala `Random` in the imports, but never uses the Scala `Random`.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -191,32 +192,39 @@ object AvroUtils {
...
       val feature = iterator.next()
       val name = feature.getName.toString
       val term = feature.getTerm.toString
       val featureKey = Utils.getFeatureKey(name, term)
-      if (featureMap.contains(featureKey)) {
+      if (featureMap.isDefined && featureMap.get.contains(featureKey)) {
         val value = feature.getValue
-        val index = featureMap.getOrElse(featureKey,
+        val index = featureMap.get.getOrElse(featureKey,
           throw new NoSuchElementException(s"nameAndTerm $featureKey not found in the feature map"))
```
ashelkovnykov: Why do we have this check if we're doing an explicit `contains` check beforehand? As I understand, currently this will never be called and instead we will ignore features not found in the index.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
```diff
@@ -51,12 +52,18 @@ protected[ml] case class Coefficients(means: Vector[Double], variancesOption: Op
...
+      }
+
     that match {
       case other: Coefficients =>
-        val sameMeans = this.means.equals(other.means)
+        val sameMeans = cmp(this.means, other.means)
         val sameVariance =
           (this.variancesOption.isDefined && other.variancesOption.isDefined &&
-              this.variancesOption.get.equals(other.variancesOption.get)) ||
+              cmp(this.variancesOption.get, other.variancesOption.get)) ||
```
joshvfleming: I think we should also give this a more descriptive name and move it to VectorUtils.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
```diff
@@ -22,27 +22,29 @@ import org.apache.spark.storage.StorageLevel
...
    * @param name The model name
-   * @return An option value containing the value associated with model name `name` in the GAME model, or `None`
-   *         if none exists.
+   * @return An option value containing the value associated with model `name` in the GAME model,
+   *         or `None` if none exists.
    */
-  def getModel(name: String): Option[DatumScoringModel] = gameModels.get(name)
+  def getModel(name: String): Option[DatumScoringModel] = {
+    gameModels.get(name)
+  }
```
joshvfleming: Yes, this should be a one-liner.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -191,32 +192,39 @@ object AvroUtils {
...
+      } else if (featureMap.isEmpty) {
+        val value = feature.getValue
+        val index = counter
+        counter += 1
+        indexAndValueArrayBuffer += ((index, value))
       }
     }
 
-    val length = featureMap.featureDimension
+    val length = if (featureMap.isDefined) featureMap.get.featureDimension else counter
```
joshvfleming: +1

photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
joshvfleming: IDEs considered harmful. :-)

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -151,7 +151,8 @@ object AvroUtils {
     numPartitions: Int): NameAndTermFeatureSetContainer = {
 
     val nameAndTermFeatureSets = featureSectionKeys.map { featureSectionKey =>
-      (featureSectionKey, AvroUtils.readNameAndTermSetFromGenericRecords(genericRecords, featureSectionKey, numPartitions))
+      (featureSectionKey, AvroUtils.readNameAndTermSetFromGenericRecords(genericRecords,
+        featureSectionKey, numPartitions))
```
ferd36: For me personally, this style is about as unpleasant and esthetically unsatisfactory as cutting up verses as in: "Serait-ce déjà lui ? - C'est bien à l'escalier / Dérobé.". But I've "fixed" it.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
ferd36: Changed "cmp" to "withinTolerance".

photon-ml/src/main/scala/com/linkedin/photon/ml/model/GAMEModel.scala
ferd36: Fixed.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
ferd36: Updated to the "hanging (" style.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
ferd36: Damn IDE! Fixed :-)

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -242,7 +245,7 @@ object AvroUtils {
    * @return The Avro record that contains the information of the input latent factor
    */
   protected[avro] def convertLatentFactorToLatentFactorAvro(effectId: String, latentFactor: Vector[Double])
-    : LatentFactorAvro = {
+  : LatentFactorAvro = {
```
ashelkovnykov: This should be on the previous line, if it fits. If it doesn't:
```
protected[avro] def convertLatentFactorToLatentFactorAvro(
  effectId: String,
  latentFactor: Vector[Double]): LatentFactorAvro = {
```

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -256,10 +259,55 @@ object AvroUtils {
    * @return The (effectId, latentFactor) pair converted from the input Avro record
    */
   protected[avro] def convertLatentFactorAvroToLatentFactor(latentFactorAvro: LatentFactorAvro)
-    : (String, Vector[Double]) = {
+  : (String, Vector[Double]) = {
```
ashelkovnykov: See above

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/model/ModelProcessingUtils.scala
```diff
@@ -109,7 +118,8 @@ object ModelProcessingUtils {
         val Array(featureShardId) = IOUtils.readStringsFromHDFS(idInfoPath, configuration).toArray
 
         // Load the coefficients
-        val featureNameAndTermToIndexMap = featureShardIdToIndexMapLoader(featureShardId).indexMapForDriver()
+        val indexMapLoaders = featureShardIdToIndexMapLoader.flatMap(_.get(featureShardId))
+        val featureNameAndTermToIndexMap = indexMapLoaders.map(_.indexMapForDriver())
```
ashelkovnykov: It would be helpful for reader understanding to change the names of these vals to reflect that they are of class `Option[_]` (i.e. indexMapLoadersOption).

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
```diff
@@ -256,10 +259,55 @@ object AvroUtils {
...
+   * @return A DefaultIndexMapLoader, which associates a unique integer id to each feature name
+   */
+  protected[avro] def featureMapFromBayesianLinearModelAvro(
+      modelAvros: RDD[BayesianLinearModelAvro]): IndexMapLoader = {
+
+    DefaultIndexMapLoader.makeFeatureIndexLoader(modelAvros.flatMap(modelAvro =>
+      modelAvro.getMeans.map(nameTermValue =>
+        Utils.getFeatureKey(nameTermValue.getName.toString, nameTermValue.getTerm.toString)))
+      .collect
+      .toSet)
```
ashelkovnykov:
```
DefaultIndexMapLoader.makeFeatureIndexLoader(
  modelAvros
    .flatMap(modelAvro =>
      modelAvro
        .getMeans
        .map(nameTermValue => Utils.getFeatureKey(nameTermValue.getName.toString, nameTermValue.getTerm.toString)))
    .collect
    .toSet)
```
And the same above

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
ferd36: Fixed.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/AvroUtils.scala
ferd36: Fixed.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/model/ModelProcessingUtils.scala
ferd36: I have mixed feelings about this one, in particular because it makes the variable names pretty long. In some places in the code, the type name is carried exactly by the variable name, resulting in e.g. ```featureShardIdToFeatureNameAndTermToIndexMapMap```. This is of course an extreme case, but it stems from the same idea, that the variable name should provide type information. Not sure what is best here.

joshvfleming: It's pretty inconsistent in the Spark code. In some places, they prefix the var with "opt", in others they don't specify at all. I lean more toward not specifying, because we don't include var types in the names in any other case, and like @ferd36 said, it makes the var names very long.

ferd36: Yeah - we have type inference, and Scala borrows from Python, which hides types as much as it can, so I'm guessing the philosophy is more to make life easier for devs by figuring out the types for them as much as possible. In my opinion, avoiding type annotations allows to better focus on the logic of the code.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/model/ModelProcessingUtils.scala
```diff
@@ -131,26 +144,33 @@ object ModelProcessingUtils {
...
-        val modelsRDD = loadModelsRDDFromHDFS(modelsRDDInputPath.toString, featureMapLoader, sparkContext)
+        val (featureIndexLoader, modelsRDD) =
+          loadModelsRDDFromHDFS(modelsRDDInputPath.toString, featureMapLoader, sc)
 
-        (name, new RandomEffectModel(modelsRDD, randomEffectType, featureShardId))
+        (name, featureIndexLoader, new RandomEffectModel(modelsRDD, randomEffectType, featureShardId))
       }
     } else {
-      Array[(String, RandomEffectModel)]()
+      Array[(String, DefaultIndexMapLoader, RandomEffectModel)]()
```
joshvfleming: The `IndexMapLoader` type would probably be more appropriate here.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/model/ModelProcessingUtils.scala
```diff
@@ -107,22 +116,26 @@ object ModelProcessingUtils {
...
         val modelPath = new Path(innerPath, COEFFICIENTS)
-        val glm = loadGLMFromHDFS(modelPath.toString, featureNameAndTermToIndexMap, sparkContext)
+        val (featureIndex, glm) = loadGLMFromHDFS(modelPath.toString, featureNameAndTermToIndexMap, sc)
 
-        (name, new FixedEffectModel(sparkContext.broadcast(glm), featureShardId))
+        (name, featureIndex, new FixedEffectModel(sc.broadcast(glm), featureShardId))
       }
     } else {
-      Array[(String, FixedEffectModel)]()
+      Array[(String, DefaultIndexMapLoader, FixedEffectModel)]()
```
joshvfleming: The `IndexMapLoader` type would probably be more appropriate here.

ferd36: You are right, done.

photon-ml/src/main/scala/com/linkedin/photon/ml/avro/model/ModelProcessingUtils.scala
```diff
@@ -88,17 +88,26 @@ object ModelProcessingUtils {
...
+   * @param sc The Spark context
+   * @return The feature index loaders and a Game model
+   */
   protected[ml] def loadGameModelFromHDFS(
-      featureShardIdToIndexMapLoader: Map[String, IndexMapLoader],
-      inputDir: String,
-      sparkContext: SparkContext): GAMEModel = {
+      featureShardIdToIndexMapLoader: Option[Map[String, IndexMapLoader]],
+      modelsDir: String,
+      sc: SparkContext): (Map[String, IndexMapLoader], GAMEModel) = {
```
ferd36: OK, I'll flip things over.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
```diff
@@ -42,23 +54,57 @@ protected[ml] case class Coefficients(means: Vector[Double], variancesOption: Op
...
+      sb.append(s"Number of declared non-zeros: ${means.activeSize}\n")
+      sb.append(s"Number of counted non-zeros (abs > $eps): ${means.toArray.count(Math.abs(_) > eps)}")
+      sb.append("\n")
+    }
+    val meanAndVar = meanAndVariance(means)
+    sb.append(s"Mean and stddev of the mean: ${meanAndVar.mean} ${meanAndVar.stdDev}\n")
+    sb.append(s"l2 norm of the mean: $meansL2Norm\n")
+    if (variancesL2NormOption.isDefined) {
+      sb.append(s"l2 norm of the variance ${variancesL2NormOption.get}")
+    }
```
joshvfleming: map or foreach preferred here.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
```diff
@@ -42,23 +54,57 @@ protected[ml] case class Coefficients(means: Vector[Double], variancesOption: Op
...
+   */
   override def toSummaryString: String = {
-    val stringBuilder = new StringBuilder()
-    stringBuilder.append(s"meanAndVarianceAndCount of the mean: ${meanAndVariance(means)}")
-    stringBuilder.append(s"\nl2 norm of the mean: ${norm(means, 2)}")
-    variancesOption.foreach(variances => s"\nmeanAndVarianceAndCount of variance: ${meanAndVariance(variances)}")
-    stringBuilder.toString()
+    val sb = new StringBuilder()
+    val isDense = means.getClass.getName.contains("Dense")
+    lazy val eps = MathConst.HIGH_PRECISION_TOLERANCE_THRESHOLD
```
joshvfleming: Don't need to use lazy here, no computation involved really.

photon-ml/src/test/scala/com/linkedin/photon/ml/model/CoefficientsTest.scala
```diff
@@ -66,7 +81,77 @@ class CoefficientsTest {
...
   }
 
+  /**
+   * Test that means and variances use the same dynamic type for Vector
+   */
+  @Test
+  def testPreconditions1(): Unit = {
+
+    val d: Vector[Double] = new DenseVector[Double](Array[Double](1, 2, 3))
+    val s: Vector[Double] = new SparseVector[Double](Array[Int](0, 2), Array[Double](0, 3), 3)
```
joshvfleming: Typically for cases like this, we'd use a DataProvider for these, and then the test would declare the expectedException.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
ferd36: Fixed.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
ferd36: Done.

PR #195, opened 2016-11-16
Fix type of "modelClass" in BayesianLinearModelAvro

joshvfleming:
This was previously generating a warning, since "string" cannot be null.


XianXing:
LGTM.

joshvfleming:
@XianXing is here, woohoo!

PR #198, opened 2016-11-28
Add more documentations to DefaultIndexMap and PalDBIndexMap, and fix…

XianXing:
… some related Scala style issues.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
@@ -15,10 +15,15 @@
 package com.linkedin.photon.ml.util
 
 /**
-  * Use the default system HashMap to construct an index map, highly inefficient in terms of memory usage; but easier to
-  * handle. Recommended for small feature space cases (<= 200k).
-  */
-class DefaultIndexMap(@transient val featureNameToIdMap: Map[String, Int]) extends IndexMap {
+ * Use the default system HashMap to construct an index map, highly inefficient in terms
+ * of memory usage; but easier to handle.
+ * Recommended for small feature space cases (<= 200k).
+ * For use cases where number of features > 200k, please try [[PalDBIndexMap]] instead.
+ * @param featureNameToIdMap The map from raw feature string (name) to feature index (id)
+ */
+class DefaultIndexMap(@transient val featureNameToIdMap: Map[String, Int])
+  extends IndexMap {
+
   @transient
   private var _idToNameMap: Map[Int, String] = _
   private val _size: Int = featureNameToIdMap.size
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
@@ -22,14 +22,14 @@ import java.io.{File => JFile}
 import collection.JavaConverters._
 
 /**
-  * An off heap index map implementation using PalDB.
+  * An off heap index map implementation using [[PalDB]].
   *
   * The internal implementation assumed the following things:
   * 1. One DB storage is partitioned into multiple pieces we call partitions. It should be generated and controlled by
   * [[com.linkedin.photon.ml.FeatureIndexingJob]]. The partition strategy is via the hashcode of the feature names,
   * following the rules defined in [[org.apache.spark.HashPartitioner]].
   *
-  * 2. Each time when a user is querying the index of a certain feature, the indexmap will first compute the hashcode,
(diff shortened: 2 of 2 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
```diff
@@ -36,7 +41,9 @@ class DefaultIndexMap(@transient val featureNameToIdMap: Map[String, Int]) exten
     _idToNameMap.get(idx)
   }
 
-  override def getIndex(name: String): Int = featureNameToIdMap.getOrElse(name, IndexMap.NULL_KEY)
+  override def getIndex(name: String): Int = {
+    featureNameToIdMap.getOrElse(name, IndexMap.NULL_KEY)
+  }
```
ashelkovnykov: The previous way is correct according to the Scala style guide ([here specifically](http://docs.scala-lang.org/style/declarations.html#body)).

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
```diff
@@ -157,7 +158,8 @@ class PalDBIndexMap extends IndexMap {
 
   //noinspection ScalaStyle
   override def -(key: String): Map[String, Int] = {
-    throw new RuntimeException("This map is a read-only immutable map, remove operation is unsupported.")
+    throw new RuntimeException("This map is a read-only immutable map, " +
+      "remove operation is unsupported.")
```
ashelkovnykov: Any reason for this change and the one above it? The exception messages fit on one line.

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
```diff
@@ -173,12 +175,18 @@ object PalDBIndexMap {
...
-    s"paldb-partition-${namespace}-${partitionId}.dat"
+  def partitionFilename(
+    partitionId: Int,
+    namespace: String = IndexMap.GLOBAL_NS): String = {
 
-  class PalDBIndexMapIterator(private val indexMap: PalDBIndexMap) extends Iterator[(String, Int)] {
-    private var _iter: Iterator[JMap.Entry[Any, Any]] = null
-    private var _currentItem: (String, Int) = null
+    s"paldb-partition-$namespace-$partitionId.dat"
+  }
```
ashelkovnykov: The previous way is correct according to the Scala style guide ([here specifically](http://docs.scala-lang.org/style/declarations.html#body)).

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
XianXing: It appears to my IDE that it does not fit in one line (length of this line == 106).

photon-ml/src/main/scala/com/linkedin/photon/ml/util/DefaultIndexMap.scala
XianXing: It appears to my IDE that it does not fit in one line (length of this line == 99).

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
XianXing: It appears to my IDE that it does not fit in one line (length of this line == 101).

photon-ml/src/main/scala/com/linkedin/photon/ml/util/PalDBIndexMap.scala
```diff
@@ -198,19 +202,22 @@ object PalDBIndexMap {
...
       }
 
       while (_iter.hasNext && _currentItem == null) {
         val entry = _iter.next()
-        if (entry.getKey().isInstanceOf[String]) {
-          val key = entry.getKey().asInstanceOf[String]
-          val i = indexMap.getPartitionId(key)
-          val value = entry.getValue().asInstanceOf[Int] + indexMap._offsets(i)
-          _currentItem = (key, value)
+        entry.getKey match {
```
joshvfleming: Nice! We could also match `entry.getValue` here and eliminate the other `asInstanceOf`, e.g.:

```scala
(entry.getKey, entry.getValue) match {
  case (key: String, entryValue: Int) =>
     val i = indexMap.getPartitionId(key)
     val value = entryValue + indexMap._offsets(i)
     _currentItem = (key, value)
  case _ =>
}
```

XianXing: Great! I really missing having a Scala expert around!

PR #203, opened 2016-12-15
Add basic event producer/consumer code.

ashelkovnykov:
Add basic events to Photon Driver.

Also included:
- small build file changes
- a few style changes
- removed unused Schemas
- changed used schemas to have the same namespace
- upgraded to Spark 1.6.2
- upgraded to Avro 1.7.7

Some changes were made in response to LinkedIn internal work. If you have questions, please let me know.


photon-test/build.gradle
```diff
@@ -18,7 +18,7 @@ apply from: '../build-scripts/integration-test.gradle'
 dependencies {
   compile("org.testng:testng:6.9.9")
 
-  compile("org.apache.spark:spark-core$scalaSuffix:1.4.0")
+  compile("org.apache.spark:spark-core$scalaSuffix:1.6.2")
```
ashelkovnykov: @XianXing That's still TBD, but unfortunately I doubt it will be soon.

photon-ml/src/main/scala/com/linkedin/photon/ml/event/EventEmitter.scala
```diff
@@ -0,0 +1,44 @@
...
+ */
+trait EventEmitter {
+  private var registeredListeners = List[EventListener]()
+
+  /**
+   * Register an [[EventListener]] to consume [[Event]] objects produced by this class.
+   *
+   * @param listener The listener.
+   */
+  protected def registerListener(listener: EventListener): Unit = registeredListeners = listener :: registeredListeners
```
joshvfleming: This might belong to the category of issues we don't care about right now, but this guy is not thread-safe. The following sequence of events would break it:

Thread 1 calls `registerListener`, grabs `registeredListeners`, and then gets preempted.
Thread 2 calls `registerListener`, grabs `registeredListeners` (still the old one), and gets preempted.
Thread 1 resumes and finishes adding its listener
Thread 2 resumes and finishes

At this point the result from Thread 2 overwrites the result from Thread 1, and we lose that listener.

photon-ml/src/main/scala/com/linkedin/photon/ml/event/EventEmitter.scala
```diff
@@ -0,0 +1,44 @@
...
+    registeredListeners.foreach(_.close())
+    registeredListeners = List()
+  }
+
+  /**
+   * Produce an event for all registered listeners to consume.
+   *
+   * @param event The event.
+   */
+  protected def sendEvent(event: Event): Unit = registeredListeners.foreach(_.handle(event))
```
joshvfleming: We should call `_.handle` in a `try` block, so that we still notify subsequent listeners if a previous listener fails. Also, we probably shouldn't allow it to propagate up and kill the calling function.

ashelkovnykov:
Very important comments - I'll hop on them

PR #205, opened 2016-12-15
Revert the DataFrame PR

joshvfleming:
This change causes some OOM issues on Spark 1.6. Reverting until we figure out how to fix them.

This also backports the feature duplication detection that was added as part of the DataFrame change.


XianXing:
I really liked the previous DataFrame PR.

Is there anything I can help with? I'd like to see us leveraging the DataFrame API sometime soon again.

joshvfleming:
@XianXing I'm looking into this now actually. For some reason memory utilization shot up with this implementation. I'm trying to find the cause, and will revert the revert once it's fixed. Hoping to have it back in before or shortly after the break.

PR #206, opened 2016-12-20
Exclude netty from dependency on avro-mapred

joshvfleming:
We don't use this, and it causes version conflicts in downstream projects.


ferd36:
LGTM.

ashelkovnykov:
Thanks for the help with this one

PR #210, opened 2017-01-17
Reimplement AvroDataReader with a direct-to-avro approach rather than…

joshvfleming:
… using spark-avro

The original PR #179 used spark-avro for reading Avro files, which had two problems:

1) It used a lot more memory on the grid (cause for this still not identified)

2) It caused a lot of shuffling, which made data reading much slower. The shuffling happened because the pattern for Spark's DataFrame typically hides partition counts from the user, treating this as an implementation detail that a consumer shouldn't need to worry about at the DataFrame level of abstraction. This is usually true, but in the case of Photon, we still use partition counts in a few places to tune the performance of algorithms. The DataFrame runtime has no knowledge of this and badly underestimates the number of partitions needed, which causes a huge repartition/shuffle step during reading. Here we pass the partition count explicitly to the "read" function to get around this problem. Once we revisit Photon's sensitivity to choice of partition count, we can remove this parameter.


joshvfleming:
@XianXing @ashelkovnykov @timyitong @ferd36 

This is a new version of PR #179 (the DataFrame PR) that addresses some of the memory issues we had in production. The main difference is that AvroDataReader has been completely reimplemented. The rest is pretty much the same as before.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/data/AvroDataReaderTest.scala
```diff
@@ -0,0 +1,143 @@
...
+import com.linkedin.photon.ml.constants.MathConst
+import com.linkedin.photon.ml.test.SparkTestUtils
+import com.linkedin.photon.ml.util._
+
+import org.apache.spark.mllib.linalg.{SparseVector, Vectors}
+import org.apache.spark.sql.DataFrame
+import org.apache.spark.SparkException
+import org.apache.spark.sql.functions._
+import org.testng.Assert._
+import org.testng.annotations.{DataProvider, Test}
```
ashelkovnykov: DataProvider is unused

photon-ml/src/integTest/scala/com/linkedin/photon/ml/data/AvroDataReaderTest.scala
```diff
@@ -0,0 +1,143 @@
...
+  @Test
+  def testReadMultipleFiles(): Unit = sparkTest("testReadMultipleFiles") {
+    val dr = new AvroDataReader(sc)
+    val (df, _) = dr.readMerged(Seq(inputPath, inputPath2), featureSectionMap, numPartitions)
+
+    verifyDataFrame(df, expectedRows = 44005)
+  }
+
+  @Test
+  def testReadDuplicateFeatures: Unit = sparkTest("testReadDuplicateFeatures") {
```
ashelkovnykov: Missing empty parameter parentheses

photon-ml/src/integTest/scala/com/linkedin/photon/ml/data/AvroDataReaderTest.scala
```diff
@@ -0,0 +1,143 @@
...
+    }
+  }
+
+  /**
+   * Verifies that the DataFrame has expected shape and statistics
+   *
+   * @param df the DataFrame to test
+   * @param expectedRows the number of rows the DataFrame should have
+   */
+  def verifyDataFrame(df: DataFrame, expectedRows: Int): Unit = {
```
ashelkovnykov: May be worth adding comments about how these numbers were arrived at

photon-ml/src/integTest/scala/com/linkedin/photon/ml/data/GameConvertersTest.scala
```diff
@@ -0,0 +1,181 @@
...
+
+    val sqlContext = new SQLContext(sc)
+    val schema = StructType(Seq(
+      StructField(GameConverters.FieldNames.UID, StringType),
+      StructField(USER_ID_NAME, StringType),
+      StructField(JOB_ID_NAME, LongType)))
+
+    val dataFrame = sqlContext.createDataFrame(sc.parallelize(
+      Seq(Row(uid, userIdStr, jobIdVal))
+    ), schema)
```
ashelkovnykov: This can fit on one line

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,375 @@
...
+import com.databricks.spark.avro._
+import org.apache.avro.Schema
+import org.apache.avro.Schema.Type._
+import org.apache.avro.generic.GenericRecord
+import org.apache.spark.SparkContext
+import org.apache.spark.mllib.linalg.{SparseVector, Vector, VectorUDT}
+import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.types.DataTypes._
+import org.apache.spark.sql.types.{DataType, MapType, StructField, StructType}
+import org.apache.spark.sql.{DataFrame, Row, SQLContext}
```
joshvfleming: Okay, I guess I need to find an emacs mode that will flag these for me. :-)

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,375 @@
...
+      featureColumnMap: Map[String, Set[String]],
+      numPartitions: Int): (DataFrame, Map[String, IndexMapLoader]) = {
+
+    require(paths.nonEmpty, "No paths specified. You must specify at least one input path.")
+    require(numPartitions > 0, "Partition count must be greater than zero.")
+
+    val records = AvroUtils.readAvroFiles(sc, paths, numPartitions)
+    val indexMapLoaders = generateIndexMapLoaders(records, featureColumnMap)
+
+    (readMerged(paths, indexMapLoaders, featureColumnMap, numPartitions), indexMapLoaders)
```
joshvfleming: You're right, good catch. I'm going to make a new method that accepts RDD[GenericRecord] as input, so that we can reuse the RDD instead of re-reading.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,375 @@
...
+        Row.fromSeq(cols ++ featureCols)
+      }
+    }
+
+    // Add schema fields for the feature vector columns
+    val featureFields = featureColumnMap.map { case (destCol, _) =>
+      StructField(destCol, new VectorUDT())
+    }
+
+    val sqlSchema = new StructType((schemaFields ++ featureFields).toArray)
```
joshvfleming: Yes, that's correct.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -0,0 +1,375 @@
...
+          name,
+          MapType(StringType, valueSchema.dataType, valueContainsNull = valueSchema.nullable),
+          nullable = false)
+      }
+
+    case UNION =>
+      if (avroSchema.getTypes.asScala.exists(_.getType == NULL)) {
+        // In case of a union with null, take the first non-null type for the value type
+        val remainingUnionTypes = avroSchema.getTypes.asScala.filterNot(_.getType == NULL)
+        avroTypeToSql(name, remainingUnionTypes.head).map(_.copy(nullable = true))
```
joshvfleming: Yes, this only handles union fields for the case where the type has been unioned with "null" (we see this pretty often for optional fields). Any other kind of union is ignored. We could do something a little more sophisticated like you describe, where we maintain a table of "unionable" types and read it into a predetermined gcd type.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
```diff
@@ -0,0 +1,172 @@
...
+    featureShards: Set[String],
+    idTypeSet: Set[String],
+    isResponseRequired: Boolean): RDD[(Long, GameDatum)] = {
+
+    // Add unique id
+    val recordsWithUniqueId = data.withColumn(UNIQUE_ID_COLUMN_NAME, monotonicallyIncreasingId)
+
+    recordsWithUniqueId.map { row: Row =>
+      val id = row.getAs[Long](UNIQUE_ID_COLUMN_NAME)
+      (id, getGameDatumFromRow(
```
joshvfleming: This is more of a personal style thing but I find the code more readable this way.

PR #217, opened 2017-01-26
Throw a more informative error message for empty input

joshvfleming:


photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -164,6 +164,7 @@ class AvroDataReader(
     // Infer sql schema from avro schema
     val srcFeatureCols = featureColumnMap.values.flatten.toSet
     val schemaFields = inferSchemaFields(records)
+      .getOrElse(throw new IllegalArgumentException("No records found at given input paths."))
```
fastier-li: How about adding the actual path in the error message? In my experience, that speeds up debugging: you have everything right there in one message, no need to look somewhere else to see what the input path is.

joshvfleming: I was going to do that, but at this point the path is not available, and it would require some non-trivial reorganization of the code to surface it. Also we generally log the paths before passing them to this function.

PR #218, opened 2017-01-29
Postponed - Adding random Game model generator

fastier-li:
**# _2017/02/01: This is ON HOLD for now, as the structure of the modules will change soon._** 

Brand new random Game model generator that we might use in tests for example.

### DIFF
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameTestUtils.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameTestUtils.scala
@@ -16,7 +16,7 @@ package com.linkedin.photon.ml
 
 import java.util.concurrent.atomic.AtomicLong
 
-import org.apache.spark.HashPartitioner
+import org.apache.spark.{HashPartitioner, SparkConf, SparkException}
 
 import com.linkedin.photon.ml.algorithm.{FixedEffectCoordinate, RandomEffectCoordinateInProjectedSpace}
 import com.linkedin.photon.ml.data._
@@ -266,5 +266,26 @@ trait GameTestUtils {
       featureShardId)
 
     (coordinate, model)
+  }
+
+  /**
+   * Overridden spark test provider that allows for specifying whether to use kryo serialization
+   *
+   * @param name the test job name
+   * @param body the execution closure
+   */
+  def sparkTest(name: String, useKryo: Boolean)(body: => Unit) {
+    SparkTestUtils.SPARK_LOCAL_CONFIG.synchronized {
+      sc = SparkContextConfiguration.asYarnClient(
+        new SparkConf().setMaster(SparkTestUtils.SPARK_LOCAL_CONFIG), name, useKryo)
+
+      try {
+        body
+      } finally {
+        sc.stop()
+        System.clearProperty("spark.driver.port")
+        System.clearProperty("spark.hostPort")
+      }
+    }
   }
 }
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/avro/ModelProcessingUtilsTest.scala
@@ -23,7 +23,7 @@ import org.testng.annotations.{DataProvider, Test}
 
 import com.linkedin.photon.ml.avro.Constants._
 import com.linkedin.photon.ml.avro.model.ModelProcessingUtils
-import com.linkedin.photon.ml.cli.game.training.Params
+import com.linkedin.photon.ml.estimators.GameParams
 import com.linkedin.photon.ml.constants.MathConst
 import com.linkedin.photon.ml.model._
 import com.linkedin.photon.ml.optimization.GLMOptimizationConfiguration
@@ -44,11 +44,11 @@ class ModelProcessingUtilsTest extends SparkTestUtils with TestTemplateWithTmpDi
    * @param updates A sequence of pairs (Params field name, field value) to customize the Params returned
    * @return An instance of Params
    */
-  def setupParams(updates: (String, Any)*): Params = {
+  def setupParams(updates: (String, Any)*): GameParams = {
 
     import GLMOptimizationConfiguration.{SPLITTER => S}
 
-    val params = new Params
+    val params = new GameParams
 
     // Some default optimization configurations
     val feConfig1 = GLMOptimizationConfiguration(s"10${S}1e-2${S}1.0${S}0.3${S}TRON${S}L2")
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverTest.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverTest.scala
@@ -18,26 +18,33 @@ import java.nio.file.{FileSystems, Files, Path}
 
 import scala.collection.JavaConversions._
 
-import org.apache.spark.{SparkConf, SparkException}
+import org.apache.spark.SparkException
 
 import org.testng.Assert._
 import org.testng.annotations.{DataProvider, Test}
 
-import com.linkedin.photon.ml.avro.generated.BayesianLinearModelAvro
+import com.linkedin.photon.ml.GameTestUtils
+import com.linkedin.photon.ml.TaskType
+import com.linkedin.photon.ml.TaskType.TaskType
 import com.linkedin.photon.ml.avro.AvroIOUtils
 import com.linkedin.photon.ml.avro.data.NameAndTerm
+import com.linkedin.photon.ml.avro.generated.BayesianLinearModelAvro
 import com.linkedin.photon.ml.avro.model.ModelProcessingUtils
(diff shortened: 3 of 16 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
```diff
@@ -145,10 +147,13 @@ protected[ml] object Coefficients {
...
-    require(0 < length)
-    require(indices.length == nnz.length)
-    require(indices.sorted == indices)
-    // TODO: check for duplicates?
+    require(0 < length, s"Specified length < 0 for Coefficients: $length")
+    require(indices.length == nnz.length,
+      s"""There are ${indices.length} indices but ${nnz.length} non-zero specified for Coefficients.
+         | There should be as many indices as non-zero values.
+       """.stripMargin)
+    require(indices.sorted == indices, s"Indices need to be in strictly increasing order for Coefficients")
```
joshvfleming: Why not go ahead and sort the indices here? Also, this duplicates logic where we already enforce this invariant:

https://github.com/linkedin/photon-ml/blob/master/photon-ml/src/main/scala/com/linkedin/photon/ml/util/VectorUtils.scala#L55

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
```diff
@@ -0,0 +1,102 @@
+/*
```
joshvfleming: This belongs in `photon-test` rather than `photon-ml`.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+ * number of features in each random effect, ...).
+ */
+class GameModelGenerator {
+
+  private val log : Logger = Logger.getLogger(classOf[GameModelGenerator])
+  private val rng = scala.util.Random
+
+  private def randInt(M: Int) = 1 + math.abs(rng.nextInt) % M
+
+  private def randSparseVector(M: Int) = {
```
joshvfleming: See: https://github.com/linkedin/photon-ml/blob/master/photon-test/src/main/scala/com/linkedin/photon/ml/test/CommonTestUtils.scala

This would be better moved there / unified with the existing vector generators.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+   * @param maxNSubs The maximum number of sub-models per random effect. This is again only a maximum, the actual
+   *                 number being drawn at random from the uniform distribution over [1, maxNSubs).
+   * @return a well formed Game model.
+   */
+  def generateGameModel(
+      sc: SparkContext,
+      taskType: TaskType,
+      NRE: Int = 100,
+      maxNF: Int = 100,
+      maxNSubs: Int = 100): GAMEModel = {
```
joshvfleming: `numRandomEffects`, `maxNumFeatures`, `maxNumSubModels` would be clearer.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+          } else {
+            val subs = (0 until nSubs).map { _ => model(Coefficients(numFeatures(effect))(indices:_*)(values:_*)) }
+            val names = (0 until nSubs).map { i => s"$effect-item$i" }
+            new RandomEffectModel(sc.parallelize(names zip subs), "RE" + effect, effect)
+          })
+      }.toSeq
+
+    log.info("Done generating Game model")
+
+    GAMEModel(taskType, models:_*)
```
ashelkovnykov: This section could use better spacing, IMO. Also: the `NRE` doesn't actually correspond to the # of random effects, since the first effect is treated as a fixed effect. Plus, we know that the first effect is always the one which will become a fixed effect model, so it doesn't make sense to add the conditional check into the `map` call. Possible alternative below?
```
val fixedMaxNumFeatures = randInt(maxNF)
val (fixedIndices, fixedValues) = randSparseVector(fixedMaxNumFeatures)
val fixedEffectModel = new FixedEffectModel(
  sc.broadcast(modelGenFunc(Coefficients(fixedMaxNumFeatures)(fixedIndices: _*)(fixedValues: _*))),
  "fixedEffect")

val models = numFeatures.map { case (effect, nf) =>
  val (indices, values) = randSparseVector(nf)
  val nSubs = randInt(maxNSubs)
  val subs = (0 until nSubs).map { _ => modelGenFunc(Coefficients(nf)(indices: _*)(values: _*)) }
  val names = (0 until nSubs).map { i => s"$effect-item$i" }
  val model = new RandomEffectModel(sc.parallelize(names zip subs), "RE" + effect, effect)

  (effect, model)
}

log.info("Done generating Game model")

GAMEModel(taskType, (models + ("fixedEffect" -> fixedEffectModel)).toSeq :_*)
```

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
fastier-li: According to me, the check in VectorUtils is too distant from the actual object that contains the coefficients, which makes it easy to forget/not use that function in the future. If we _localize _ that check in the Coefficients constructor, even when we refactor, or no longer go though the function in VectorUtils, we'll still check. On balance, I would vote to move the check to be in the constructor of the object.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
fastier-li: Will move it to photon-test.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
fastier-li: Will move/unify.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
fastier-li: Will change names.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
fastier-li: Very good point! Thanks!

fastier-li: Yeah, it's pseudo-uniform, and there are shortcuts. It's a trade-off between ease and being completely correct. I opted for ease on the assumption that this would be used in tests only, where aspects such as the size of the model would matter more than the distribution of the non-zeros over the feature space. I could be wrong about that assumption. The best option would be to let the user specify the distribution of the non-zeros, but I thought that was overkill here.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
```diff
@@ -0,0 +1,102 @@
...
+      s"Creating a $taskType Game model " +
+        s"with $NRE random effects, " +
+        s"at most $maxNF features per random effect " +
+        s"and at most $maxNSubs sub-models per random effect")
+
+    val numFeatures = (0 until NRE).map { i => (s"Effect$i", randInt(maxNF)) }.toMap
+
+    val model = taskType match {
+      case TaskType.LINEAR_REGRESSION => LinearRegressionModel.apply _
+      case TaskType.LOGISTIC_REGRESSION => LogisticRegressionModel.apply _
```
fastier-li: This is taken from the converter for Laser, which doesn't handle Poisson regression yet. But you are right, I should add it. No reason we can't have Poisson regression here.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
fastier-li: OK, will take into account all those excellent points.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/Coefficients.scala
fastier-li: For now, I'll remove the check from Coefficients.scala, pending a refactoring of the check in convertIndexAndValuePairArrayToSparseVector. Not having a preconditions in Coefficients directly is dangerous, as Coefficients should be a component that stands on its own, including the preconditions that verify its well-formedness. Separating the check is a recipe for having ill-formed Coefficients later upon refactoring.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/GameModelGenerator.scala
fastier-li: I adopted your better version.

PR #219, opened 2017-01-30
Fix nullable numeric and Utf8 map value bugs in AvroDataReader

joshvfleming:


photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -403,18 +403,34 @@ object AvroDataReader {
...
-      case FloatType => Some(Utils.getFloatAvro(record, field.name))
-      case LongType => Some(Utils.getLongAvro(record, field.name))
-      case MapType(_, _, _) => Some(Utils.getMapAvro(record, field.name, field.nullable).asScala)
+      case BooleanType => checkNull(record, field).orElse(Some(Utils.getBooleanAvro(record, field.name)))
+      case DoubleType => checkNull(record, field).orElse(Some(Utils.getDoubleAvro(record, field.name)))
+      case FloatType => checkNull(record, field).orElse(Some(Utils.getFloatAvro(record, field.name)))
+      case LongType => checkNull(record, field).orElse(Some(Utils.getLongAvro(record, field.name)))
+      case MapType(_, _, _) => Some(Utils.getMapAvro(record, field.name, field.nullable))
       case _ =>
         // Unsupported field type. Drop this for now.
```
fastier-li: Would there be any merit to isolating invariants? It can be helpful for readability and if checks need to be called from multiple places. Just want to bring up that possibility, might not be needed in your case.

joshvfleming: Typically yes, but in this case we have different checks depending on the type of the field.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/AvroDataReader.scala
```diff
@@ -403,18 +403,34 @@ object AvroDataReader {
    * @param schemaFields the spark sql schema to apply when reading the record
    * @return column values
    */
-  def readColumnValuesFromRecord(record: GenericRecord, schemaFields: Seq[StructField]) = schemaFields
+  protected[data] def readColumnValuesFromRecord(record: GenericRecord, schemaFields: Seq[StructField]) = schemaFields
     .flatMap { field: StructField => field.dataType match {
-      case IntegerType => Some(Utils.getIntAvro(record, field.name))
+      case IntegerType => checkNull(record, field).orElse(Some(Utils.getIntAvro(record, field.name)))
       case StringType => Some(Utils.getStringAvro(record, field.name, field.nullable))
```
namitk: Why is the null behavior for a string different? It seems if a string field is nullable, we read it as an empty string. Is that intended?

joshvfleming: This is left over from previous code, where we depend on certain behaviors. For example, the code that generates feature keys depends on null being changed to empty string to form name/term keys properly. We need to make a pass and clean this kind of thing up.

PR #220, opened 2017-02-01
Upgrade gradle to 3.3

joshvfleming:


ashelkovnykov:
Looks fine. Just wondering - any new benefits to Gradle 3.3, or are we just keeping up with our dependencies?

joshvfleming:
@ashelkovnykov The rumor is that 3.0 fixes some of the problems where a folder gets into a weird state and gradle breaks. Also, just keeping up.

PR #222, opened 2017-02-06
Initial cut of moving Game code from Driver to GameEstimator

joshvfleming:
This is an initial nudge in the direction of the new API. I've made this change mostly to enable hyperparameter optimization, and also to get the API ball rolling.

### Summary of changes

- Optimization logic moved from Driver to GameEstimator
- Tests updated to reflect code changes
- Some minor cleanup around the periphery

### Performance comparison

| Code | AUC | Time (m) |
| --- | --- | --- |
| Before #222 | 0.65435068 | 13.15227361 |
| After #222 | 0.65435068 | 12.27700118 |

No difference in coefficient stats.

### DIFF
--- a/photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverTest.scala
+++ b/photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverTest.scala
@@ -18,26 +18,33 @@ import java.nio.file.{FileSystems, Files, Path}
 
 import scala.collection.JavaConversions._
 
-import org.apache.spark.{SparkConf, SparkException}
+import org.apache.spark.SparkException
 
 import org.testng.Assert._
 import org.testng.annotations.{DataProvider, Test}
 
-import com.linkedin.photon.ml.avro.generated.BayesianLinearModelAvro
+import com.linkedin.photon.ml.GameTestUtils
+import com.linkedin.photon.ml.TaskType
+import com.linkedin.photon.ml.TaskType.TaskType
 import com.linkedin.photon.ml.avro.AvroIOUtils
 import com.linkedin.photon.ml.avro.data.NameAndTerm
+import com.linkedin.photon.ml.avro.generated.BayesianLinearModelAvro
 import com.linkedin.photon.ml.avro.model.ModelProcessingUtils
-import com.linkedin.photon.ml.data.{FixedEffectDataSet, RandomEffectDataSet}
+import com.linkedin.photon.ml.data
+import com.linkedin.photon.ml.data.{AvroDataReader, FixedEffectDataSet, GameConverters, RandomEffectDataSet}
+import com.linkedin.photon.ml.estimators.GameParams
 import com.linkedin.photon.ml.evaluation._
 import com.linkedin.photon.ml.io.ModelOutputMode
 import com.linkedin.photon.ml.optimization.OptimizerType
 import com.linkedin.photon.ml.optimization.OptimizerType.OptimizerType
-import com.linkedin.photon.ml.TaskType.TaskType
 import com.linkedin.photon.ml.test.{CommonTestUtils, SparkTestUtils, TestTemplateWithTmpDir}
-import com.linkedin.photon.ml.util.{PhotonLogger, Utils}
-import com.linkedin.photon.ml.{SparkContextConfiguration, TaskType}
+import com.linkedin.photon.ml.util.{LongHashPartitioner, PhotonLogger, Utils}
 
-class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
+
+/**
+ * Test cases for the GAME training driver
+ */
+class DriverTest extends SparkTestUtils with GameTestUtils with TestTemplateWithTmpDir {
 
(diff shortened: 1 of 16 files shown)


photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverTest.scala
```diff
@@ -123,7 +130,7 @@ class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
   def testRandomEffectsWithIntercept(): Unit = sparkTest("testRandomEffectsWithIntercept", useKryo = true) {
     val outputDir = s"$getTmpDir/randomEffects"
     // This is a baseline RMSE capture from an assumed-correct implementation on 4/14/2016
-    val errorThreshold = 2.2
+    val errorThreshold = 2.34
```
ashelkovnykov: Why has this changed?

joshvfleming: The test was previously using the wrong dataset to compute validation error. When I corrected it, the value changed. Need to revisit these tests, since they're not really giving much more signal than "yes, this still produces predictions in a somewhat reasonable range of reals".

fastier-li:
Is the difference in runtimes before and after 222 significant?

joshvfleming:
@fastier-li Nope.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -40,7 +42,7 @@ class CoordinateDescent(
     coordinates: Seq[(String, Coordinate[_ <: DataSet[_], _ <: Coordinate[_, _]])],
     trainingLossFunctionEvaluator: Evaluator,
     validatingDataAndEvaluatorsOption: Option[(RDD[(Long, GameDatum)], Seq[Evaluator])],
-    logger: PhotonLogger) {
+    logger: Logger) {
```
joshvfleming: PhotonLogger is an implementation of Logger. These classes shouldn't care about the backing implementation.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
```diff
@@ -0,0 +1,305 @@
+/*
+ * Copyright 2016 LinkedIn Corp. All rights reserved.
```
joshvfleming: Actually, were don't really need to update these. Copyright lasts for a really long time.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -222,33 +225,36 @@ class CoordinateDescent(
...
-            bestEval = Some(currentEval)
+        case Some(evaluators: EvaluationResults) if evaluators.nonEmpty =>
+          // The first evaluator is used for model selection
+          val (evaluator, eval) = evaluators.head
+
+          if (bestEval.map(e => evaluator.betterThan(eval, e.head._2)).getOrElse(true)) {
+            bestEval = Some(evaluators)
             bestModel = Some(updatedGAMEModel)
-            logger.debug(s"Found better GAME model, with evaluation: $currentEval")
+            logger.debug(s"Found better GAME model, with evaluation: $eval")
```
joshvfleming: There's no change in logic really. Only difference is that we keep around all the evaluations and return them from the function.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/CoordinateDescent.scala
```diff
@@ -222,33 +225,36 @@ class CoordinateDescent(
...
 
-          (firstEvaluator, currentEvaluation)
+          currentEvaluation
         } // validatingDataAndEvaluatorsOption.map
 
         logger.info(s"Updating coordinate $coordinateId finished, time elapsed: " +
             s"${coordinateTimer.stop().durationSeconds} (s)\n")
 
-        currentEvaluation // the first evaluator of the last coordinate evaluates the full model
+        currentEvaluation
```
joshvfleming: I think I removed this because it's not quite right now -- this will contain all the evaluators, not just the first one.

photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/DriverTest.scala
```diff
@@ -560,7 +391,7 @@ class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
     * @param args the command-line arguments
     */
   def runDriver(args: Array[String]): Driver = {
-    val params = Params.parseFromCommandLine(args)
+    val params = GameParams.parseFromCommandLine(args)
```
joshvfleming: Holding back on changes to Params, because this is going through a big overhaul anyway with Alex's work.

PR #233, opened 2017-02-17
Fix bug where GAME scoring Driver test sometimes failed with NaN

joshvfleming:
For some random seeds and different sequences of random value generation, this test would fail because all the labels for either queryId or documentId happened to be the same, which produces an AUC of NaN. The fix is to add static data that ensures that each item id has both positive and negative labels.


photon-ml/src/integTest/scala/com/linkedin/photon/ml/cli/game/scoring/DriverTest.scala
```diff
@@ -185,15 +185,39 @@ class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
...
+          idTypeToValueMap = Map(
+            "queryId" -> queryId.toString,
+            "documentId" -> documentId.toString)
+        )
+      }
+    }
+
+    // Add ids and parallelize
+    val staticGameDataSet = sc.parallelize(
+      (numSamples until numSamples + staticGameData.length)
```
fastier-li: Is this right?

joshvfleming: Yes, I believe so. Did I miss a bug?

fastier-li: Nope - looks good - I was thrown off for a second thinking this was a loop, but it is a Range.

PR #234, opened 2017-02-17
Fix map zipping bug in ModelProcessingUtilsTest

namitk:
Calling zip directly on maps doesn't zip on keys but just zips the underlying (key, value) iterables. So our current test is incorrect


fastier-li:
Just lost my previous comment, but I'd like to see another fix for this, by replacing GAMEModel.toMap with:

` def toMap: SortedMap[String, DatumScoringModel] = SortedMap[String, DatumScoringModel](gameModels.toSeq: _*)`

Which is a more powerful fix, because more general. It will avoid further issues down the road, and any possible performance impact should be minimal or irrelevant to our critical path.

joshvfleming:
@fastier-li If we make that change, I think it would be best to also rename the method to `toSortedMap`. Otherwise, it won't be immediately obvious in the consuming code that `zip` is going to do the right thing. It could also be dangerous if folks start following the pattern of zipping maps without realizing that the original case was specifically a sorted map.

fastier-li:
Makes sense. Then I would vote to rename. I can do it if it's a lot of work.

joshvfleming:
LGTM

PR #235, opened 2017-02-18
Changes for scaling GAME scoring

namitk:
Two main changes
1. Adds a new data structure ScoredGameDatum
2. The scoring using random effect model is now done using a replicated partitioned hash join

Still needs holistic testing on existing workflows to ensure there aren't side-effects


photon-ml/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
```diff
@@ -140,29 +140,34 @@ object RandomEffectModel {
...
-            s"random effect type $randomEffectType")
-
-        if (modelsIterable.isEmpty) {
-          uniqueIdAndFeaturesIterable.map { case (uniqueId, _) => (uniqueId, 0.0) }
-        } else {
-          val model = modelsIterable.head
-          uniqueIdAndFeaturesIterable.map { case (uniqueId, features) =>
-            (uniqueId, model.computeScore(features))
+      .partitionBy(hashPartitioner)
+      .zipPartitions(modelsRDD.partitionBy(hashPartitioner))(
```
XianXing: Since we already set the partitioner explicitly for both dataPoints and modelsRDD, why not directly call cogroup on them, which has arguably better readability? Note that if two RDDs share the same partitioner, the cogroup operation won't trigger extra shuffling, and after calling cogroup, the operations below should be able to be reused.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
```diff
@@ -140,29 +140,34 @@ object RandomEffectModel {
...
+        (dataIt, modelIt) => {
+          val lookupTable = modelIt.toMap
+          dataIt.map {
```
XianXing: I like it! Only materializing the smaller object, which is the model, is very smart.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,118 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.data
+
+import scala.collection.Map
+
+import com.linkedin.photon.ml.util.Utils.isAlmostZero
+
+/**
```
fastier-li: Formatting should follow DataBricks guideline for comments (align on first star of `/**`).

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,118 @@
...
+  * @param weight importance weight
+  * @param score computed score for this instance
+  * @param idTypeToValueMap The id type to value map that holds different types of ids associated with this data
+  *                         point. A few examples of the ids types are: (i) ids used to build the random effect model
+  *                         such as userId and itemId; (ii) ids used to compute certain metrics like precision@k such
+  *                         as documentId or queryId; (iii) ids that are used to uniquely identify each training record.
+  */
+case class ScoredGameDatum(response: Double, offset: Double, weight: Double,
+                           var score: Double, idTypeToValueMap: Map[String, String]) extends Serializable {
+  /**
```
fastier-li: Any invariants we can assert in the constructor? This should help prevent bugs, now while thinking about the invariants, and later when creating instances.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,118 @@
...
+    * Get a copy of the current instance with a score of [[ScoredGameDatum.DEFAULT_SCORE]]
+    *
+    * @return copy of the current scored datum instance with zero score
+    */
+  def getZeroScoreDatum(): ScoredGameDatum = {
```
fastier-li: Can we just use the constructor, with a score default value of `ScoredGameDatum.DEFAULT_SCORE`?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -77,16 +82,20 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, Double)]) extends RDDLi
     val subtractedScores =
       this.scores
         .fullOuterJoin(that.scores)
-        .mapValues { case (thisScore, thatScore) => thisScore.getOrElse(0.0) - thatScore.getOrElse(0.0) }
+        .mapValues { case (thisScore, thatScore) =>
+          require(!thisScore.isEmpty || !thatScore.isEmpty, s"Should not have reached this state. A full outer join " +
```
fastier-li: Can this case actually happen? If so, it would be worth explaining the circumstances in a comment. If not, we should retire this `require`.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -63,7 +63,12 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, Double)]) extends RDDLi
     val addedScores =
       this.scores
         .fullOuterJoin(that.scores)
-        .mapValues { case (thisScore, thatScore) => thisScore.getOrElse(0.0) + thatScore.getOrElse(0.0) }
+        .mapValues { case (thisScore, thatScore) =>
+          require(!thisScore.isEmpty || !thatScore.isEmpty, s"Should not have reached this state. A full outer join " +
```
fastier-li: Same question as below.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -63,7 +63,12 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, Double)]) extends RDDLi
     val addedScores =
```
fastier-li: Can we factor out the code for the addition and subtraction, passing in a lambda for the operation to perform? Only the +/- changes between these 2 functions.

joshvfleming: +1 for `binaryOp` function that accepts `that` and `+` or `-`.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -63,7 +63,12 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, Double)]) extends RDDLi
...
       this.scores
         .fullOuterJoin(that.scores)
-        .mapValues { case (thisScore, thatScore) => thisScore.getOrElse(0.0) + thatScore.getOrElse(0.0) }
+        .mapValues { case (thisScore, thatScore) =>
+          require(!thisScore.isEmpty || !thatScore.isEmpty, s"Should not have reached this state. A full outer join " +
+            s"should yield a result that has at least one of $thisScore or $thatScore non-empty")
+          thisScore.getOrElse(thatScore.get.getZeroScoreDatum()) +
+            thatScore.getOrElse(thisScore.get.getZeroScoreDatum())
```
joshvfleming: Why is `thatScore` the zero for `thisScore` and vice versa? Not that it would practically matter but his had me scratching my head a little.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
joshvfleming: I actually prefer "zero" here, since we're using it as additive identity in a couple places. Maybe we should change the constant to `ZERO_SCORE`?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,118 @@
...
+  * @param offset offset
+  * @param weight importance weight
+  * @param score computed score for this instance
+  * @param idTypeToValueMap The id type to value map that holds different types of ids associated with this data
+  *                         point. A few examples of the ids types are: (i) ids used to build the random effect model
+  *                         such as userId and itemId; (ii) ids used to compute certain metrics like precision@k such
+  *                         as documentId or queryId; (iii) ids that are used to uniquely identify each training record.
+  */
+case class ScoredGameDatum(response: Double, offset: Double, weight: Double,
+                           var score: Double, idTypeToValueMap: Map[String, String]) extends Serializable {
```
joshvfleming: Style convention requires giving each argument its own line, with indent of 4 spaces, e.g.:

```scala
case class ScoredGameDatum(
    response: Double,
    offset: Double,
    weight: Double,
    var score: Double,
    idTypeToValueMap: Map[String, String])
  extends Serializable {
```

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,118 @@
...
+ * under the License.
+ */
+package com.linkedin.photon.ml.data
+
+import scala.collection.Map
+
+import com.linkedin.photon.ml.util.Utils.isAlmostZero
+
+/**
+  * Representation of a single scored GAME data point
```
joshvfleming: Need a few words here explaining the purpose.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
joshvfleming: Yeah. Reading this, the crux of the original issue is not obvious to me. If it was that the data/models partitions were not aligned, which caused lots of shuffling during cogroup, then the repartition alone should fix the problem. I believe the cogroup join would internally become a hash join similar to below in this case.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
```diff
@@ -26,16 +26,20 @@ import com.linkedin.photon.ml.cli.game.GAMEDriver
...
 
+
+
 /**
  * Driver for GAME full model scoring
  */
 class Driver(val params: Params, val sparkContext: SparkContext, val logger: Logger)
   extends GAMEDriver(params, sparkContext, logger) {
 
+  val gameDataPartitioner = new LongHashPartitioner(parallelism)
```
joshvfleming: Doesn't look like this is used anymore.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: Fine with me, as long as we are consistent. But if it is zero, yes, it's additive identity, so it can be something else. The problem is that the "zero" score datum is actually a vector, and only one element of that vector is going to be added (the score), and is going to be set to zero. There's some amount of built-in ambiguity.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: Can we just add a ```var score: Double``` to the existing ```GameDatum```?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: See my suggestion to just add a `var score` to `GameDatum`.

fastier-li:
Is there a problem statement or a ticket for what this PR intends to accomplish?

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/FixedEffectCoordinate.scala
```diff
@@ -157,8 +157,8 @@ object FixedEffectCoordinate {
    */
   private def score(fixedEffectDataSet: FixedEffectDataSet, fixedEffectModel: FixedEffectModel): KeyValueScore = {
```
fastier-li: This private function in `object FixedEffectCoordinate` is exclusively called by `FixedEffectCoordinate.score` in the class `FixedEffectCoordinate`. Is there a reason to not inline it? Same for `updateModel`, in both fixed and random coordinates. With that and `apply` in the `ScoredGameDatum` object, the code gets much simpler:

```scala
 override protected[algorithm] def score(model: DatumScoringModel): KeyValueScore = {

    model match {
      case fixedEffectModel: FixedEffectModel =>
        val modelBroadcast = fixedEffectModel.modelBroadcast
        KeyValueScore(
          dataSet
            .labeledPoints
            .mapValues {
              dataPoint => ScoredGameDatum(dataPoint, modelBroadcast.value.computeScore(dataPoint.features))
            })

      case _ =>
        throw new UnsupportedOperationException(s"Updating scores with model of type ${model.getClass} " +
          s"in ${this.getClass} is not supported!")
    }
  }
```

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -190,7 +192,12 @@ object RandomEffectCoordinate {
     //TODO: Need a better design that properly unpersists the broadcasted variables and persists the computed RDD
     val modelsForPassiveDataBroadcast = passiveData.sparkContext.broadcast(modelsForPassiveData)
     val passiveScores = passiveData.mapValues { case (randomEffectId, labeledPoint) =>
-      modelsForPassiveDataBroadcast.value(randomEffectId).computeScore(labeledPoint.features)
+      ScoredGameDatum(
```
fastier-li: We could have a `ScoreGameDatum.apply` in the object that takes a labeled point and supplies the default `Map()` at the end.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
fastier-li: Confirmed.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
```diff
@@ -114,32 +117,37 @@ class Driver(val params: Params, val sparkContext: SparkContext, val logger: Log
...
    *
-   * @param gameDataSet The GAME data set
    * @param scores The computed scores
    */
-  protected def saveScoresToHDFS(
-      gameDataSet: RDD[(Long, GameDatum)],
-      scores: KeyValueScore): Unit = {
+  protected def saveScoresToHDFS(scores: KeyValueScore): Unit = {
 
     // Take the offset information into account when writing the scores to HDFS
```
fastier-li: I would put comments like that in the function comment block, as it is an important piece of information for users: it specifies the contract for the scores on HDFS.

fastier-li: Also, further down, we call `ScoreProcessingUtils.saveScoredItemsToHDFS`, why not inline?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  /**
+    * The plus operation for scored data. Two data can only be added if their [[response]], [[offset]], [[weight]] and
+    * [[idTypeToValueMap]] are identical. See [[similarTo()]]
+    *
+    * @param that The other scored datum instance
+    * @return A new scored datum instance with score being the sum of the two input scores
+    */
+  def +(that: ScoredGameDatum): ScoredGameDatum = {
+    require(similarTo(that), s"Can only add two ScoredGameDatum objects if their response, offset, weight" +
```
fastier-li: Also, adding a `score` field to `GameDatum` directly could further simplify, unless there is a solid reason to have a new `ScoredGameDatum` class that I missed?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
joshvfleming: @fastier-li I think idea is to have a scored object that specifically does not contain the feature vectors. Once scoring is done they're not used anymore and unnecessarily take up memory during subsequent operations.

fastier-li: OK.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
joshvfleming: +1 Good suggestion.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: See Josh's comment above. Let's keep `ScoredGameDatum` after all.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
fastier-li: If a reviewer had to scratch their head, it deserves a comment :-)

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
namitk: @fastier-li `GameDatum` is a beefy data structure since it contains the feature vectors for all shards. This is impractical to pass around for shuffles. While we can be careful with tricks like co-partitioning, this would be a contract contributors will have to keep track of which leads to maintenance risks. Even if we enforced that, co-partitioning does not ensure co-location and passing around GameData over network traffic is not optimal either.

I am also working on a document that explains what we tried and how we got from unable to score certain datasets to not meeting required SLAs to the current solution. I will also add some of the relevant information as documentation in the code.

photon-ml/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
namitk: That's a good point about readability. I'll change it to co-group

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: @namitk - thanks! Some of this reasoning to have ScoredGameDatum would be welcome in comments for class ScoredGameDatum :-)

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
fastier-li: That was my first impression - but then maybe there is a special case where that could happen? If it does, it deserves a comment.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
fastier-li: I am not a big fan of that ```isSimilar``` that I hope we can eliminate by re-working the ```fullOuterJoin```.

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
fastier-li: A ```KeyValueScore.apply``` is also needed - I needed one when I wrote the snippet above for `FixedEffectCoordinate.score`.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
fastier-li: @ashelkovnykov - why?

photon-ml/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -144,7 +144,9 @@ object RandomEffectCoordinate {
       .join(randomEffectModel.modelsRDD)
       .flatMap { case (randomEffectId, (localDataSet, model)) =>
         localDataSet.dataPoints.map { case (uniqueId, labeledPoint) =>
-          (uniqueId, model.computeScore(labeledPoint.features))
+          (uniqueId,
+            ScoredGameDatum(labeledPoint.label, labeledPoint.offset, labeledPoint.weight,
+              model.computeScore(labeledPoint.features), Map()))
```
fastier-li: This can also be alleviated by having a `ScoreGameDatum.apply(LabeledPoint, score...)`. It will make the line shorter.

photon-ml/src/test/scala/com/linkedin/photon/ml/data/ScoredGameDatumTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    Map("src" -> "1234", "dst" -> "5678"))
+  private val similarCustomDatumWithNanResponse = ScoredGameDatum(
+    Double.NaN,
+    0.5 + epsilon,
+    1.2 - epsilon,
+    -2.0,
+    Map("src" -> "1234", "dst" -> "5678"))
+
+  @Test
+  def zeroScoreDatumTest(): Unit = {
```
fastier-li: Start? Did you mean "end"?

photon-ml/src/main/scala/com/linkedin/photon/ml/data/LocalDataSet.scala
```diff
@@ -15,14 +15,14 @@
 package com.linkedin.photon.ml.data
 
 import java.util.Random
```
fastier-li: @namitk - Namit, make sure you get the "Spark Style" setting from me or Alex to install in your IDEA. It has e.g. the formatting we want for the imports, and of course, a lot more. We try to standardize on that.

photon-ml/src/test/scala/com/linkedin/photon/ml/data/ScoredGameDatumTest.scala
ashelkovnykov: Open up any other unit/integration test class and every `@Test` annotated method begins with "test".

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
ashelkovnykov: Ideally, we would get rid of the `Driver` classes, or keep them around as examples of how Photon-ML users may want to string together several API calls. In such a case, the code responsible for outputting scored records in AVRO format should be its own function made available through the client to whomever wants to use it and in whatever manner. The Driver shouldn't know or care about how the scores are saved, just that they are (on which note, the method should maybe be renamed to `saveScoredItermsToHdfsAsAvro`). Logical chunks of code responsible for only one action should be encapsulated and made available in a logical location to encourage reusing them.

ashelkovnykov:
@fastier-li As I understand it, it is taken care of by construction, since each initial `ScoredGameDatum` is generated from a `GameDatum`, meaning that any `(Long, ScoredGameDatum)` pairs with identical `Long = uniqueId` should be mergeable because they were generated from the same `GameDatum` and thus must be the same, aside from the score.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
joshvfleming: With a full outer join, one of the two values must be there. I think this `require` is a defensive move since we're calling `get` on the options in L69-70. As @ashelkovnykov suggested, a match/case block would be a more natural and idiomatic way of dealing with this.

photon-ml/src/test/scala/com/linkedin/photon/ml/data/ScoredGameDatumTest.scala
fastier-li: OK - either way, we just have to be consistent.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
fastier-li: I was not asking to inline `saveScoredItemsToHDFS` in the `Driver`. The code of the `Driver` itself is fine.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
fastier-li: That looks much better, but I would suggest we merge that with the code of the `fullOuterJoin` on the previous line, which is (so parallel!):

```scala
this.cogroup(other, partitioner).flatMapValues {
      case (vs, Seq()) => vs.iterator.map(v => (Some(v), None))
      case (Seq(), ws) => ws.iterator.map(w => (None, Some(w)))
      case (vs, ws) => for (v <- vs.iterator; w <- ws.iterator) yield (Some(v), Some(w))
 }
```

So, if we just use a `cogroup` and 3 cases (the 4th cannot happen), we are done, and it should be shorter and faster.

photon-ml/src/test/scala/com/linkedin/photon/ml/data/ScoredGameDatumTest.scala
```diff
@@ -0,0 +1,126 @@
...
+    val copy = customDatum.getZeroScoreDatum()
+    Assert.assertTrue(copy.similarTo(customDatum))
+    Assert.assertTrue(Utils.isAlmostZero(copy.score))
+  }
+
+  @Test
+  def similarToTest(): Unit = {
```
fastier-li: See discussion above about eliminating `similarTo`.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: Also, I'm realizing that `+` is not necessarily appropriate here: all that happens is that we add the score. So the method should be named `addScore` which conveys at once exactly what it does.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
ashelkovnykov: What did you want to inline?

ashelkovnykov:
After talking offline with @fastier-li offline:

Right now we're using the `isSimilarTo` function as a proxy for what we really want to know, which is did two `ScoredGameDatum` come from the same `GameDatum`, meaning that they can be merged. I propose that instead, each `ScoredGameDatum` contain either a unique ID or a reference to the `GameDatum` which generated it. The `isSimilarTo` method can be replaced by two methods: one to check whether joining two `ScoredGameDatum` is possible by comparing the aforementioned IDs/references, and one to do the joining. I'm okay with either of Frank's ideas (`addScore` or `combineScores` with a lambda defining the combination) so long as they take a `ScoredGameDatum` as input, and not a `Double`.

photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
fastier-li: `ScoreProcessingUtils.saveScoredItemsToHDFS` - this class/object is a bit light to be a class/object on its own. It's really just 2 functions used just here.

joshvfleming:
@ashelkovnykov Sounds reasonable to me. I'd go with unique id since we already assign a unique id to each `GameDatum`, and since including a reference back to the original `GameDatum` would defeat the purpose of `ScoredGameData`.

ashelkovnykov:
@joshvfleming Right, that whole "distributed" thing...

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
fastier-li: I did a small offline benchmark, and a direct `cogroup -> flatMapValues` was about 6 times faster than `fullOuterJoin -> mapValues`.

namitk:
The latest is as follows

1. Had an offline discussion with Alex about id changes. The problem there is replication of data. All uses of GameDatum and ScoredGameDatum are in PairRDD of (Long, Datum). Adding unique id in any of these data structures will mean copying the long into the value as well. Therefore for now, we have left the code as is devoid of all the require statements (which means a user can add any two arbitrary scored data) and we will do a test to see how much the require statements hurt performance. If it isn't significant, we both agree that it is a good practice to program defensively

2. Had an offline discussion with Frank about the co-group. While I have made the change, there were certain assumptions about the partitioning as well as missing assert statements. In addition, single box performance is not a good proxy for distributed performance and there is no guarantee that these will provide real benefit. If with the current changes, we are able to meet SLA on the datasets we attempted to score previously, we can keep this change around. Otherwise, it will have to be reverted

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -63,14 +63,21 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, ScoredGameDatum)]) exte
...
-          case (None, Some(thatScore)) => thatScore
+    val scores = this.scores
+      .cogroup(that.scores)
+      .flatMapValues { case (s1It, s2It) =>
+        assert(
+          s1It.size <= 1 && s2It.size <= 1,
+          s"More than one scored data found in co-grouped RDD for an id. Found ${s1It.size} records on the left, " +
+            s"${s2It.size} on the right")
+        if (s1It.isEmpty) {
+          Seq(op(s2It.head.getZeroScoreDatum, s2It.head))
```
joshvfleming: Do we need to add the zero value here? Since we know we have only s2It, I think we can just return its value.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -63,14 +63,21 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, ScoredGameDatum)]) exte
...
+        } else if (s2It.isEmpty) {
+          Seq(op(s1It.head, s1It.head.getZeroScoreDatum))
+        } else {
+          Seq(op(s1It.head, s2It.head))
         }
+      }
```
joshvfleming: Does this approach perform significantly better? I find this result a little surprising, and it might need to inform some changes down the road on how we organize chained operations. Can we publish the benchmarking scripts and analysis on a wiki or something? @fastier-li

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
namitk: Think of the subtract operation. If s1It is absent, the result is negative of s2It which is 0-s2It

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -63,14 +63,21 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, ScoredGameDatum)]) exte
...
+          s1It.size <= 1 && s2It.size <= 1,
+          s"More than one scored data found in co-grouped RDD for an id. Found ${s1It.size} records on the left, " +
+            s"${s2It.size} on the right")
+        if (s1It.isEmpty) {
+          Seq(op(s2It.head.getZeroScoreDatum, s2It.head))
+        } else if (s2It.isEmpty) {
+          Seq(op(s1It.head, s1It.head.getZeroScoreDatum))
+        } else {
+          Seq(op(s1It.head, s2It.head))
         }
```
joshvfleming: Can we use a match/case block for this? As is, the readability of this version is much worse than the original.

photon-ml/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
joshvfleming: Right, makes sense. Forgot about subtract.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -202,9 +214,15 @@ final class Driver(val params: GameParams, val sparkContext: SparkContext, val l
...
    *
+<<<<<<< HEAD:photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
    * @param featureShardIdToFeatureMapLoader The shard ids
    * @param models All the models that were producing during training
    * @param bestModel The best model
+=======
+   * @param featureShardIdToFeatureMapLoader the shard ids
+   * @param models all the models that were produced during training
+   * @param bestModel the best model
+>>>>>>> Changes for scaling GAME scoring: add a new data structure ScoredGameDatum, implement replicated partitioned hash join:photon-ml/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```
ashelkovnykov: Looks like this file had some issues during rebase

ashelkovnykov:
@namitk 
@joshvfleming 
@fastier-li 
@XianXing 

I did some performance tests on GAME training, in addition to those Namit performed for GAME scoring. Replacing the previous `Double` scores with `ScoredGameDatum` results in an object being created for each data point when computing partial scores. This adds to the amount of time spent in GC during training.

Testing on the Yahoo music dataset (3 outer iterations, 4 effects total, 700 million data points, 30k features) the impact is around 10 minutes on a 50 minute job (~50 -> ~60). Testing on a LI internal dataset (1 outer iteration, 3 effects total, 575 million data points, 300 features) the impact is around 3 minutes on a 19 minute job (~19 -> ~22).

In addition, jobs that are tuned to be right at the edge of resource usage need to have their memory tuned up a little.

One possible solution is an alternative set of classes/methods (`KeyValueScore`, `Coordinate.updateModel`, `DataSet.addScoresToOffsets`, etc.) for training that use `Double` (as they previously did). This solution isn't the prettiest (for those that remember, think of what `Optimizer` looked like when it had methods for `RDD` and `Iterable`). Another solution is to document this performance degradation, push it, and hold off on making the decision until we're aware of major negative impact to our users.

fastier-li:
I would vote to hold off till we can compare with Alex's alternative that doesn't seem to degrade performance. It would be ideal if we could have that in a PR/diff/patch format.

ashelkovnykov:
Take a look at the 'oldTrain' branch on [Namit's fork](https://github.com/namitk/photon-ml/). It contains some changes I made to verify that it was the `ScoredGameDatum` objects slowing training down.

Compare it against the scaling branch for this PR. It's not super easy to do on GitHub, so you may want to pull it locally.

joshvfleming:
I wonder: is it the new object creation that causes the problem? Or is it that we're now retaining extra (duplicate) information in the scoring path while training? Either way, it makes sense to me to have a "light" version of scoring when we don't need that information (e.g. during training).

fastier-li:
I don't have time to go over both proposals today but I want to do it. Is it ok to wait till tomorrow before making a decision here?

fastier-li:
Apparently need to rebase - I tried to apply the patch to an up-to-date branch this morning, and it failed. Can you update?

photon-lib/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,146 @@
...
+  }
+
+  /**
+   * Get a default scored datum instance
+   *
+   * @param score input score (default: [[ZERO_SCORE]])
+   * @return Returns a default scored datum with a response=1.0, offset=0.0, weight=1.0, score=input score or default
+   *         and an empty map
+   */
+  def getDefaultScoredDatumWithScore(score: Double = ZERO_SCORE): ScoredGameDatum = {
```
fastier-li: How about using default values on `apply` to avoid a lengthy function name? Or a second `apply` which allows only setting the score.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,146 @@
...
+ *                         point. A few examples of the ids types are: (i) ids used to build the random effect model
+ *                         such as userId and itemId; (ii) ids used to compute certain metrics like precision@k such
+ *                         as documentId or queryId; (iii) ids that are used to uniquely identify each training record.
+ */
+case class ScoredGameDatum(
+  response: Double,
+  offset: Double,
+  weight: Double,
+  var score: Double,
+  idTypeToValueMap: Map[String, String]) extends Serializable {
```
fastier-li: Is there a way to use integers rather than `String` for this `Map`? It would decrease the weight of this class further, which could be advantageous for performance.

fastier-li:
Namit - 
- can you add a description of the problem, or point to a document that explains the problem and design? 
- can you address Josh's question about where we actually degrade the performance?
- can you take a look at the further comments I added this morning?

photon-lib/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
```diff
@@ -0,0 +1,146 @@
...
+ * Representation of a single scored GAME data point. This data structure handles the transition from a training
+ * example with features to an example that has been scored. Adding a score field to GameDatum is infeasible because
+ * after scoring, the feature shards are no longer needed and keeping them around increases the cost to shuffle the
+ * data structure around
+ *
+ * @param response The response or label
+ * @param offset offset
+ * @param weight importance weight
+ * @param score computed score for this instance
+ * @param idTypeToValueMap The id type to value map that holds different types of ids associated with this data
```
fastier-li: Also, are default values useful on the `apply` method too? If anywhere, they could be more useful on the `apply` (as in, hiding the constructor in favor of the factory that `apply` is?)?

photon-lib/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
fastier-li: Dropping this for now. In any case, it seems the gain would not be significant.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -76,49 +81,46 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, ScoredGameDatum)]) exte
...
-    val scores = this.scores
-      .cogroup(that.scores)
-      .flatMapValues {
-        case (Seq(sd1), Seq(sd2)) => Seq(op(sd1, sd2))
-        case (Seq(), Seq(sd2)) => Seq(op(sd2.getZeroScoreDatum, sd2))
-        case (Seq(sd1), Seq()) => Seq(op(sd1, sd1.getZeroScoreDatum))
-      }
-    new KeyValueScore(scores)
-  }
+  def binaryOp(op: (ScoredGameDatum, ScoredGameDatum) => ScoredGameDatum, that: KeyValueScore): KeyValueScore =
```
fastier-li: More descriptive name? This is in effect implementing a join operation. Should the name reflect that?

photon-lib/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
```diff
@@ -76,49 +81,46 @@ protected[ml] class KeyValueScore(val scores: RDD[(Long, ScoredGameDatum)]) exte
...
-    * The plus operation for the key value scores
+    * The plus operation for the key-value scores.
     *
     * @param that The other key value score instance
     * @return A new key value score instance encapsulating the accumulated values
     */
-  def +(that: KeyValueScore): KeyValueScore = {
-    this.binaryOp((a, b) => a + b, that)
-  }
+  def +(that: KeyValueScore): KeyValueScore = binaryOp((a, b) => a.copy(score = a.score + b.score), that)
```
fastier-li: Is there a merit to `.copy` rather than new object construction? I'm thinking that eventually, a constructor would be the right location for invariants checking, and copy would sidestep that.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/ScoredGameDatum.scala
ashelkovnykov: Since `ScoredGameDatum` is a case class, it has an `apply` method auto-generated. Alternatively, we could have it not be a case class and add another `apply` method with default values, but this would have identical behaviour unless we made the constructor private/protected, in which case the apply method is just forcing the code to use an identical apply method instead of the constructor

photon-api/src/integTest/scala/com/linkedin/photon/ml/evaluation/ShardedAreaUnderROCCurveEvaluatorTest.scala
```diff
@@ -79,7 +79,7 @@ class ShardedAreaUnderROCCurveEvaluatorTest extends SparkTestUtils {
 
     val evaluator = getEvaluator(labels, ids)
     val actualResult = evaluator.evaluate(sc.parallelize(scores.map { case (id, score) =>
-      (id, ScoredGameDatum.getDefaultScoredDatumWithScore(score))
+      (id, ScoredGameDatum(score = score))
```
ashelkovnykov: Score isn't the first item in the constructor, so to create a new instance using a defined score but otherwise default values, we must declare which parameter we're providing

photon-lib/src/main/scala/com/linkedin/photon/ml/data/KeyValueScore.scala
ashelkovnykov: In Scala, all object creation goes through the base constructor, so invariant checks cannot be avoided

photon-api/src/integTest/scala/com/linkedin/photon/ml/algorithm/FixedEffectCoordinateTest.scala
```diff
@@ -37,7 +37,7 @@ class FixedEffectCoordinateTest extends SparkTestUtils with GameTestUtils {
 
     // Score before model update
     val score = coordinate.score(model)
-    assertTrue(score.scores.map(_._2).collect.forall(_.score == 0.0))
+    assertTrue(score.scores.map(_._2).collect.forall(scoredDatum => MathUtils.isAlmostZero(scoredDatum.score)))
```
ashelkovnykov: I'm gonna leave this one - it's basically identical to this (which I think is what you're asking for)
```
score.scores.map(_._2).collect.foreach(scoredDatum => assertEquals(scoredDatum, 0.0, MathConst.HIGH_PRECISION_THRESHOLD))
```

PR #236, opened 2017-02-22
Adding documentation about random projections

fastier-li:
After discussion with Bee-Chung.

### DIFF
--- a/photon-ml/src/main/scala/com/linkedin/photon/ml/projector/README.md
+++ b/photon-ml/src/main/scala/com/linkedin/photon/ml/projector/README.md
@@ -1,0 +1,20 @@
+# Random projections and scaling
+
+Random projections can be helpful to reduce the number of degrees of freedom in the model when the
+training data is insufficient and thus avoid over-fitting. For pure scaling, Game's strategy is to optimize
+"by parts", by optimizing each random effect in parallel, so there is no need to use projections to tackle
+scaling when there is enough training data.
+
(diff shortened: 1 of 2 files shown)


photon-ml/src/main/scala/com/linkedin/photon/ml/projector/README.md
```diff
@@ -0,0 +1,7 @@
+# Random projections and scaling
+
+Random projections can be helpful to reduce the number of degrees of freedom in the model when the
+training data is insufficient and thus avoid over-fitting. For pure scaling, Game's strategy is to optimize
+"by parts", by optimizing each random effect in parallel, so there is no need to use projections to tackle
+scaling when there is enough training data.
```
XianXing: The index map based projector is effective enough to scale up the model training by exploiting the feature sparsity of the random effect model, so random projection and other projection don't appear to be critical given the following size of data: ~ 100 millions of random effects, and ~1000 *unique* features per random effect.

As a result, our recommendation is to use the index map based projection as default for model training.

Also please feel free to add the following two references to the doc that I used when implementing the random projection method:
http://scikit-learn.org/stable/modules/random_projection.html
https://en.wikipedia.org/wiki/Random_projection

fastier-li: Thanks! I have updated the documentation. Let me know what you think.

XianXing: +1. Ship it!

PR #240, opened 2017-02-22
Correct suggested regularization values

Hydrotoast:
If default regularization is L2, the regularization constant should be less than 1. 

For the case of gradient descent: `new_param = old_param - eta * ( lambda * old_param + gradient)` where `eta` is the learning rate. Suppose `eta = 1` for simplicity. If `lambda = 1`, this would annihilate the parameter and only learn the gradient at each step. If `lambda > 1`, this would oscillate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,7 +249,7 @@ spark-submit \
   --output-directory "out" \
   --task "LOGISTIC_REGRESSION" \
   --num-iterations 50 \
-  --regularization-weights "0.1,1,10,100" \
+  --regularization-weights "0.001,0.01,0.1" \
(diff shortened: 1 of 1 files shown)


fastier-li:
@Hydrotoast - `eta` in our implementation is actually adaptive, rather than fixed. It has a decay schedule to try and avoid oscillations.

fastier-li:
At this point, we are not planning on merging this PR, as the training rate is actually adaptive, so we shouldn't get into oscillations.

fastier-li:
Closing after talking with Ankan and other scientists on our team.

PR #243, opened 2017-03-05
Removing several compile-time warnings

fastier-li:
As per title. This shouldn't affect any functionality, I just removed several compile-time warnings, so that it is easier to see the real errors at once. 

Also includes various minor cleanups.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 *~
-*.iml
 *.ipr
 *.iws
 .classpath
@@ -9,7 +8,6 @@
 .project
 .settings
 */bin
-*/build
 */out
 bin
 build
@@ -19,3 +17,5 @@ photon-ml/LOGISTIC*
 photon-ml/LINEAR*
 photon-ml/POISSON*
 photon-ml/SMOOTHED*
+derby.log
+metastore_db/
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
@@ -48,7 +48,8 @@ protected[ml] object Utils {
   }
 
   /**
-   * Get the feature key as a concatenation of name and term delimited by [[io.GLMSuite.DELIMITER]].
+   * Get the feature key as a concatenation of name and term delimited by
+   * [[com.linkedin.photon.ml.io.GLMSuite.DELIMITER]].
    *
    * @param name Feature name
    * @param term Feature term
@@ -66,9 +67,7 @@ protected[ml] object Utils {
    */
   def getFeatureNameFromKey(key: String, delimiter: String = GLMSuite.DELIMITER): String = {
     require(delimiter.r.findAllIn(key).length == 1, s"Provided input [$key] is not a valid feature key")
-    key.split(delimiter)
-      .lift(0)
-      .getOrElse("")
+    key.split(delimiter).headOption.getOrElse("")
   }
 
   /**
@@ -80,9 +79,7 @@ protected[ml] object Utils {
    */
   def getFeatureTermFromKey(key: String, delimiter: String = GLMSuite.DELIMITER): String = {
     require(delimiter.r.findAllIn(key).length == 1, s"Provided input [$key] is not a valid feature key")
-    key.split(delimiter)
-      .lift(1)
-      .getOrElse("")
+    key.split(delimiter).lift(1).getOrElse("")
   }
 
   /**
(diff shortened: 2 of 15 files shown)


.gitignore
```diff
@@ -1,5 +1,4 @@
 *~
-*.iml
```
ashelkovnykov: Why remove `.iml` files from ignore, I still have them for IDEA

photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
```diff
@@ -29,18 +29,19 @@ import com.linkedin.photon.ml.util.Logging
...
+
   val MEAN_ABSOLUTE_ERROR = "Mean absolute error"
   val MEAN_SQUARE_ERROR = "Mean square error"
   val ROOT_MEAN_SQUARE_ERROR = "Root mean square error"
   val AREA_UNDER_PRECISION_RECALL = "Area under precision/recall"
   val AREA_UNDER_RECEIVER_OPERATOR_CHARACTERISTICS = "Area under ROC"
   val PEAK_F1_SCORE = "Peak F1 score"
   val DATA_LOG_LIKELIHOOD = "Per-datum log likelihood"
-  val AIKAKE_INFORMATION_CRITERION = "Aikake information criterion"
+  val AKAIKE_INFORMATION_CRITERION = "Akaike information criterion"
```
ashelkovnykov: I swear I thought I had fixed this already...

.gitignore
fastier-li: The gitignore plugin in IDEA told me that .iml could go if we had .idea, but it didn't say anything about .iws or .ipr. .iws/.ipr seems to be used for "file based projects", so it might not be covered by ".idea".

photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
fastier-li: Yeah that's what I thought too. Oh well :-)

fastier-li:
All tests (unit and integration) pass on my box.

photon-client/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
```diff
@@ -294,7 +292,7 @@ protected[ml] object Utils {
   @throws(classOf[IllegalArgumentException])
   def getKeyFromMapOrElse[T](map: Map[String, Any], key: String, elseBranch: Either[String, T]): T = {
     map.get(key) match {
-      case Some(x: T) => x
+      case Some(x: Any) => x.asInstanceOf[T] // to avoid type erasure warning
```
joshvfleming: IMO leaving the warning in place is better here. This code does the same thing as before, but without any notification that it might behave unexpectedly. Maybe we could include a class tag so that the runtime can resolve the type? Also, use of `asInstanceOf` is generally frowned upon.

photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
```diff
@@ -64,40 +65,42 @@ object Evaluation extends Logging {
 
     // Compute regression facet metrics
     model match {
-      case r: Regression =>
+      case _: Regression =>
         val regressionMetrics = new RegressionMetrics(scoreAndLabel)
-        metrics ++= Map[String, Double](MEAN_ABSOLUTE_ERROR -> regressionMetrics.meanAbsoluteError,
-                                        MEAN_SQUARE_ERROR -> regressionMetrics.meanSquaredError,
-                                        ROOT_MEAN_SQUARE_ERROR -> regressionMetrics.rootMeanSquaredError)
```
joshvfleming: I think this is okay, but just want to point out that there's really nothing wrong with the original way. The warning in this case is a compiler bug.

photon-client/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
fastier-li: I like the idea of a class tag, and it is another suggested way to fix this issue. Let me try to go that route.

fastier-li: On the other hand, the overhead to using type tags in this case seems prohibitive. Let's keep the warning :-/

photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
fastier-li: It is, and it's apparently fixed ... only in 2.12, according what I could gather on the internet, which is disappointing. But again, the point is to try to make it clearer what is error and what is warning when looking at the results of the build.

fastier-li:
Verifying build, unit tests and integration tests offline since Travis is down at the moment.

PR #248, opened 2017-03-08
Fix Travis CI build, plus minor style changes.

ashelkovnykov:


fastier-li:
Looks good offline, and twice in a row in Travis.

PR #249, opened 2017-03-09
Jacoco and scalaStyle

fastier-li:
Reviving Jacoco and scalaStyle
Killing #33, which this one (#249) now replaces


fastier-li:
Abandoning this for now - will revisit later.

PR #255, opened 2017-03-20
Avro

fastier-li:
- Grouped everything Avro related in one place in the client.
- Tried to simplify `AvroFieldNames` but it will require further attention.
- Merged `AvroIOUtils` and `AvroUtils`, but that will also require further attention.

All in all, I hope Avro stuff is better localized in photon-client data.avro.


fastier-li:
If there are simple cleanups that I could do in this PR, I'm up for it, let me know.

photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
```diff
@@ -29,7 +29,7 @@ import com.linkedin.photon.ml.PhotonOptionNames._
 import com.linkedin.photon.ml.TaskType.TaskType
 import com.linkedin.photon.ml.constants.MathConst
 import com.linkedin.photon.ml.diagnostics.DiagnosticMode
-import com.linkedin.photon.ml.io.{FieldNamesType, GLMSuite, InputFormatType}
+import com.linkedin.photon.ml.photon_io.{FieldNamesType, InputFormatType}
```
joshvfleming: Underscores shouldn't be used in package names. Also, why are we distinguishing IO from Photon IO? The name "Photon" encompasses everything in the project, including GAME.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroUtils.scala
```diff
@@ -90,9 +228,9 @@ object AvroUtils {
    * @param record The input Avro record
    * @return The nameAndTerm parsed from the Avro record
    */
-  protected[ml] def readNameAndTermFromGenericRecord(record: GenericRecord): NameAndTerm = {
-    val name = Utils.getStringAvro(record, AvroFieldNames.NAME)
-    val term = Utils.getStringAvro(record, AvroFieldNames.TERM, isNullOK = true)
+  def readNameAndTermFromGenericRecord(record: GenericRecord): NameAndTerm = {
```
joshvfleming: Why not keep these as protected?

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroConstants.scala
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * Constants used in Avro related functions and objects only.
+ */
+object AvroConstants {
+
+  val DEFAULT_AVRO_FILE_NAME = "part-00000.avro"
+  val ID_INFO = "id-info"
+  val COEFFICIENTS = "coefficients"
+  val FIXED_EFFECT = "fixed-effect"
+  val RANDOM_EFFECT = "random-effect"
```
joshvfleming: I don't believe these last 4 are really avro-specific.

photon-client/src/main/scala/com/linkedin/photon/ml/Constants.scala
```diff
@@ -12,15 +12,28 @@
...
+  /**
+   * Wildcard character used for specifying the feature constraints. Only the term is allowed to be a wildcard normally
+   * unless one wants to apply bounds to all features in which case both name and term can be specified as wildcards.
+   * Currently, we do not support wildcards in name alone.
+   */
+  val WILDCARD = "*"
+
+  val INTERCEPT_NAME = "(INTERCEPT)"
+  val INTERCEPT_TERM = ""
+  val INTERCEPT_NAME_TERM: String = INTERCEPT_NAME + DELIMITER + INTERCEPT_TERM
```
joshvfleming: Would be better to use the `getFeatureKey` function here, rather than duplicating key construction. Also, why did it change from `INTERCEPT_KEY` to `INTERCEPT_NAME_TERM`? The name/term thing is a detail of how the key is constructed, but code downstream from this doesn't need to know that.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataReader.scala
```diff
@@ -279,7 +274,7 @@ object AvroDataReader {
           val nameAndTerm = AvroUtils.readNameAndTermFromGenericRecord(record)
           val featureKey = Utils.getFeatureKey(nameAndTerm.name, nameAndTerm.term)
 
-          featureKey -> Utils.getDoubleAvro(record, AvroFieldNames.VALUE)
+          featureKey -> Utils.getDoubleAvro(record, FieldNames.VALUE)
```
joshvfleming: Wasn't this removed?

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/README.md
```diff
@@ -0,0 +1,3 @@
+Refactorings needed here:
+- Simplify avro field names, stop having different names "label" and "response"
```
joshvfleming: Actually, the plan is to get rid of it entirely, and allow user to pass in the "label" column name.

photon-client/src/main/scala/com/linkedin/photon/ml/photon_io/README.md
```diff
@@ -0,0 +1 @@
+This package is slated for removal when photon disappears (and photon models are handled as Game fixed effect models).
```
joshvfleming: Photon is not disappearing. :) The the legacy GLMSuite, Driver, and related code is.

photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
fastier-li: The idea was to isolate "photon" from "game" stuff, this photon_io dir being on its way out when we finalize the transition to game.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroUtils.scala
fastier-li: I can put `protected` back - in the midst of the refactoring, it was messing with tests, then I forgot about it.

photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
fastier-li: Talked offline - we agreed to mark dirs/modules that are used only in "photon" but won't make it into "game" as `.deprecated`.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroConstants.scala
fastier-li: Right now, they are in use only in `ModelProcessingUtils`, which is itself in `data.avro`.

fastier-li: I'm going to leave them as they are now, but I am leaving notes and todos to rework this.

photon-client/src/main/scala/com/linkedin/photon/ml/Constants.scala
fastier-li: There was both KEY and NAME_TERM, and I did a global replace. I can do the opposite global replace. For `getFeatureKey`, it is in `Utils` which includes this file, `Constants`. That's the only reason why I was reluctant to use the function.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataReader.scala
fastier-li: That's yet another `FieldNames`, local to class `AvroDataReader` :-) I'll try to cleanup.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/README.md
fastier-li: Noted in README.

photon-client/src/main/scala/com/linkedin/photon/ml/photon_io/README.md
fastier-li: Yes - I am using "photon" to mean legacy here, as opposed to "game", but "photon" is still the product name. Maybe we should get rid of the name "game" when the legacy stuff is all removed.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ResponsePredictionFieldNames.scala
```diff
@@ -17,12 +17,7 @@ package com.linkedin.photon.ml.data.avro
 /**
  * Response prediction format fields names.
```
ashelkovnykov: Would it be worth documenting what the 'Response Prediction' schema is?

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ScoreProcessingUtils.scala
```diff
@@ -45,7 +45,7 @@ object ScoreProcessingUtils {
       val weight = Option(scoreAvro.getWeight).map(_.toDouble)
       val ids = scoreAvro.getMetadataMap.asScala.map { case (k, v) => (k.toString, v.toString) }.toMap
       val idsWithUid = uid match {
-        case Some(id) => ids + (AvroFieldNames.UID -> id)
+        case Some(id) => ids + (ResponsePredictionFieldNames.UID -> id)
```
ashelkovnykov: I think the `ScoreProcessingUtils` object is agnostic to the two different input schema types we have defined, so it should use `AvroFieldNames` instead. Regardless, it definitely shouldn't be using `ResponsePredictionFieldNames` as the `ScoringResultAvro` schema has the response field named "label", which would mean `TrainingExampleFieldNames` if not `AvroFieldNames`.

photon-client/src/test/scala/com/linkedin/photon/ml/data/avro/AvroDataReaderTest.scala
```diff
@@ -147,9 +147,9 @@ object AvroDataReaderTest {
     .record("testNameAndTermSchema")
     .namespace("com.linkedin.photon.ml.avro.data")
     .fields()
-    .name(AvroFieldNames.NAME).`type`().stringType().noDefault()
-    .name(AvroFieldNames.TERM).`type`().stringType().noDefault()
-    .name(AvroFieldNames.VALUE).`type`().doubleType().noDefault()
+    .name(ResponsePredictionFieldNames.NAME).`type`().stringType().noDefault()
+    .name(ResponsePredictionFieldNames.TERM).`type`().stringType().noDefault()
+    .name(ResponsePredictionFieldNames.VALUE).`type`().doubleType().noDefault()
```
ashelkovnykov: These tests don't actually take either of the two types of input Avro schemas Photon handles as input, so they should use AvroFieldNames instead.

photon-client/src/integTest/scala/com/linkedin/photon/ml/data/avro/ScoreProcessingUtilsTest.scala
```diff
@@ -31,22 +31,22 @@ class ScoreProcessingUtilsTest extends SparkTestUtils with TestTemplateWithTmpDi
       predictionScore = 1.0,
       label = Some(1.0),
       weight = Some(1.0),
-      idTypeToValueMap = Map(AvroFieldNames.UID -> "1", "id2" -> "2")),
+      idTypeToValueMap = Map(FieldNames.uid -> "1", "id2" -> "2")),
```
ashelkovnykov: Not sure what happened here, but `FieldNames` doesn't exist anymore - you got rid of it, so all of these should be `AvroFieldNames`

photon-client/src/main/scala/com/linkedin/photon/ml/data/README.md
```diff
@@ -1 +1,2 @@
-This directory contains various data readers/writers for various data formats (Avro, Parquet...)
\ No newline at end of file
+This directory contains various data readers for various data formats (Avro, Parquet...), where "data" means
+"training data", "validation data', "scoring data".
```
fastier-li: Simplified.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataReader.scala
```diff
@@ -271,7 +265,7 @@ object AvroDataReader {
     fieldNames
       .toSeq
       .flatMap { fieldName => Option(record.get(fieldName)) match {
-        case Some(recordList: JList[_]) => recordList.asScala.toSeq
+        case Some(recordList: JList[_]) => recordList.asScala.toSeq // this .toSeq intentional
```
fastier-li: Well, here - it would be appropriate. I don't feel like inserting a line for the comment. This comment was just supposed to be helpful because the IDE is complaining about that `toSeq`. I'll remove it.

photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -342,7 +259,7 @@ protected[ml] object IOUtils {
    * which can itself be either Success or Failure, wrapping an instance of the type returned by the lambda.
    * Failure.map(lambda) ignores lambda, and returns itself, but changing the contained type to the type
    * returned by the lambda (see scala.util.Try). Failure thus contains an exception, if one is thrown.
-   * On the last line, flatMap is used to avoid returning Try[Try[Unit]].
+   * On the last line, flatMap is used to avoid returning Try[Try[Unit] ].
```
fastier-li: It is actually correct Scala syntax, is it not? If I leave `]]` the IDE flags it in red because it thinks it's one of those (unreadable) comment reference thing. I removed the line explaining the semantics of `flatMap` altogether. We can assume somebody reading this knows what `flatMap` does.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ResponsePredictionFieldNames.scala
fastier-li: Not sure what you mean by "schema"? Avro schema?

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ResponsePredictionFieldNames.scala
```diff
@@ -17,12 +17,7 @@ package com.linkedin.photon.ml.data.avro
...
-  val name: String = "name"
-  val term: String = "term"
-  val value: String = "value"
-  val response: String = "response"
-  val offset: String = "offset"
-  val weight: String = "weight"
-}
+object ResponsePredictionFieldNames extends AvroFieldNames {
+
+  override val RESPONSE: String = "response" // do we really need a different name?
```
fastier-li: "Not liking them" is not a reasonable, rational argument. I "like" inline comments very much: they are real convenient to avoid disturbing the flow of the code. For this comment, it's the todo that we should not have "label" and "response". But it's captured somewhere else, so I'm ok to remove it here. I completely "reserve the right" to use inline comments elsewhere in general if really appropriate.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ResponsePredictionFieldNames.scala
ashelkovnykov: Yeah - documenting why we have these two different (but very similar) `AvroFieldNames` derived classes. Should we document it? Should we move them to some LinkedIn-internal codebase and a LinkedIn-internal Driver?

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ResponsePredictionFieldNames.scala
ashelkovnykov: The main crux of the argument is that the prevailing style as of today in the codebase is to not use inline comments, me not liking them is just additional context. Our style should be consistent throughout, and for things that don't affect the correctness of the code, it usually comes down to the personal preferences of the main contributors, as well as the historic impact of those who were contributors before. As far as I know, up to the present time that has not included inline comments.

I think this is specific issue is one of personal taste entirely - IMO correctly formatted comment blocks don't disturb the flow of the code, but rather help group it into finite chunks.

As for the substance of the comment: then make it a TODO. The current text is ambiguous, and in addition it's kind of "playing dumb". We know exactly why we need to have `AvroFieldNames` derived as such into two classes with different "response" field names. We can document it, or we can not - but we shouldn't leave comments like this one wondering why we're doing it and if we should.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ResponsePredictionFieldNames.scala
```diff
@@ -17,12 +17,7 @@ package com.linkedin.photon.ml.data.avro
...
+}
```
fastier-li: Fixed.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/TrainingExampleFieldNames.scala
```diff
@@ -17,12 +17,7 @@ package com.linkedin.photon.ml.data.avro
...
-  val features: String = "features"
-  val name: String = "name"
-  val term: String = "term"
-  val value: String = "value"
-  val response: String = "label"
-  val offset: String = "offset"
-  val weight: String = "weight"
+object TrainingExampleFieldNames extends AvroFieldNames {
+
+  override val RESPONSE: String = "label" // do we really need a different name?
```
fastier-li: Fixed.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroFieldNames.scala
```diff
@@ -19,14 +19,29 @@ package com.linkedin.photon.ml.data.avro
...
   val FEATURES: String = "features"
   val VALUE: String = "value"
-  val RESPONSE: String
+  val RESPONSE: String = "response"
   val OFFSET: String = "offset"
   val WEIGHT: String = "weight"
   val META_DATA_MAP: String = "metadataMap"
 }
 
+
```
fastier-li: Fixed.

photon-client/src/test/scala/com/linkedin/photon/ml/data/avro/AvroDataReaderTest.scala
fastier-li: Yeah - ain't going to happen with this design :-/

photon-client/src/integTest/scala/com/linkedin/photon/ml/data/avro/ScoreProcessingUtilsTest.scala
fastier-li: Indeed. Fixed.

PR #256, opened 2017-03-21
Add alternate scoring class for GAME training

ashelkovnykov:
- Replace KeyValueScore with 2 classes: one heavier containing ScoredGameDatum objects
  and one lighter containing Doubles
- Modify all classes that previously depended on KeyValueScore: classes derived from
  DatumScoringModel used for scoring use the heavier class, classes derived from Coordinate
  used for training use the lighter class
- Additional documentation

**Integration tests:**
Local integration tests (which are a bit more thorough than the Travis CI tests) pass.

**Performance testing:**
Locally the performance is slightly decreased: there is a 4 minute increase in training time (38 -> 42) for a 59 minute job (59 -> 63).

Training info:
1,545,820 samples, 616 features
3 iterations
L2 Regularization
No Normalization
1 fixed effect
3 random Effects
 - RE1:
   - 673373 unique IDs
   - Samples stats:
     - min = 1, mean = 2.29, max = 276
   - Features stats:
     - min = 34, mean = 36.71, max = 565
 - RE2:
   - 260 unique IDs
   - Samples stats:
     - min = 68, mean = 5,945, max = 144,446
   - Features stats:
     - min = 31, mean = 31, max = 31
 - RE3:
   - 288 unique IDs
   - Samples stats:
     - min = 68, mean = 5,367, max = 141,134
   - Features stats:
     - min = 31, mean = 31, max = 31

Distributed testing on a similar (but much larger) dataset revealed little change in Photon-ML runtime between #250 and this PR.

The change in RMSE was < 1e-5.


joshvfleming:
Looks good. I guess the gist would be something like:

- In a Coordinate, `score` means "score lite", with just the raw score.
- In Model, `score` means "score plus", with raw score and some metadata.

Do I have it right? This seems correct to me, since a `Coordinate` is an internal structure and a `Model` is something we'd export via the API.

ashelkovnykov:
@joshvfleming That's exactly it.

Right now there is one case of repeated methods: `evaluate` in the `Evaluator` trait. Ideally it would only deal with models/validation data, but we use it as a short-cut to log the objective loss during coordinate descent. Right now I've added a duplicate function that handles "lite" scores (`Double` only).

Possible alternatives are: convert those scores into "full" scores and use the same method call as the models (doing this for all the scores once per outer iteration would cause significant slow-down); remove the objective loss logging; leave it as is.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -134,21 +135,20 @@ object RandomEffectCoordinate {
...
    * @return The computed scores
    */
   protected[algorithm] def score(
     randomEffectDataSet: RandomEffectDataSet,
-    randomEffectModel: RandomEffectModel): KeyValueScore = {
+    randomEffectModel: RandomEffectModel): SimpleDataScores = {
 
     val activeScores = randomEffectDataSet
       .activeData
       .join(randomEffectModel.modelsRDD)
```
fastier-li: I'd like to see a comment somewhere (where most appropriate, maybe not right here), about passive and active. Can you add one?

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/RandomEffectOptimizationProblem.scala
```diff
@@ -43,27 +43,57 @@ protected[ml] class RandomEffectOptimizationProblem[Objective <: SingleNodeObjec
     val isTrackingState: Boolean)
   extends RDDLike {
 
+  /**
+   * Get the Spark context.
+   *
+   * @return The Spark context
+   */
   def sparkContext: SparkContext = optimizationProblems.sparkContext
```
fastier-li: sc? "sc" is very standard, and it has the merit of being short ...

fastier-li: We should replace alle `sparkContext` with `sc` across the board once and for all.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/scoring/DataScores.scala
```diff
@@ -0,0 +1,98 @@
...
+ * @param scores Data point scores, as described above
+ */
+abstract protected[ml] class DataScores[T, D <: DataScores[T, D]](scores: RDD[(Long, T)]) extends RDDLike {
+
+  /**
+   * Get the Spark context for the distributed scores.
+   *
+   * @return The Spark context
+   */
+  override def sparkContext: SparkContext = scores.sparkContext
```
fastier-li: Let's go with `sc` across the board.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/scoring/SimpleDataScores.scala
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.spark.rdd.RDD.rddToPairRDDFunctions
+
+/**
+ * The class used to track scored data points throughout training. The score objects are scores only, with no additional
+ * information.
+ *
+ * @param scores The scores consist of (unique ID, score) pairs as explained above.
+ */
+protected[ml] class SimpleDataScores(val scores: RDD[(Long, Double)])
```
fastier-li: `CoordinateDataScores` here and `ModelDataScores` for the other one? It's a more descriptive and useful, because it orientates right away, and also explains right away why we have two.

ashelkovnykov: Will do

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/RandomEffectOptimizationProblem.scala
ashelkovnykov: I think `sc` is good for the name of the `SparkContext` member, but I think methods to access `sc` should still be called `sparkContext`.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
ashelkovnykov: Take a look at `RandomEffectDataSet`, lines 30-37

fastier-li: Thanks!

fastier-li: I'm thinking - for comments like that, maybe a README in the module would be more useful: it would centralize "how it works" comments that can be useful when looking at various places in the code, whereas currently, you have to know where the comment is to even see it... :-(

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/RandomEffectOptimizationProblem.scala
fastier-li: Well, but in Scala, whether is a value or a function is intentionally made irrelevant... in other words, it makes no difference, so even "functions" should be `sc` IMHO. That one won't even be called with function call syntax, by the way.

ashelkovnykov: I have a strong rejection to having a method called `sc`. If I had it my way, I would have the Spark guys rename the members to `sparkContext` as well. One thing to note is that in the actual `RDD` source, they do what I mentioned above: member = `sc`, method = `sparkContext`.

You have a good point about consistency, but the Spark guys aren't consistent themselves and I hate the idea of a method named `sc`...

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
ashelkovnykov: I'm definitely open to copying the mentioned documentation to additional locations - where do you think is best?

ashelkovnykov:
Testing update - please see the OP

photon-lib/src/main/scala/com/linkedin/photon/ml/data/scoring/CoordinateDataScores.scala
```diff
@@ -0,0 +1,81 @@
...
+  extends DataScores[Double, CoordinateDataScores](scores) {
+
+  /**
+   * Generic method to combine two [[CoordinateDataScores]] objects.
+   *
+   * @param op The operator to combine two [[CoordinateDataScores]]
+   * @param that The [[CoordinateDataScores]] instance to merge with this instance
+   * @return A merged [[CoordinateDataScores]]
+   */
+  private def fullOuterJoin(op: (Double, Double) => Double, that: CoordinateDataScores): CoordinateDataScores =
```
joshvfleming: I know this is from a previous commit, but I think this function is misnamed now. It's doing a join + binary op.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/scoring/ModelDataScores.scala
```diff
@@ -0,0 +1,90 @@
...
+  extends DataScores[ScoredGameDatum, ModelDataScores](scores) {
+
+  /**
+   * Generic method to combine two [[ModelDataScores]] objects.
+   *
+   * @param op The operator to combine two [[ModelDataScores]]
+   * @param that The [[ModelDataScores]] instance to merge with this instance
+   * @return A merged [[ModelDataScores]]
+   */
+  private def fullOuterJoin(
```
joshvfleming: Same comment about naming here.

photon-lib/src/main/scala/com/linkedin/photon/ml/data/scoring/ModelDataScores.scala
```diff
@@ -0,0 +1,90 @@
...
+    that match {
+      case other: ModelDataScores =>
+        this.scores
+          .fullOuterJoin(other.scores)
+          .mapPartitions(iterator =>
+            Iterator.single(iterator.forall { case (_, (thisScoreOpt1, thisScoreOpt2)) =>
+              thisScoreOpt1.isDefined && thisScoreOpt2.isDefined && thisScoreOpt1.get.equals(thisScoreOpt2.get)
+            })
+          )
+          .filter(!_).count() == 0
```
joshvfleming: Similar to above:

```scala
  .mapPartitions(iterator =>
    Iterator.single(iterator.forall {
      case (_, (Some(thisScore), Some(thatScore))) => thisScore.equals(thatScore)
      case _ => false
    })         
  )
  .count(!_) == 0
```

photon-api/src/main/scala/com/linkedin/photon/ml/model/MatrixFactorizationModel.scala
```diff
@@ -54,20 +55,52 @@ class MatrixFactorizationModel(
...
-    MatrixFactorizationModel.score(dataPoints, rowEffectType, colEffectType, rowLatentFactors, colLatentFactors)
+  private def isSameLatentFactors(
+      latentFactors1: RDD[(String, Vector[Double])],
+      latentFactors2: RDD[(String, Vector[Double])]): Boolean = {
+
+    latentFactors1.fullOuterJoin(latentFactors2).mapPartitions(iterator =>
+      Iterator.single(iterator.forall { case (_, (factor1, factor2)) =>
+        factor1.isDefined && factor2.isDefined && factor1.get.equals(factor2.get)
+      })
+    ).filter(!_).count() == 0
```
ashelkovnykov: I don't think that's possible - do you have a link to the docs of a count function that takes a filter as input?

photon-api/src/main/scala/com/linkedin/photon/ml/model/MatrixFactorizationModel.scala
```diff
@@ -220,6 +252,55 @@ object MatrixFactorizationModel {
...
+   * Score the given GAME data points with the row and column latent factors.
+   *
+   * @param dataPoints The GAME data points
+   * @param rowEffectType The type of row effect used to retrieve row effect id from each GAME data point
+   * @param colEffectType The type of column effect used to retrieve column effect id from each GAME data point
+   * @param rowLatentFactors The row latent factors
+   * @param colLatentFactors The col latent factors
+   * @return The computed scores
+   */
+  private def scoreLite(
```
fastier-li: `scoreForCoordinateDescent` ? - it would have the merit of telling right away where it is used, so it would be good for orientation in the code?

photon-api/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
```diff
@@ -232,6 +269,54 @@ object RandomEffectModel {
...
+      }
+      .partitionBy(hashPartitioner)
+      .zipPartitions(modelsRDD.partitionBy(hashPartitioner))(
+        (dataIt, modelIt) => {
+          val lookupTable = modelIt.toMap
+          dataIt.map {
+            case (id, (uid, datum)) =>
+              val score = lookupTable.get(id).map(model =>
+                model.computeScore(datum.featureShardContainer(featureShardId))).getOrElse(0.0)
+              (uid, score)
```
fastier-li: +1

photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
```diff
@@ -439,7 +439,7 @@ class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
     args += HEAVY_MAX_NUM_ITERATIONS_FOR_LBFGS.toString
     MockDriver.runLocally(
       args = args.toArray,
-      sparkContext = sc,
+      sc = sc,
```
fastier-li: _Vitius repetitionis_

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/NameAndTerm.scala
```diff
@@ -14,37 +14,51 @@
...
+  override def toSummaryString: String = s"name: $name, term: $term"
 
   /**
+   * Returns a string representation of the [[NameAndTerm]]
    *
-   * @param that
-   * @return
+   * @return A string representation of the name-term pair
+   */
+  override def toString: String = s"$name\t$term"
```
fastier-li: If the name or term has spaces, we might not know what is what with this format?

ashelkovnykov: Only if it they contain tabs - this is only used when creating the feature bags so it doesn't actually need to be human readable. If you've got suggestions, I can implement them now - otherwise I'd prefer to keep out of the Avro mess

PR #258, opened 2017-03-23
Store per-model Game config snapshots in a struct rather than String

joshvfleming:
In some cases (e.g. hyperparameter tuning) we need programmatic access to the configuration of the Game optimizers for a trained model. This exposes those settings in a struct and provides `toString` for the previous usages.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/FactoredRandomEffectOptimizationConfiguration.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/FactoredRandomEffectOptimizationConfiguration.scala
@@ -1,0 +1,27 @@
+/*
+ * Copyright 2017 LinkedIn Corp. All rights reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.optimization.game
+
+/**
+ * Configuration object for a factored random effect
+ *
+ * @param randomEffectOptimizationConfiguration the random effect optimization configuration
+ * @param latentFactorOptimizationConfiguration the latent factor optimization configuration
+ * @param mfOptimizationConfiguration the matrix factorization optimization configuration
+ */
+protected[ml] case class FactoredRandomEffectOptimizationConfiguration(
+    randomEffectOptimizationConfiguration: GLMOptimizationConfiguration,
+    latentFactorOptimizationConfiguration: GLMOptimizationConfiguration,
+    mfOptimizationConfiguration: MFOptimizationConfiguration)
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameModelOptimizationConfiguration.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameModelOptimizationConfiguration.scala
@@ -25,8 +25,7 @@ package com.linkedin.photon.ml.optimization.game
 case class GameModelOptimizationConfiguration(
     fixedEffectOptimizationConfiguration: Map[String, GLMOptimizationConfiguration],
     randomEffectOptimizationConfiguration: Map[String, GLMOptimizationConfiguration],
-    factoredRandomEffectOptimizationConfiguration: Map[String,
-      (GLMOptimizationConfiguration, GLMOptimizationConfiguration, MFOptimizationConfiguration)]) {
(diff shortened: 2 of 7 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameOptimizationConfigSnapshot.scala
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+  * Represents the complete Game optimization configuration as it appeared for a single run with a particular setting
+  * for each hyperparameter.
+  *
+  * @param fixedEffectOptimizationConfiguration the optimization configuration for each fixed effect
+  * @param randomEffectOptimizationConfiguration the optimization configuration for each random effect
+  * @param factoredRandomEffectOptimizationConfiguration the optimization configuration for each factored random effect
+  */
+case class GameOptimizationConfigSnapshot(
```
fastier-li: Do we need `Snapshot`?

joshvfleming: I couldn't think of a better name, but I'm open to suggestions. I don't want to call it `GameOptimizationConfiguration`, because that implies that it's a full configuration object, which it's not.

fastier-li: Would it be hurtful to pass the full configuration object?

joshvfleming: We can't pass the full configuration object here, because we need the part that's specific to the current model (i.e. the specific regularization parameter values).

fastier-li: Is that a clue for naming? Should it be e.g. `CurrentModelOptimizationConfiguration`?

joshvfleming: Hmm, using "current" in the name like that is a little awkward. Also, it doesn't capture that the object contains all the Game components. How about `GameModelSpecificOptimizationConfig`?

ashelkovnykov: Out of everything pitched so far, I prefer the `Snapshot` version. Maybe `TransitionalGameOptimizationConfiguration`?

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameOptimizationConfigSnapshot.scala
```diff
@@ -0,0 +1,37 @@
...
+    fixedEffectOptimizationConfiguration: Map[String, GLMOptimizationConfiguration],
+    randomEffectOptimizationConfiguration: Map[String, GLMOptimizationConfiguration],
+    factoredRandomEffectOptimizationConfiguration: Map[String,
+      (GLMOptimizationConfiguration, GLMOptimizationConfiguration, MFOptimizationConfiguration)]) {
```
ashelkovnykov: 1. I think these val names should be plural
2. If we don't have a wrapper object for `FactoredRandomEffectOptimizationConfiguration`, we should (or at least a custom type in `Types` to make dealing with it easier)

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameOptimizationConfigSnapshot.scala
fastier-li: The problem for me with `Snapshot` is that it evokes a design pattern (memento), or that it sounds like a "snapshot" for persistence/high availability. In other words, the term is alrady taken with different meaning(s), so misleading here. How about `GameModelOptimizationConfig`?

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameOptimizationConfigSnapshot.scala
fastier-li: Also, maybe we should shorten all `Configuration` to `Config`. Long variable names like that are a laughingstock in some teams, AND they crowd the code. `REOptimConfig` could be acceptable (and I know there are some who maintain names should be spelled out in full).

joshvfleming: I used singular here because we use plural elsewhere to signify an array of these.

joshvfleming: @fastier-li If we do that, it should happen across the board in a different PR.

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameOptimizationConfigSnapshot.scala
joshvfleming: I'm going to go with `GameModelOptimizationConfig`. It's not quite right, but IMO it's close enough and carrying this discussion any further is beyond my tolerance for bikeshedding.

PR #259, opened 2017-03-28
Change caching levels of RDDs

namitk:
Based on experiments, changing levels of RDDs in scoring driver as follows:
1. gameDataSet: disk_only
2. gameModel: memory_only
3. scores: memory_and_disk

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
@@ -115,7 +115,7 @@ class Driver(val params: Params, val sparkContext: SparkContext, val logger: Log
     // Load the model from HDFS, ignoring the feature index loader
     val (gameModel, _) =
       ModelProcessingUtils.loadGameModelFromHDFS(Some(featureShardIdToIndexMapLoader), gameModelInputDir, sparkContext)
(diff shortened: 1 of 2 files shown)


namitk:
The caching levels previously were all disk_only. Changing some of them to memory_and_disk or memory_only should lead to a boost in performance unless the memory config is so tight that they keep needing to be evicted from memory constantly. I, therefore, believe this should be a safe change to make.

For the particular use case, we are able to cut the run time down from 6-8 hours to near 4 hours. The caveat, however, is that there are multiple changes to go along with it (increase task parallelism, increase timeouts, increase memory fraction etc) None of these are related to the caching levels but I mentioned these because I don't want to suggest that the reduction in runtime is purely due to caching level changes.

One thing in particular that the change in caching levels is helping with is the issue of filling up tmp space on the nodes. When we had all of them being cached to disk, we were seeing instances where the executors would be killed due to tmp space being full.

fastier-li:
Namit - is there a useful doc/best practice guideline to extract from your experience? Let's do that change, and distill a best practice for tuning that we can put in github somewhere.

namitk:
In terms of deciding caching levels, it's really an experimental process and there isn't really a best practice to set. The same RDD that is okay to cache in memory with a high executor memory configuration can need to be cached using disk with a lower setting so there isn't a one-rule-fits-all kind of an answer to decide how an RDD should be cached.

Note also that there are serialized and deserialized versions of each of these levels, serialization decreases memory footprint but increases computational cost (since deserialization would be needed to read). Deserialization has a higher memory footprint but lower computational cost.

PR #260, opened 2017-03-29
Starting work on flexible input columns names for Game

fastier-li:


photon-client/src/main/scala/com/linkedin/photon/ml/Constants.scala
```diff
@@ -26,16 +26,16 @@ object Constants {
    *
    * WARNING: This is not visible in println!
    */
-  val DELIMITER = "\u0001"
+  val DELIMITER: String = "\u0001"
```
joshvfleming: This is overkill. No need for a type annotation on vals when the type is clear.

fastier-li: IDEA says it's the required style for public members. I can take it out.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
```diff
@@ -242,6 +251,16 @@ object Params {
...
+      opt[String]("input-columns-names")
+        .text(s"A map of the input columns names, in format: response:label|offset:offset|uid:uid ... ")
+        .foreach(x =>
+          x.split("\\|")
+            .foreach { line =>
+              val Array(key, value) = line.split(":").map(_.trim)
+              params.inputColumnsNames.updated(InputColumnsNames.withName(key), value)
+            }
+        )
+
```
joshvfleming: IMO we should have an individual parameter for each of these.

fastier-li: You can set any of them, you don't have to set all of them - one parameter for each is kind of annoying, is it not? I can do it - but I don't want to add 5 arguments to each function that uses those params...

ashelkovnykov: I prefer the map. It's more flexible for extension, should we need to extend it. @joshvfleming What's your reason for preferring one parameter per replacement column name?

joshvfleming: I prefer having individual parameters because there's a small number of them, and our algorithms depend on them directly. I.e. we'd never add or remove one without some change to the code. So a dynamic structure seems out of place to me.

The way they handle this in spark.ml is to have traits for each column that mix the setters/getters into the Estimator. E.g.:

```scala
estimator
  .setLabelCol("response")
  .setWeightCol("weight")
  .setOffsetCol("offset")
```

ashelkovnykov: Hmm... It may be worth doing so then, since our API is heading towards unity with spark.ml

fastier-li: Even if we have those `set` methods on the `GameEstimator`, we don't have to expose that to the user with 5 distinct parameters. We can still have a user API with a "convenient" way to set some or all of the column names, ship those in a `Map` to the `Driver` and have the `Driver` call these setters with the contents of the `Map`. That reduces our internal code clutter, while still using spark.ml-style setters on `GameEstimator`.

fastier-li: Also, `GameParams` is already passed to `GameEstimator`, so if the column names are in `GameParams`, we probably won't call setters on `GameEstimator` to set them.

ashelkovnykov: These params only have significance to Avro, so as a part of the client I don't think we need 5 individual params. How we handle the API is a different matter. I'd prefer the dynamic map

PR #261, opened 2017-03-30
Verifying sample weights for training and validation data (optionally)

fastier-li:
We had a case where the weights for all the training samples were zero, causing the optimizer to not even start. 

This PR introduces code to check that the weights are all >= 0 and that at least one is > 0. Since this test could be costly, there is a new boolean parameter (`checkData`) to turn this check on or off.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
@@ -19,6 +19,7 @@ import scala.util.{Failure, Success, Try}
 import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkContext
 import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.functions.col
 import org.slf4j.Logger
 
 import com.linkedin.photon.ml.Types.{FeatureShardId, SparkVector}
@@ -63,12 +64,22 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
     val featureIndexMapLoaders = Timed("prepare features") {
       prepareFeatureMaps()
     }
-    val trainingData = Timed("read training data") {
+    val trainingDataToVerify = Timed("read training data") {
       readTrainingData(featureIndexMapLoaders)
     }
-    val validationData = Timed("read validation data") {
+
+    val trainingData = Timed("verify training data") {
+      preprocessData(trainingDataToVerify)
+    }
+
+    val validationDataToVerify = Timed("read validation data") {
       readValidationData(featureIndexMapLoaders)
     }
+
+    val validationData = Timed("verify validation data") {
+      validationDataToVerify.tap(preprocessData(_))
+    }
+
     val featureShardStats = Timed("calculate statistics for each feature shard") {
       calculateAndSaveFeatureShardStats(trainingData, featureIndexMapLoaders)
     }
@@ -158,6 +169,31 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
             params.featureShardIdToFeatureSectionKeysMap,
             params.minPartitionsForValidation)
     }
+
+  /**
+   * Preprocess data.
+   *
+   * This can be used to transform a data set (training or validation data), or to validate it with various checks.
+   *
+   * For now, we just check that the data contains at least a sample with non-zero weight). We throw out samples that
+   * have a weight less than or equal to 0. If we don't find at least one sample with a strictly positive weight, we
+   * throw an IllegalArgumentException, otherwise we return the filtered data set.
+   *
+   * @note this is somewhat expensive (it iterates over the whole data set), so we can turn it off
+   *
+   * @param data The data to check
+   * @return A data set where zero-weight samples have been removed, or an exception if we couldn't find at least one
+   *         sample with a strictly positive weight
+   */
+  protected[training] def preprocessData(data: DataFrame): DataFrame = {
+    if (params.checkData) {
+      data
+        .filter(col("weight") > 0.0)
+        .tap(df => require(df.count > 0, "Didn't find any data point with a weight >= 0. Aborting."))
+    } else {
+      data
+    }
+  }
 
   /**
    * Calculate basic statistics (same as spark-ml) on a DataFrame.
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameParams.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameParams.scala
@@ -16,7 +16,6 @@ package com.linkedin.photon.ml.estimators
 
 import scopt.OptionParser
 
-import com.linkedin.photon.ml.PhotonOptionNames._
 import com.linkedin.photon.ml.TaskType
 import com.linkedin.photon.ml.TaskType.TaskType
 import com.linkedin.photon.ml.Types._
@@ -172,6 +171,11 @@ class GameParams extends FeatureParams with PalDBIndexMapParams with EvaluatorPa
    */
   var applicationName: String = "Game-Full-Model-Training"
 
+  /**
+   * Turn on additional tests on the training and validation data
+   */
+  var checkData: Boolean = false
+
   override def toString: String =
     s"trainDirs: ${trainDirs.mkString(", ")}\n" +
       s"trainDateRangeOpt: $trainDateRangeOpt\n" +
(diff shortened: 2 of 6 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -160,6 +171,28 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
...
+   *
+   * Checks performed
+   * ================
+   * - all weights >= 0.0 and at least 1 weight > 0
+   *
+   * @param data The data to check
+   */
+  protected[training] def checkData(data: DataFrame): DataFrame =
+
+    if (params.checkData)
```
ashelkovnykov: Could you please add some brackets around the body of this `if` - I'd prefer we do that just to be clear and safe (if we're not using the `if` as a ternary operator)

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -63,12 +64,22 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
...
       prepareFeatureMaps()
     }
-    val trainingData = Timed("read training data") {
+    val trainingDataToVerify = Timed("read training data") {
       readTrainingData(featureIndexMapLoaders)
     }
-    val validationData = Timed("read validation data") {
+
+    val trainingData = Timed("verify training data") {
+      checkData(trainingDataToVerify)
```
joshvfleming: The name `checkData` is a little misleading here, because this is actually modifying in addition to checking.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -160,6 +171,28 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
...
+   * ================
+   * - all weights >= 0.0 and at least 1 weight > 0
+   *
+   * @param data The data to check
+   */
+  protected[training] def checkData(data: DataFrame): DataFrame =
+
+    if (params.checkData)
+      data
+        .filter(col("weight") >= 0.0)
```
ashelkovnykov: Didn't you want to make this check against epsilon?

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
fastier-li: Very true. Let me change that.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -160,6 +171,28 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
     }
 
   /**
+   * Run various checks on data (training or validation. Throws a RuntimeException if any of the checks does not pass.
+   *
+   * @note this is somewhat expensive, so we can turn it off
+   *
+   * Checks performed
+   * ================
+   * - all weights >= 0.0 and at least 1 weight > 0
```
ashelkovnykov: So just `at least 1 weight > 0` - that's what it decomposes into, unless we were explicitly checking for weights < 0

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -160,6 +171,28 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
...
+   *
+   * @param data The data to check
+   */
+  protected[training] def checkData(data: DataFrame): DataFrame =
+
+    if (params.checkData)
+      data
+        .filter(col("weight") >= 0.0)
+        .tap { df => if (df.count == 0)
```
ashelkovnykov: Could you please move the start of the `if` statement to the next line? Think of it like your method header and body separation style preference

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
joshvfleming: Missing braces.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameParams.scala
```diff
@@ -481,6 +489,10 @@ object GameParams {
         .text("An optional directory to output statistics about the training data")
         .foreach(x => params.summarizationOutputDirOpt = Some(x))
 
+      opt[Boolean](CHECK_DATA)
+        .text(s"Whether to check the data extensively or not (time consuming), default: ${defaultParams.checkData}")
+        .foreach(x => params.checkData = x)
```
ashelkovnykov: Should this be part of a greater 'data validation' parameter (like in Photon)? In either case, I think the parameter text should be more descriptive (what constitutes extensively, why is it time consuming?)

photon-client/src/main/scala/com/linkedin/photon/ml/io/deprecated/GLMSuite.scala
```diff
@@ -344,13 +344,22 @@ class GLMSuite(
           1
         }
 
+      assert(weight >= 0)
```
ashelkovnykov: Do we want to fail on negative weights, or do we want to filter them? If the former, this should be a `require` instead of an assert. If the latter, do we want to do so silently or log info about it?

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -160,6 +171,28 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
...
+          throw new RuntimeException("Didn't find any data point with a weight >= 0. Aborting.")
+        }
```
joshvfleming: How about:

```scala
  .tap(df => require(df.count > 0, "Didn't find any data point with a weight >= 0. Aborting."))
```

photon-client/src/main/scala/com/linkedin/photon/ml/io/deprecated/GLMSuite.scala
```diff
@@ -344,13 +344,22 @@ class GLMSuite(
...
-      iter.map { r => parseAvroRecord(r, map) }
-    }
+    val labeledPoints = avroRDD
+      .mapPartitions { iter =>
+        val map = _indexMapLoader.indexMapForRDD()
+        iter.map { r => parseAvroRecord(r, map) }
+      }
+      .filter(aPoint => aPoint.weight > 0.0)
+
+    // TODO: This is caught later when we check that we have at least 1 training sample
```
ashelkovnykov: Add a comment about where it's caught later (Photon Driver.prepareTrainingData; note that the same method is called again from Driver.prepareValidatingData and count is NOT called there)

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
joshvfleming: Shouldn't it be:

```scala
  .filter(col("weight") > 0.0)
```

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
fastier-li: Don't really need braces for a 1 liner function?

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
fastier-li: Looks nice!

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
fastier-li: I'm aching to transform that if then else into something more "functional" looking.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
fastier-li: I am explicitly checking for weights < 0. We need to.

photon-client/src/main/scala/com/linkedin/photon/ml/io/deprecated/GLMSuite.scala
fastier-li: Yes, we want to fail on negative weights.

photon-client/src/main/scala/com/linkedin/photon/ml/io/deprecated/GLMSuite.scala
fastier-li: Done.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
joshvfleming: The databricks style guide requires braces unless you can fit the entire expression on one line:

https://github.com/databricks/scala-style-guide#curly

PR #264, opened 2017-04-06
Add user-defined input columns names for Game

fastier-li:
Added the option for users to specify input column names.
Cleaned up some warning messages at compile time.


photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
```diff
@@ -57,6 +57,8 @@ class GameEstimatorTest extends SparkTestUtils with GameTestUtils {
    *       is specified in [[GameParams.featureShardIdToInterceptMap]]. This happens in
    *       [[GameDriver.prepareFeatureMapsDefault]], and there only. Here, we have to setup an intercept manually,
    *       otherwise [[GameEstimator]] learns only a dependence on the features.
+   *
+   * KEEP THIS DIDACTIC EXAMPLE
```
ashelkovnykov: Don't forget to delete this

photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
```diff
@@ -211,20 +213,23 @@ class GameEstimatorTest extends SparkTestUtils with GameTestUtils {
...
     val data = getData(trainPath, featureSectionMap)
 
     val partitioner = new LongHashPartitioner(data.rdd.partitions.length)
-    val gameDataSet = GameConverters.getGameDataSetFromDataFrame(
-      data,
-      featureSectionMap.keys.toSet,
-      idTypeSet,
-      isResponseRequired = true)
-      .partitionBy(partitioner)
+    val GameDataSet =
```
ashelkovnykov: The name case was correct beforehand

photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
```diff
@@ -241,19 +246,22 @@ class GameEstimatorTest extends SparkTestUtils with GameTestUtils {
...
       val featureSectionMap = fixedAndRandomEffectFeatureSectionMap
       val data = getData(trainPath, featureSectionMap)
       val partitioner = new LongHashPartitioner(data.rdd.partitions.length)
-      val gameDataSet = GameConverters.getGameDataSetFromDataFrame(
-        data,
-        featureSectionMap.keys.toSet,
-        idTypeSet,
-        isResponseRequired = true)
-        .partitionBy(partitioner)
+      val GameDataSet =
```
ashelkovnykov: Same as above

photon-client/src/main/scala/com/linkedin/photon/ml/Constants.scala
```diff
@@ -26,16 +26,16 @@ object Constants {
    *
    * WARNING: This is not visible in println!
    */
-  val DELIMITER = "\u0001"
+  val DELIMITER: String = "\u0001"
```
ashelkovnykov: I don't think explicit types are necessary here, it's clear that all of these are `String`

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
```diff
@@ -222,7 +225,7 @@ object Driver {
 
   /**
    *
-   * @param outputDir
+   * @param outputDir Directory where the scores will be written
```
ashelkovnykov: No it's not!

photon-client/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
```diff
@@ -19,55 +19,47 @@ import org.apache.spark.rdd.RDD
...
+      featureShards: Set[String],
+      idTypeSet: Set[String],
+      isResponseRequired: Boolean,
+      inputColumnsNames: Map[InputColumnsNames.Value, String]): RDD[(Long, GameDatum)] = {
 
-    // Add unique id
     val recordsWithUniqueId = data.withColumn(UNIQUE_ID_COLUMN_NAME, monotonicallyIncreasingId)
+    // Convert the (possibly partially populated) map of user-specified input columns names to a smaller array of
+    // strings, but follow the order of the fields in InputColumnsNames so that we can use InputColumnsNames field
+    // indexes to access user-defined column names directly
```
ashelkovnykov: While the `Map` COULD be partially updated (since it's just a `Map`), in practice it should never be so, since it's initialized with the default names of each field in `Params`.

photon-client/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
```diff
@@ -19,55 +19,47 @@ import org.apache.spark.rdd.RDD
...
+    val inputColumnsNamesArray = inputColumnsNames.toArray.map { case (k, v) => (k.id, v) }.sortBy(_._1).map(_._2)
```
ashelkovnykov: This `Array` should be broadcast

joshvfleming: Could we hide these details behind a map-like data structure? That way you don't need to worry about the details of keeping things in order while reading this consuming code.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameParams.scala
```diff
@@ -494,6 +503,16 @@ object GameParams {
           s"default: ${defaultParams.checkData}")
         .foreach(x => params.checkData = x)
 
+      opt[String](INPUT_COLUMNS_NAMES)
+        .text(s"A map of the input columns names, in format: response:label|offset:offset|uid:uid ... ")
+        .foreach(x =>
+          x.split("\\|")
+            .foreach { line =>
+              val Array(key, value) = line.split(":").map(_.trim)
```
joshvfleming: This will generate a confusing `MatchError` in the case of a malformed entry (e.g. `"response:label:hello"`). It would be better to match on this and generate a more helpful error in the case that it's wrong, e.g.:

```scala
  .foreach { line =>
    line.split(":").map(_.trim) match {
      case Array(key, value) => params.inputColumnsNames.updated(InputColumnsNames.withName(key), value)
      case other => s"Whoops, couldn't interpret '$other'"
    }
  }
```

photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
fastier-li: Well - I'd like to keep this, so that this test is not removed later, or changed, and we lose the pedagogical aspect of this code. It is extremely useful to follow step-by-step to see how to set up the `GameEstimator`, and we don't have an equivalent anywhere else. The other places are much harder to read.

photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
fastier-li: Fixed. Not sure why that was even changed. Must have been a global replace.

photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
fastier-li: Thanks!

photon-client/src/main/scala/com/linkedin/photon/ml/Constants.scala
fastier-li: According to our code style settings, IDEA complains that `INTERCEPT_KEY` just a few lines below should have an explicit type. So, we end up with that one `: String`, the others without. Looks non-uniform.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Driver.scala
fastier-li: Fixed (and improved whole comment too, and a few more).

photon-client/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorTest.scala
ashelkovnykov: Would a better solution be to have an 'examples' module or folder?

photon-client/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
fastier-li: Fixed comment after talking offline with Alex.

photon-client/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
fastier-li: Yes and no. We could, but it costs. An `Array[String]` is smaller and can be addressed with indexes. On the other hand, besides being details, this is also brittle... Let me think about it.

joshvfleming: We could keep the `Array` implementation but give it a map-like interface. This shouldn't cost any more memory or add to the lookup time.

fastier-li: Well - I think it's too complicated with an array, and we don't know if the gain is there - the idea was that indexing through an array was faster than a map, since we are going to look up names a lot, for each row of the data. But we can wait and see on performance. Alex is adamant that he wants a broadcast, so I guess I'll do a broadcast or a `Map[String, String]` in the end.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameParams.scala
fastier-li: Done.

PR #265, opened 2017-04-08
Bumping version of John Rengelman's shadow plugin to 1.2.4

ashelkovnykov:
We need this for LinkedIn internal reasons.


namitk:
I am the consumer of this change internally. LGTM

Issue #267, opened 2017-04-11
Track IDs of random effects that fail to converge


ashelkovnykov:
When optimization tracking is enabled, random effects currently track and summarize the convergence reason (or lack thereof) for each ID. The IDs that failed to converge should be logged and output at the end of training to aid tracking bugs.

It's also important to question: what causes models to fail to converge - can we detect it before training through data validation alone?

PR #268, opened 2017-04-12
Fix bug that incorrectly logged all random effect models as not converging

ashelkovnykov:
- The RandomEffectOptimizationTracker was initialized using OptimizationStateTrackers the RDD of SingleNodeOptimizationProblems stored in a RandomEffectOptimizationProblem after optimization - however, this RDD is persisted to disk, so it was pulling the trackers in their original empty state
- Added a 'smartMaterialize' method to the RDDLike trait to only materialize unmaterialized RDDs

These changes had no significant impact on performance locally (5 second speed-up on a  16 minute job). Performance tests on the cluster did not reveal any significant performance change(+/- 100 seconds of training for a 90 minute job with 50 minutes of training).

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
@@ -104,28 +104,39 @@ object RandomEffectCoordinate {
       randomEffectOptimizationProblem: RandomEffectOptimizationProblem[Function],
       randomEffectModel: RandomEffectModel): (RandomEffectModel, Option[RandomEffectOptimizationTracker]) = {
 
-    val updatedModels = randomEffectDataSet
+    val updatedModelsAndTrackers = randomEffectDataSet
       .activeData
       .join(randomEffectOptimizationProblem.optimizationProblems)
       .join(randomEffectModel.modelsRDD)
       .mapValues {
         case (((localDataSet, optimizationProblem), localModel)) =>
           val trainingLabeledPoints = localDataSet.dataPoints.map(_._2)
+          val updatedModel = optimizationProblem.run(trainingLabeledPoints, localModel)
+          val stateTrackers = optimizationProblem.getStatesTracker
 
-          optimizationProblem.run(trainingLabeledPoints, localModel)
+          (updatedModel, stateTrackers)
       }
+      .setName(s"Updated models and state trackers for random effect ${randomEffectDataSet.randomEffectType}")
+      .persist(StorageLevel.VERY_FREQUENT_REUSE_RDD_STORAGE_LEVEL)
     val updatedRandomEffectModel = randomEffectModel
-      .update(updatedModels)
-      .setName(s"Updated random effect model")
-    val optimizationTracker = if (randomEffectOptimizationProblem.isTrackingState) {
-        val stateTrackers = randomEffectOptimizationProblem.optimizationProblems.map(_._2.getStatesTracker.get)
+      .update(updatedModelsAndTrackers.mapValues(_._1))
+      .setName(s"Updated models for random effect ${randomEffectDataSet.randomEffectType}")
+      .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
+      .materialize()
+    val optimizationTracker: Option[RandomEffectOptimizationTracker] =
+      if (randomEffectOptimizationProblem.isTrackingState) {
+        val stateTrackers = updatedModelsAndTrackers.map(_._2._2.get)
         val randomEffectTracker = new RandomEffectOptimizationTracker(stateTrackers)
-          .setName(s"Random effect optimization tracker")
+          .setName(s"State trackers for random effect ${randomEffectDataSet.randomEffectType}")
           .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
           .materialize()
 
         Some(randomEffectTracker)
-      } else None
+      } else {
+        None.asInstanceOf[Option[RandomEffectOptimizationTracker]]
+      }
+
+    updatedModelsAndTrackers.unpersist()
 
     (updatedRandomEffectModel, optimizationTracker)
   }
@@ -138,7 +149,6 @@ object RandomEffectCoordinate {
    *
    * @note The score is the dot product of the model coefficients with the feature values (in particular, does not go
    *       through non-linear link function in logistic regression!).
-   *
    * @param randomEffectDataSet The active dataset to score
    * @param randomEffectModel The model to score the dataset with
    * @return The computed scores
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/FixedEffectDataSet.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/FixedEffectDataSet.scala
@@ -42,6 +42,7 @@ protected[ml] class FixedEffectDataSet(
    * @return An updated dataset with scores added to offsets
    */
   override def addScoresToOffsets(scores: CoordinateDataScores): FixedEffectDataSet = {
+
     val updatedLabeledPoints = labeledPoints
       .leftOuterJoin(scores.scores)
       .mapValues { case (LabeledPoint(label, features, offset, weight), scoredDatumOption) =>
@@ -67,7 +68,9 @@ protected[ml] class FixedEffectDataSet(
    * @return This object with the name of [[labeledPoints]] assigned
    */
   override def setName(name: String): FixedEffectDataSet = {
+
     labeledPoints.setName(name)
+
     this
   }
 
@@ -79,7 +82,9 @@ protected[ml] class FixedEffectDataSet(
    * @return This object with the storage level of [[labeledPoints]] set
    */
   override def persistRDD(storageLevel: StorageLevel): FixedEffectDataSet = {
+
     if (!labeledPoints.getStorageLevel.isValid) labeledPoints.persist(storageLevel)
+
     this
   }
 
@@ -89,7 +94,9 @@ protected[ml] class FixedEffectDataSet(
    * @return This object with [[labeledPoints]] marked non-persistent
    */
   override def unpersistRDD(): FixedEffectDataSet = {
+
     if (labeledPoints.getStorageLevel.isValid) labeledPoints.unpersist()
+
     this
   }
 
@@ -99,7 +106,9 @@ protected[ml] class FixedEffectDataSet(
    * @return This object with [[labeledPoints]] materialized
    */
   override def materialize(): FixedEffectDataSet = {
-    labeledPoints.count()
+
+    materializeOnce(labeledPoints)
+
     this
   }
 
@@ -109,10 +118,12 @@ protected[ml] class FixedEffectDataSet(
    * @return A String representation of the dataset
    */
   override def toSummaryString: String = {
+
     val numSamples = labeledPoints.count()
     val weightSum = labeledPoints.values.map(_.weight).sum()
     val responseSum = labeledPoints.values.map(_.label).sum()
     val featureStats = labeledPoints.values.map(_.features.activeSize).stats()
+
     s"numSamples: $numSamples\nweightSum: $weightSum\nresponseSum: $responseSum" +
         s"\nnumFeatures: $numFeatures\nfeatureStats: $featureStats"
   }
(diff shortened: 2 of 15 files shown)


ashelkovnykov:
Updated OP with results of testing performance on the cluster. No major performance changes discernible.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -104,28 +104,39 @@ object RandomEffectCoordinate {
...
           .persistRDD(StorageLevel.INFREQUENT_REUSE_RDD_STORAGE_LEVEL)
           .materialize()
 
         Some(randomEffectTracker)
-      } else None
+      } else {
+        None.asInstanceOf[Option[RandomEffectOptimizationTracker]]
+      }
+
+    updatedModelsAndTrackers.unpersist()
```
fastier-li: It looks like the 3 loops are tied together by `updatedModelsAndTrackers`. Since this is unpersisted, can we use the model and the tracker(s) right where we need them in the first loop, instead of having 3 loops, and persisting in between? Presumably that would be faster, and possibly simpler?

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -138,7 +149,6 @@ object RandomEffectCoordinate {
    *
    * @note The score is the dot product of the model coefficients with the feature values (in particular, does not go
    *       through non-linear link function in logistic regression!).
-   *
```
fastier-li: Previous white line was correct. Notes are not the same category as parameters, so logically they need to be separated.

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
```diff
@@ -148,9 +155,12 @@ protected[ml] class RandomEffectDataSet(
    * @return This object with [[activeData]], [[uniqueIdToRandomEffectIds]], and [[passiveDataOption]] materialized
    */
   override def materialize(): RandomEffectDataSet = {
-    activeData.count()
-    uniqueIdToRandomEffectIds.count()
-    passiveDataOption.foreach(_.count())
+
+    passiveDataOption match {
```
fastier-li: I would comment on what is going on here, and why we "smart materialize".

photon-api/src/main/scala/com/linkedin/photon/ml/data/FixedEffectDataSet.scala
```diff
@@ -99,7 +106,9 @@ protected[ml] class FixedEffectDataSet(
    * @return This object with [[labeledPoints]] materialized
    */
   override def materialize(): FixedEffectDataSet = {
-    labeledPoints.count()
+
+    smartMaterialize(labeledPoints)
```
fastier-li: How about `materializeOnce`? "smart" is not particularly descriptive.

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
```diff
@@ -89,57 +97,56 @@ protected[ml] class RandomEffectOptimizationTracker(val optimizationStatesTracke
...
-  def countConvergenceReasons: Map[String, Int] = {
-    optimizationStatesTrackers.map { optimizationStatesTracker =>
-      if (optimizationStatesTracker.convergenceReason.isDefined) {
-        (optimizationStatesTracker.convergenceReason.get.reason, 1)
-      } else {
-        (RandomEffectOptimizationTracker.NOT_CONVERGED, 1)
-      }
-    }.reduceByKey(_ + _).collectAsMap()
-  }
+  def countConvergenceReasons: Map[String, Int] = optimizationStatesTrackers.map { optimizationStatesTracker =>
```
fastier-li: Can we please add a carriage return between the signature and the body of the function for a multiline body? This makes it harder to read the return type of the function for me (esp. because the body doesn't even start with a `{`). And it is not consistent with `def getRegularizationTermValue(modelsRDD` below.

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
```diff
@@ -89,57 +97,56 @@ protected[ml] class RandomEffectOptimizationTracker(val optimizationStatesTracke
...
+      (optimizationStatesTracker
+        .convergenceReason
+        .map(_.reason)
+        .getOrElse(RandomEffectOptimizationTracker.NOT_CONVERGED), 1)
```
fastier-li: What are other possible statuses? Should we count e.g. only the successes instead, and warn in case of any status but success?

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
```diff
@@ -89,57 +97,56 @@ protected[ml] class RandomEffectOptimizationTracker(val optimizationStatesTracke
...
+  def getNumIterationStats: StatCounter = optimizationStatesTrackers.map(_.getTrackedStates.length).stats()
 
   /**
    * Get stats counter on the elapsed time tracked in all optimization states trackers
    *
    * @return A [[StatCounter]] of the elapsed time tracked in all optimization states trackers
    */
-  def getElapsedTimeStats: StatCounter = {
-    optimizationStatesTrackers.map { optimizationStatesTracker =>
+  def getElapsedTimeStats: StatCounter = optimizationStatesTrackers.map { optimizationStatesTracker =>
```
fastier-li: Same signature/body issue as above.

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
```diff
@@ -89,57 +97,56 @@ protected[ml] class RandomEffectOptimizationTracker(val optimizationStatesTracke
...
     val timeElapsedStats = getElapsedTimeStats
+
     RandomEffectOptimizationTracker.SUMMARY_FORMAT.format(convergenceReasons, numIterationsStats, timeElapsedStats)
   }
 }
 
 object RandomEffectOptimizationTracker{
-  protected[optimization] val NOT_CONVERGED = "Not converged"
 
+  protected[optimization] val NOT_CONVERGED = "Not converged"
```
fastier-li: We need to analyze the possible reasons of non-convergence to tackle them at the root as much as possible (just a reminder, let's keep a todo item about that somewhere).

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -402,7 +402,7 @@ object Driver {
         val logsDir = new Path(params.outputDir, LOGS).toString
         implicit val logger = new PhotonLogger(logsDir, sc)
         // TODO: This Photon log level should be made configurable
-        logger.setLogLevel(PhotonLogger.LogLevelInfo)
+        logger.setLogLevel(PhotonLogger.LogLevelDebug)
```
fastier-li: Let's keep it at `Info` by default, lest we generate too much log. We need to implement the TODO on the line just above rather than tweak to `Debug`, IMHO.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
ashelkovnykov: I don't understand the question.

We're trying to produce (up to) two new objects, each of which owns an `RDD`. One owns an `RDD` of models, the other an `RDD` of states. These two `RDD`s share a common ancestor in their lineage, because the models and states are computed at the same time. We compute the ancestor `RDD` (which is most quickly computed now, while IT'S ancestors are also persisted), temporarily persist it, force computation of the two derived `RDD`s (while the ancestor `RDD` exists), and then unpersist the ancestor (so that we do not need to track it and unpersist it later).

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
ashelkovnykov: I agree - Scala is doing this on its own, and I can't find an option in the style to make it stop. I think we should leave it be for now, until we can find a way to lock it in the style.

photon-api/src/main/scala/com/linkedin/photon/ml/data/FixedEffectDataSet.scala
ashelkovnykov: Good point - that's a better name

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
ashelkovnykov: With the `materializeOnce` change and the documentation, I think this is clear

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
ashelkovnykov: Spoke offline - will do

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
ashelkovnykov: Noted

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
ashelkovnykov: See the `ConvergenceReason` trait (enumeration) for the other reasons. I think that your second question is related to #267.

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/RandomEffectOptimizationTracker.scala
ashelkovnykov: #267

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
ashelkovnykov: I disagree - I'm not sure exactly which commit changed the logging level to "Info" but it has been "Debug" for a while before that, so I think we should keep it at "Debug" until the option becomes configurable.

My reasoning is that the log isn't too much larger in "Debug" mode, there have been no complaints, and it keeping it at "Info" would publish less info than users may be used to.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
fastier-li: Let me look at this again later.

fastier-li: I was wondering about performance, but if this is not a problem, then I'm OK.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
fastier-li: If debug level doesn't incur too much penalty when looking for root cause of issues in prod by mining the logs, I'm OK with it. My concern is to not generate a log that is so big that it: degrades runtime performance ; crashes some space quota on the grid ; makes it hard to figure out issues because you have to grep through multiple TBs of log files on remote machines.

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
fastier-li: OK.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
fastier-li: OK.

PR #270, opened 2017-04-20
Testing photon-ml build against Spark 2.1.0/Scala 2.11.8.

fastier-li:
It works all the way to the integration tests. 

However, beyond making it compile and run, there are simplifications/updates we could/should do in the code, such as in `AvroDataReader` where Spark SQL offers better APIs in 2.1.0. I have not attempted to do those, I have only fixed build issues and verified the integration tests.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
@@ -18,7 +18,7 @@ import scala.util.{Failure, Success, Try}
 
 import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkContext
-import org.apache.spark.sql.DataFrame
+import org.apache.spark.sql.{DataFrame, SparkSession}
 import org.slf4j.Logger
 
 import com.linkedin.photon.ml.Types.{FeatureShardId, SparkVector}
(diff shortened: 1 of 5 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -186,9 +186,12 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
...
       if (params.checkData) {
-        val numBad = dataframe
-          .map { row => if (row.getAs[Double](weightColumnName) <= 0.0) 1 else 0 }
-          .reduce(_ + _)
+        val numBad = dataframe.filter(s"$weightColumnName <= 0.0").count()
+        // This doesn't work with Spark 2.1.0
+//        val numBad = dataframe
+//          .map { row =>
+//            if (row.getAs[Double](weightColumnName) <= 0.0) 1 else 0 }
+//          .reduce(_ + _)
```
joshvfleming: It does work, but you have to explicitly call `rdd` on the dataframe, e.g.:

```scala
  val numBad = dataframe
    .rdd
    .map { row =>
      if (row.getAs[Double](weightColumnName) <= 0.0) 1 else 0 }
    .reduce(_ + _)
```

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/Driver.scala
```diff
@@ -186,9 +186,12 @@ final class Driver(val sc: SparkContext, val params: GameParams, implicit val lo
 
     data.map { dataframe =>
       if (params.checkData) {
-        val numBad = dataframe
-          .map { row => if (row.getAs[Double](weightColumnName) <= 0.0) 1 else 0 }
-          .reduce(_ + _)
+        val numBad = dataframe.filter(s"$weightColumnName <= 0.0").count()
```
joshvfleming: Consider using expressing the predicate in Scala rather than sticking it all in a String:

```scala
val numBad = dataframe.filter(col("weightColumnName") <= 0.0).count()
```

fastier-li:
#275 supersedes this one.

PR #272, opened 2017-04-27
Comments, names and various small fixes or improvements.

fastier-li:
Small things that should be all good.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
@@ -22,6 +22,7 @@ import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.{StorageLevel => SparkStorageLevel}
 import org.apache.spark.{Partitioner, SparkContext}
 
+import com.linkedin.photon.ml.Types.{FeatureShardId, REId, REType, UniqueSampleId}
 import com.linkedin.photon.ml.constants.StorageLevel
 import com.linkedin.photon.ml.data.scoring.CoordinateDataScores
 import com.linkedin.photon.ml.spark.{BroadcastLike, RDDLike}
@@ -44,12 +45,12 @@ import com.linkedin.photon.ml.spark.{BroadcastLike, RDDLike}
  * @param featureShardId The feature shard ID
  */
 protected[ml] class RandomEffectDataSet(
-    val activeData: RDD[(String, LocalDataSet)],
-    protected[data] val uniqueIdToRandomEffectIds: RDD[(Long, String)],
-    val passiveDataOption: Option[RDD[(Long, (String, LabeledPoint))]],
+    val activeData: RDD[(REId, LocalDataSet)],
+    protected[data] val uniqueIdToRandomEffectIds: RDD[(UniqueSampleId, REId)],
+    val passiveDataOption: Option[RDD[(UniqueSampleId, (REId, LabeledPoint))]],
     val passiveDataRandomEffectIdsOption: Option[Broadcast[Set[String]]],
-    val randomEffectType: String,
-    val featureShardId: String)
+    val randomEffectType: REType,
+    val featureShardId: FeatureShardId)
   extends DataSet[RandomEffectDataSet]
   with RDDLike
   with BroadcastLike {
@@ -184,8 +185,9 @@ protected[ml] class RandomEffectDataSet(
    * @return A new updated dataset
    */
   def update(
-      updatedActiveData: RDD[(String, LocalDataSet)],
-      updatedPassiveDataOption: Option[RDD[(Long, (String, LabeledPoint))]]): RandomEffectDataSet =
+      updatedActiveData: RDD[(REId, LocalDataSet)],
+      updatedPassiveDataOption: Option[RDD[(UniqueSampleId, (REId, LabeledPoint))]]): RandomEffectDataSet =
+
     new RandomEffectDataSet(
       updatedActiveData,
       uniqueIdToRandomEffectIds,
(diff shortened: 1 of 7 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
```diff
@@ -235,10 +237,10 @@ object RandomEffectDataSet {
...
    * @return A new random effect dataset with the given configuration
    */
-  protected[ml] def buildWithConfiguration(
-      gameDataSet: RDD[(Long, GameDatum)],
-      randomEffectDataConfiguration: RandomEffectDataConfiguration,
-      randomEffectPartitioner: Partitioner): RandomEffectDataSet = {
+  protected[ml] def apply(
+    gameDataSet: RDD[(UniqueSampleId, GameDatum)],
+    randomEffectDataConfiguration: RandomEffectDataConfiguration,
+    randomEffectPartitioner: Partitioner): RandomEffectDataSet = {
```
ashelkovnykov: Looks like IDEA messed with your indentation all throughout this file

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataReader.scala
```diff
@@ -48,6 +48,9 @@ import com.linkedin.photon.ml.util._
  *       count explicitly to the "read" function to get around this problem. Once we revisit Photon's sensitivity to
  *       choice of partition count, we can remove this parameter.
  *
+ * @note About "feature bags" and "feature shards":
+ *       -
+ *
```
ashelkovnykov: Unfinished

photon-lib/src/main/scala/com/linkedin/photon/ml/Types.scala
```diff
@@ -24,6 +24,16 @@ object Types {
...
+  type UniqueSampleId = Long
+
+  // A "coordinate" is made up of several random effects
   type CoordinateId = String
+
+  // RE stands for "random effect"
+  type REType = String
+  type REId = String
+
+  // A "feature shard" is a "feature bag", i.e. an arbitrary set of features
```
ashelkovnykov: A feature shard is one or more feature bags

photon-lib/src/main/scala/com/linkedin/photon/ml/Types.scala
```diff
@@ -24,6 +24,16 @@ object Types {
   type SSV = org.apache.spark.mllib.linalg.SparseVector
   type SparkVector = org.apache.spark.mllib.linalg.Vector
 
+  // A "sample" is a training, validation or scoring vector of feature values
+  type UniqueSampleId = Long
+
+  // A "coordinate" is made up of several random effects
```
ashelkovnykov: A coordinate is a fixed effect, random effect, or matrix factorization model - a random effect coordinate is made up of several fixed effect models, one per unique random effect ID

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Timed.scala
```diff
@@ -70,8 +70,24 @@ object Timed {
...
+   * @return A Double with only n significant figures
+   */
+  def roundToSignificantFigures(num: Double, n: Int = 2): Double = {
+    if (num == 0) return 0
+    val d = Math.ceil(Math.log10(if (num < 0) -num else num))
+    val power = n - d.toInt
+    val magnitude = Math.pow(10, power)
+    val shifted = num * magnitude.round
+    shifted.toInt / magnitude
+  }
```
ashelkovnykov: This returns incorrect results whenever `num > pow(10, n)`. Isn't what we really want closer to the below?
```
def roundToSignificantDecimals(num: Double, n: Int = 3): Double = {
  require(n >= 0, s"Cannot have fewer than 0 significant decimal digits: $n")

  if (num = 0) {
    0
  } else {
    val base = num.floor
    val decimal = num - base
    val power = Math.pow(10, n)
    val rounded = (decimal * power).round / power

    base + rounded
  }
}
```

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataSet.scala
fastier-li: Fixed.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataReader.scala
fastier-li: Later.

photon-lib/src/main/scala/com/linkedin/photon/ml/Types.scala
fastier-li: It this ok:
`A CoordinateId is a fixed effect type, random effect type, or matrix factorization model type (e.g. "per-user", "per-song" ...)`

photon-lib/src/main/scala/com/linkedin/photon/ml/Types.scala
fastier-li: Fixed.

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Timed.scala
fastier-li: I don't think so. In the case of 0.0031415926535, I want to see 0.0031, not 0.0, and the code you suggest returns 0.0. We can combine with scientific notation, so that we see 3.1e-3.

photon-lib/src/main/scala/com/linkedin/photon/ml/Types.scala
ashelkovnykov: I think that's fine, though the I would cut the `"per-user", "per-song", ...` part since we would need to explain that refers to the random effect coodinates, as well as the factored random effect coordinates, which use but aren't quite a regular matrix factorization coordinate, and so on

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Timed.scala
ashelkovnykov: You can combine it with scientific notation - what I meant though was that what we actually care about in this case is rounding to the nearest millisecond for our timer when using seconds. If we did it with significant digits, we'll end up with timings such as `3.14e3 SECONDS to train model` which is counter to the original inspiration of this change: make the log timings more readable.

fastier-li: It all depends what you want to see. I want to see significant figures, but not too many. I don't want to round to 0 when the number is too small.

PR #273, opened 2017-04-27
Adding parameter to turn off warm start

fastier-li:
Probably some comments missing, but logic can start being reviewed. This deserves real close attention.

### DIFF
--- a/photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
+++ b/photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
@@ -38,19 +38,21 @@ import com.linkedin.photon.ml.optimization.{OptimizerType, RegularizationType}
 import com.linkedin.photon.ml.supervised.classification.LogisticRegressionModel
 import com.linkedin.photon.ml.supervised.model.GeneralizedLinearModel
 import com.linkedin.photon.ml.test.{CommonTestUtils, SparkTestUtils, TestTemplateWithTmpDir}
-import com.linkedin.photon.ml.util.Utils
+import com.linkedin.photon.ml.util.{VectorUtils, Utils}
 
 /**
-  * This class tests Driver with a set of important configuration parameters
-  */
+ * This class tests the Photon (not GAME) Driver with a set of important configuration parameters
+ */
 class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
 
   import DriverTest._
 
   @Test
   def testRunWithMinimalArguments(): Unit = sparkTest("testRunWithMinimalArguments") {
+
     val outputDir = getTmpDir + "/testRunWithMinimalArguments"
     val args = mutable.ArrayBuffer[String]()
+
     appendCommonJobArgs(args, outputDir)
     args += CommonTestUtils.fromOptionNameToArg(OPTIMIZER_TYPE_OPTION)
     args += "TRON"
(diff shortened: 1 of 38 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/Optimizer.scala
```diff
@@ -239,7 +246,7 @@ abstract class Optimizer[-Function <: ObjectiveFunction](
       (data: objectiveFunction.Data): (Vector[Double], Double) = {
 
     clearOptimizerInnerState()
-    clearOptimizationStatesTracker()
+    statesTracker = initStateTracker()
```
ashelkovnykov: Let's add a comment just in case that this is resetting the stateTracker

photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/Optimizer.scala
```diff
@@ -137,110 +91,144 @@ abstract class Optimizer[-Function <: ObjectiveFunction](
...
    * @param data The training data
-   * @return The updated state of the optimizer
+   * @return Optimized model coefficients and corresponding objective function's value
    */
-  protected def runOneIteration
-    (objectiveFunction: Function, currState: OptimizerState)
-    (data: objectiveFunction.Data): OptimizerState
+  protected[ml] def optimize
+    (objectiveFunction: Function, initialCoefficients: Vector[Double])
+    (data: objectiveFunction.Data): (Vector[Double], Double) = {
```
ashelkovnykov: Why force passing in a zero vector explicitly?

photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
```diff
@@ -931,25 +963,131 @@ class DriverTest extends SparkTestUtils with TestTemplateWithTmpDir {
...
+    // Check that the models have identical metrics with similar values.
+    // Metrics should be equal in all cases, even if there are correlated features in the training data.
+    assertEquals(coldMetrics.size, warmMetrics.size)
+    coldMetrics.foreach { case (lambda, coldMetricsMap) =>
+      assertTrue(warmMetrics.contains(lambda))
+      warmMetrics(lambda).foreach { case (metricName, warmMetricValue) =>
+        assertTrue(coldMetricsMap.contains(metricName))
+        val coldMetricValue = coldMetricsMap(metricName)
+        val diff = (coldMetricValue - warmMetricValue) / math.min(coldMetricValue, warmMetricValue)
+        assertEquals(math.abs(diff), 0D, 1e-3)
```
ashelkovnykov: It'd be better to store Epsilon in some named var

photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/Optimizer.scala
fastier-li: Because it simplifies the Optimizer API.

photon-client/src/integTest/scala/com/linkedin/photon/ml/DriverTest.scala
fastier-li: Not a var visible to the whole DriverTest: this epsilon is used only here. It had the name "WARM_START_EPSILON", which makes it clear it's specific to this case. Having the value here, since it is used only once and here, also has the merit of being immediately visible. Why introduce a variable that will need to be followed to see the value (for a value used ONCE and only HERE).

PR #280, opened 2017-05-26
Minor MathConst and GLMSuite changes

ashelkovnykov:
- Replace "high", "medium", and "low" tolerances with one "epsilon" for representing values close to 0
- Replace unit/integration test dependencies on 3 tolerance values in `MathConst` on 2 tolerance values in `CommonTestUtils`
- Move the old "low" tolerance into `VectorUtils` as a sparsity threshold (**As discussed in the comments below and offline, this should become a configurable parameter with a default value determined through testing**)
- Replace `String` and `Long` with `REId`, `REType`, and `UniqueSampleId` helper types where possible
- Copy `AvroDataReader` duplicate checking code to the `GLMSuite` until we're ready to deprecate the original Photon driver
- Add a unit test for the above `GLMSuite` change

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/FactoredRandomEffectCoordinate.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/FactoredRandomEffectCoordinate.scala
@@ -17,7 +17,7 @@ package com.linkedin.photon.ml.algorithm
 import breeze.linalg.Matrix
 import org.apache.spark.rdd.RDD
 
-import com.linkedin.photon.ml.constants.{MathConst, StorageLevel}
+import com.linkedin.photon.ml.constants.StorageLevel
 import com.linkedin.photon.ml.data._
 import com.linkedin.photon.ml.data.scoring.CoordinateDataScores
 import com.linkedin.photon.ml.function.{DistributedObjectiveFunction, SingleNodeObjectiveFunction}
@@ -149,9 +149,11 @@ protected[ml] class FactoredRandomEffectCoordinate[
         val updatedFactoredRandomEffectModel = factoredRandomEffectModel.updateFactoredRandomEffectModel(
           updatedModelsRDD,
           updatedProjectionMatrixBroadcast)
-        val factoredRandomEffectOptimizationTracker = if (isTrackingState) {
+        val factoredRandomEffectOptimizationTracker: Option[OptimizationTracker] = if (isTrackingState) {
           Some(new FactoredRandomEffectOptimizationTracker(factoredRandomEffectOptimizationTrackerArray))
-        } else None
+        } else {
+          None
+        }
 
         (updatedFactoredRandomEffectModel, factoredRandomEffectOptimizationTracker)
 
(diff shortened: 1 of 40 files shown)


fastier-li:
Large PRs are hard to review.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/FactoredRandomEffectCoordinate.scala
```diff
@@ -260,22 +259,18 @@ object FactoredRandomEffectCoordinate {
...
-      sparsityToleranceThreshold: Double = 0.0): RDD[(Long, LabeledPoint)] = {
+      modelsRDD: RDD[(String, GeneralizedLinearModel)]): RDD[(Long, LabeledPoint)] = {
 
     localDataSetRDD.join(modelsRDD).flatMap { case (_, (localDataSet, model)) =>
       localDataSet.dataPoints.map { case (uniqueId, labeledPoint) =>
         val generatedFeatures = VectorUtils.kroneckerProduct(
           labeledPoint.features,
           model.coefficients.means,
-          sparsityToleranceThreshold)
+          MathConst.SPARSITY_THRESHOLD)
```
fastier-li: Do we have a rationale to have a separate sparsity threshold and epsilon? Why not use epsilon for the sparsity threshold?

photon-lib/src/main/scala/com/linkedin/photon/ml/util/ClassUtils.scala
```diff
@@ -22,15 +22,15 @@ object ClassUtils {
   private val ANON_CLASS_MARKER = "$anon$"
 
   /**
-   * Method for detecting anonymous classes (the isAnonymousClass method is broken in Scala).
+   * Method for detecting anonymous classes (the isAnonymousClass method is broken in Scala as late as 2.11).
    *
    * @param clazz A class
    * @return True if the given class is anonymous, false otherwise.
    */
   def isAnonClass(clazz: Class[_]): Boolean = clazz.getName.contains(ANON_CLASS_MARKER)
```
joshvfleming: That we have this in production code suggests possible evil going on somewhere. Looks like it's used mainly when we infer GAME model types.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/FactoredRandomEffectCoordinate.scala
joshvfleming: They are logically distinct things, but it could be that `EPSILON` would be a good rule of thumb for a default `SPARSITY_THRESHOLD`. I think we should try it and see what the effect on model size is in the real world.

photon-lib/src/main/scala/com/linkedin/photon/ml/util/ClassUtils.scala
ashelkovnykov: It's only used as you describe above - I found that I needed to add the code to make up for the broken `isAnonymousClass` method. The reason I needed that method is due to the way modelTypes are copied currently (so that they're not recomputed over and over again).a

Coincidentally, I spent half a day trying to get `modelType` copied without anonymous classes and it's just a mess when dealing with the `RandomEffectModel` thanks to how Scala handles constructors. I've given up.

ashelkovnykov:
Added changes based on feedback. The sparsity stuff is a placeholder - I don't have time to implement the solution described above. If you're good with the changes, let me know first so that I can squash the commits.

joshvfleming:
LGTM. I'll merge after you squash.

PR #281, opened 2017-06-05
Need to remove protected[ml] on GameTrainingParams.parseFromCommandLine

fastier-li:


photon-api/src/main/scala/com/linkedin/photon/ml/optimization/game/GameModelOptimizationConfiguration.scala
```diff
@@ -24,13 +24,13 @@ package com.linkedin.photon.ml.optimization.game
  */
 case class GameModelOptimizationConfiguration(
     fixedEffectOptimizationConfiguration: Map[String, GLMOptimizationConfiguration],
```
ashelkovnykov: May as well add a default here too

PR #288, opened 2017-07-13
Fix iteration number in validation loop.

fastier-li:


ashelkovnykov:
This isn't quite the solution that's required, I've got a PR coming out for this in a second. I'll copy the more descriptive log message from this PR into mine.

PR #292, opened 2017-07-20
Fix various Photon-ML errors

ashelkovnykov:
This PR addresses several issues discovered while readying an internal tutorial for new-hires.
- Integration tests were accidentally disabled during the upgrade to Gradle 4
- During cold start training, the first state would never be logged. This would later lead to a mismatch during per-iteration metrics logging.
- During warm start training, the starting coefficients were not being converted back into the transformed space for normalization. In some instances, this would cause the training to essentially fail, as the starting coefficients would be so poor that `LBFGS` would fail to find a valid step size within the default parameters and give up on training.
- Normalization integration tests were split between two separate files.
- An extra copy of `OptimizerState` was created during the module refactor.


joshvfleming:
LGTM. Will merge once the tests pass.

PR #293, opened 2017-07-21
Refactor DataValidators to provide information about what causes failures

ashelkovnykov:
These changes provide better log messages, but change the ordering of operations, so now data validation requires multiple passes.

I still need to test how this affects performance.


photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -27,151 +26,129 @@ import com.linkedin.photon.ml.{DataValidationType, TaskType}
...
-      })).iterator
-    }).fold(true)(_ && _)
-  }
+  private def validateData(
+      dataSet: RDD[LabeledPoint],
+      perSampleValidators: List[((LabeledPoint => Boolean), String)]): Seq[String] =
+    perSampleValidators
+      .map { case (validator, msg) =>
+        val validatorBroadcast = dataSet.sparkContext.broadcast(validator)
+        val result = dataSet.aggregate(true)(
```
namitk: We have deliberated on and seen performance tradeoffs betwen aggregate vs treeAggregate and the depth of it but that was for the gradient aggregation which is called repeatedly during training. I think it should be fine here.

photon-client/src/integTest/scala/com/linkedin/photon/ml/data/DataValidatorsTest.scala
```diff
@@ -51,94 +53,188 @@ class DataValidatorsTest extends SparkTestUtils {
...
+        DataValidationType.VALIDATE_FULL, false),
+      Array(
+        sc.parallelize(List(lpPositiveLabel, lpInfOffset)),
+        TaskType.POISSON_REGRESSION,
+        DataValidationType.VALIDATE_FULL, false)
+    )
   }
 
-  /*
   // TODO This test does not work because the spark context ends up being null in the data provider. Unlike other tests
```
namitk: Since you have fixed this test, we can get rid of this comment.

ashelkovnykov: Yes - I wanted to keep the comment to explain why the tests call the data provider differently than they're supposed to, but forgot that it references a specific test.

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
ashelkovnykov: I can change it to use `treeAggregate` - the only reason I didn't is that right now we're using a default value for `treeAggregate` (1, in most cases). I'll add it to the performance checks I wanted to do - old way vs. new way vs. new way + `treeAggregate`.

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -27,151 +26,129 @@ import com.linkedin.photon.ml.{DataValidationType, TaskType}
...
+          seqOp = (result, dataPoint) => result && validatorBroadcast.value(dataPoint),
+          combOp = (result1, result2) => result1 && result2)
+
+        (result, msg)
+      }
+      .filter(!_._1)
```
ashelkovnykov: Reminder for myself that this can be replaced by `filterNot`

PR #296, opened 2017-08-23
Enable identity projection

joshvfleming:
Previously the code assumed that we're doing a factored random effect problem for identity projections. This would cause a failure when it tried to load from a nonexistent factored random effect configuration. This commit adds a dedicated identity projector and selects it when the IDENTITY projection option is selected.


ashelkovnykov:
Could you also fix the issue in the `IndexMapProjector`, the one where it uses the `activeIterator?`

joshvfleming:
@ashelkovnykov I want to do that separately.

ashelkovnykov:
@joshvfleming A perfectly cromulent idea

PR #300, opened 2017-09-01
Added data validation to GAME code path

Adama94:
Need to rename the validation functions and pull the UID out of the integration tests.


photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -151,4 +244,95 @@ object DataValidators extends Logging {
...
+                  case _ =>
+                }
+              }
+              case _ =>
+            }
+            Set((result, msg))
+          }
+        }
+          .collect {
+          case (result: Boolean, msg: String) => (result, msg)
```
ashelkovnykov: Confusing whitespace here. Also, try to keep the function call and the arguments for the lambda on the same line: `.collect { case (result: Boolean, msg: String) =>`

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -151,4 +244,95 @@ object DataValidators extends Logging {
...
+              case Success(_) => {
+                val dataSetContainsColumn = dataSet.columns.contains(transformedColumnName.get)
+                dataSetContainsColumn match {
+                  case true => validator(r, transformedColumnName.get)
+                  case _ =>
+                }
+              }
+              case _ =>
+            }
+            Set((result, msg))
```
ashelkovnykov: As above: all of these column names must be present, since a) they're the names the user is telling us to use and b) we should be loading the `DataFrame` correctly using them.

`apply` is a [special function in Scala](https://stackoverflow.com/questions/9737352/what-is-the-apply-function-in-scala) which can be invoked using only the braces: `inputColumnsNames(columnName)`.

I think lines 277 - 288 should be replaced by:
```
(validator(r, inputColumnsNames(columnName)), msg)
```

fastier-li:
Nothing to add to Alex's comments. Fix then merge.

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
Adama94: There are cases where "weight" and/or "offset" (and potentially "response" if we are scoring only) aren't required, so I left part of this in to check for those.

PR #304, opened 2017-09-20
Performance improvements for GAME scoring

joshvfleming:


photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
```diff
@@ -260,6 +267,12 @@ object Params {
             }
         )
 
+      opt[Boolean]("spill-scores")
+        .text(s"Whether to spill persisted scores to disk. This is more CPU intensive, but prevents recomputation " +
+          "of scores when memory-persisted blocks are evicted. Useful far very large scoring tasks. " +
```
fastier-li: "for" instead of "far"?

fastier-li:
Can you try on a couple of larger datasets on the grid? I'm not sure the change would be adequately tested by the unit or even integration tests we have.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/scoring/Params.scala
joshvfleming: Yep.

joshvfleming:
@fastier-li I'm going to leave this up to the folks requesting the change, since I don't really have a case that could adequately reproduce / test the problem.

Issue #306, opened 2017-09-22
The validateDataFrame function pulls entire dataset into Driver


joshvfleming:
The following code attempts to pull the entire dataset into the Driver before running validations:

https://github.com/linkedin/photon-ml/blob/237c315d9b78945c5c0d5dbafae31ebcfeef58a1/photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala#L265

We should instead map over the DataFrame rows so that the validations run in a distributed fasion, and then `collect` the results.

li-ashelkov:
Resolve by #307

PR #307, opened 2017-09-25
Fix issue #306 related to data validation code

Adama94:
The data validation code tried to pull the entire dataset into memory, as described in issue #306.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
@@ -260,25 +260,25 @@ object DataValidators extends Logging {
       inputColumnsNames: InputColumnsNames,
       featureSectionKeys: Set[FeatureShardId]): Seq[String] = {
 
-    perSampleValidators
-      .flatMap { case (validator, columnName, msg) =>
-        val rows = dataSet.collect()
-        rows.flatMap { r =>
-          if (columnName.equals(InputColumnsNames.FEATURES_DEFAULT)) {
-            featureSectionKeys.map(features => (validator(r, features), msg))
+    val columns = dataSet.columns
+    dataSet.rdd.flatMap { r =>
+      perSampleValidators.flatMap { case (validator, columnName, msg) =>
+        if (columnName == InputColumnsNames.FEATURES_DEFAULT) {
+          featureSectionKeys.map(features => (validator(r, features), msg))
+        } else {
+          val result = if (columns.contains(inputColumnsNames(columnName))) {
+            (validator(r, inputColumnsNames(columnName)), msg)
           } else {
-            val result = if (dataSet.columns.contains(inputColumnsNames(columnName))) {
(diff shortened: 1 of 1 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -259,26 +259,26 @@ object DataValidators extends Logging {
...
           }
+
+          Seq(result)
         }
       }
-      .filterNot(_._1)
-      .map(_._2)
+    }
+    .collect()
+    .filterNot(_._1)
```
ashelkovnykov: I'd like @joshvfleming to confirm, but I believe filtering first and then collecting would be more efficient.

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -259,26 +259,26 @@ object DataValidators extends Logging {
...
-      .flatMap { case (validator, columnName, msg) =>
-        val rows = dataSet.collect()
-        rows.flatMap { r =>
-          if (columnName.equals(InputColumnsNames.FEATURES_DEFAULT)) {
-            featureSectionKeys.map(features => (validator(r, features), msg))
+    
+    val columns = dataSet.columns
+    dataSet.rdd.flatMap { r =>
+      perSampleValidators.flatMap { case (validator, columnName, msg) =>
+        if (columnName.equals(InputColumnsNames.FEATURES_DEFAULT)) {
```
joshvfleming: Using `==` for comparing `String`s is safe in Scala, and preferred.

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -259,26 +259,26 @@ object DataValidators extends Logging {
...
+          featureSectionKeys.map(features => (validator(r, features), msg))
```
joshvfleming: It looks like if the validator is specified for the default feature column name, we run validators for all feature columns in one pass. What's the reason for this?

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
joshvfleming: @ashelkovnykov is right -- as it stands, what this will do is collect validation results for all data points in the Driver and then filter them locally. If we filter first, we pull only the validation results we're interested in to the Driver.

photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
Adama94: The if/else block here separates the case where we validate the "features" column from other columns. Other columns have only one field to validate (e.g. "response"), while for the "features" column we'd like to validate for all the feature shard ids.

PR #308, opened 2017-10-03
Fix bug causing log-likelihoods to be computed incorrectly.

ashelkovnykov:
- Logistic and Poisson log-likelhoods computed incorrectly
- Average computed incorrectly, causing stochastic behaviour
- The above resulted in a mismatch between the metrics reported for Poisson regression, and the ones used to choose the best model (both were wrong)

### DIFF
--- a/photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
+++ b/photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
@@ -26,8 +26,8 @@ import com.linkedin.photon.ml.supervised.regression.{PoissonRegressionModel, Reg
 import com.linkedin.photon.ml.util.Logging
 
 /**
-  * A collection of evaluation metrics and functions
-  */
+ * A collection of evaluation metrics and functions
+ */
 object Evaluation extends Logging {
 
   val MEAN_ABSOLUTE_ERROR = "Mean absolute error"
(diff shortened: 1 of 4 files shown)


photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
```diff
@@ -129,49 +130,46 @@ object Evaluation extends Logging {
...
+  private def averageLogLikelihoodRDD(logLikelihoods: RDD[(Double, Int)]): Double = {
+
+    val (logLikelihood, count) = logLikelihoods.reduce { (a, b) =>
+      val (logLikelihoodA, countA) = a
+      val (logLikelihoodB, countB) = b
+
+      (logLikelihoodA + logLikelihoodB, countA + countB)
+    }
+
+    logLikelihood - math.log(count)
```
ankans: Where is this formula coming from?

photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
```diff
@@ -129,49 +130,46 @@ object Evaluation extends Logging {
...
-        count += 1
-        mean += (x - mean) / count
-      })
-      Seq((count, mean)).iterator
-    }).reduce((a, b) => {
-      // Aggregate per-partition means
-      val (countA, meanA) = a
-      val (countB, meanB) = b
-      val newCount = countA + countB
-      val newMean = meanA + (meanB * countB) / newCount
```
ankans: Shouldn't this have been meanA*countA +.... ?

ashelkovnykov: Yeah

photon-diagnostics/src/main/scala/com/linkedin/photon/ml/Evaluation.scala
ashelkovnykov: From a misunderstanding I had - changing it

PR #314, opened 2017-11-14
Enable option to use warm-start training for GAME

ashelkovnykov:
- Introduced configurable parameter to re-use trained GameModel objects as a starting point for training further GameModels
- Note: GAME already uses warm-start training during coordinate descent (the way the legacy Photon code does)
- Added unit test for GameEstimator
- Fixed bug in GameEstimator.copy

### DIFF
--- a/photon-api/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorIntegTest.scala
+++ b/photon-api/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorIntegTest.scala
@@ -45,9 +45,9 @@ import com.linkedin.photon.ml.util._
 /**
  * Integration tests for [[GameEstimator]].
  */
-class GameEstimatorTest extends SparkTestUtils with GameTestUtils {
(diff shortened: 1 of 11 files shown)


namitk:
Quick question: Where is the model read from? If `useWarmStart` is set, shouldn't we be reading another parameter which is the path to the initial model? Is that going to be added later or am I missing something?

ashelkovnykov:
@namitk 

Hey Namit - take a look at the changes in the `GameEstimator` line 280: we start out without a model to use as the base. However, on line 290 we update the base model with the first trained result so that future iterations of training (for hyperparameter grid search) will reuse the previously computed model.

One thing we're going to add is partial retraining, and in that case we will actually load a model as you describe first and if `useWarmStart` is set use it as the base.

Another thing which we should ideally address is the current hyperparameter tuning flaw that it uses a new GameEstimator call each time, which means that we not only don't use warm-start whether it's set or not, but also reprocess the `DataFrame` each time. @joshvfleming and I are still working on an elegant solution to this.

PR #315, opened 2017-11-14
Modify DataValidators to avoid ClassCastExceptions.

ashelkovnykov:


photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -184,10 +187,11 @@ object DataValidators extends Logging {
...
    */
-  def rowHasFiniteWeight(row: Row, inputColumnName: String): Boolean = {
-    val weight = row.getAs[Double](inputColumnName)
-    !(weight.isNaN || weight.isInfinite)
-  }
+  def rowHasFiniteWeight(row: Row, inputColumnName: String): Boolean =
+    row.getAs[Any](inputColumnName) match {
+      case weight: Double => !(weight.isNaN || weight.isInfinite)
+      case _ => false
+    }
```
joshvfleming: The functions `rowHasFiniteLabel`, `rowHasFiniteOffset`, and `rowHasFiniteWeight` are really the same function. You can reduce it down to one function called `rowHasFiniteValue` or something like that.

PR #316, opened 2017-11-14
Photon ML API Improvements

ashelkovnykov:
- Added defaults values for some CoordinateDataConfiguration parameters
- Removed usage of 'monotonically_increasing_id()'; it's reported to cause trouble due to its instability (it often gets moved to the very end of operations by Spark and thus is inconsistent) and the way we were using it made it unn
- Added setter convenience functions to GameEstimator
- Removed 'featureShardColumnNames' parameter from GameEstimator
- Fixed incorrect calls in GameEstimator to 'getOrDefault' instead of 'getRequiredParam'
- Added a new GameTransformer class, a kind of combination of the GameEstimator and org.apache.spark.ml.Model
- Made the 'featureBagsDirectory' parameter optional for training/scoring (since the AvroDataReader will build a index map on its own by default)
- Replaced Timer with Timed in GameScoringDriver
- Removed Timer class


photon-api/src/main/scala/com/linkedin/photon/ml/transformers/GameTransformer.scala
```diff
@@ -0,0 +1,308 @@
...
+  def transform(data: DataFrame): ModelDataScores = {
+
+    validateParams()
+
+    val randomEffectTypes = getRequiredParam(model)
+      .toMap
+      .values
+      .flatMap {
+        case rem: RandomEffectModel => Seq(rem.randomEffectType)
+        case _ => Seq()
```
joshvfleming: I think using `Option` would be more natural here, since we have either one element or none:

```scala
  .flatMap {
    case rem: RandomEffectModel => Some(rem.randomEffectType)
    case _ => None
  }
```

photon-api/src/main/scala/com/linkedin/photon/ml/transformers/GameTransformer.scala
```diff
@@ -0,0 +1,308 @@
...
+    }
+
+    if (getOrDefault(logDataAndModelStats)) {
+      logGameDataSet(gameDataSet, randomEffectTypes)
+      logger.debug(s"GAME model summary:\n${getRequiredParam(model).toSummaryString}")
+    }
+
+    val scores = Timed("Computing scores") {
+      scoreGameDataSet(gameDataSet)
+    }
```
joshvfleming: Do the scores really get materialized here? If not, this `Timed` block will give a misleading number.

photon-api/src/main/scala/com/linkedin/photon/ml/transformers/GameTransformer.scala
```diff
@@ -0,0 +1,308 @@
...
+
+    Timed("Evaluating scores") {
+      get(validationEvaluators).foreach(
+        _.foreach { evaluatorType =>
+          val evaluationMetricValue = evaluateScores(evaluatorType, gameDataSet, scores)
+          logger.info(s"Evaluation metric value on scores with $evaluatorType: $evaluationMetricValue")
+        })
+    }
+
+    // TODO: Instead, we should merge the scores back into the DataFrame in a new column (at least optionally)
```
joshvfleming: Yeah. This would also be more in the spirit of the `Transformer` pattern in spark.ml.

ashelkovnykov: Yes, I gave it a quick go but it became a bit of an ordeal so I decided to leave it for a later commit.

photon-api/src/main/scala/com/linkedin/photon/ml/transformers/GameTransformer.scala
ashelkovnykov: They do indeed - line 278.

PR #327, opened 2017-12-02
Define equality between Evaluators and fix related bug

ashelkovnykov:
See: http://www.artima.com/lejava/articles/equality.html

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/AreaUnderROCCurveEvaluator.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/AreaUnderROCCurveEvaluator.scala
@@ -17,13 +17,15 @@ package com.linkedin.photon.ml.evaluation
 import org.apache.spark.mllib.evaluation.BinaryClassificationMetrics
 import org.apache.spark.rdd.RDD
 
+import com.linkedin.photon.ml.Types.UniqueSampleId
+
 /**
  * Evaluator that computes area under the ROC curve.
  *
  * @param labelAndOffsetAndWeights A [[RDD]] of (id, (label, offset, weight)) tuples
  */
 protected[ml] class AreaUnderROCCurveEvaluator(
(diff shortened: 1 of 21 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/AreaUnderROCCurveEvaluator.scala
```diff
@@ -49,4 +52,25 @@ protected[ml] class AreaUnderROCCurveEvaluator(
...
+
+  /**
+   * Compares two [[AreaUnderROCCurveEvaluator]] objects.
+   *
+   * @param other Some other object
+   * @return True if the both models conform to the equality contract and have the same model coefficients, false
+   *         otherwise
+   */
+  override def equals(other: Any): Boolean = other match {
+    case that: AreaUnderROCCurveEvaluator => that.canEqual(this) && super.equals(that)
```
joshvfleming: This pattern makes sense in the Java world, but IMO it's unidiomatic here in Scala. Matching on the evaluator type already guarantees that we won't call `super.equals` unless the other object is the type we expect. Also, from a performance perspective, this is doing a redundant type check at runtime -- first within the match and then again in `canEqual`.

photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/MultiEvaluator.scala
```diff
@@ -55,11 +55,44 @@ abstract class MultiEvaluator(
...
+
+    case _ => false
+  }
+
+  /**
+   * Returns a hash code value for the object.
+   *
+   * @return An [[Int]] hash code
+   */
+  override def hashCode: Int = (41 * (41 * super.hashCode()) + localEvaluator.hashCode()) + ids.hashCode()
```
joshvfleming: FYI -- There's a Java builtin for building composite hash codes now:

```java
    override def hashCode: Int = Objects.hash(super, localEvaluator, ids)
```

Though it's probably just going to do what you've done here.

PR #329, opened 2017-12-05
Update DataValidators error message

ashelkovnykov:
Update the error messages in `DataValidators` to be more specific.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
@@ -34,33 +34,31 @@ object DataValidators extends Logging {
 
   // (Validator, Error Message) pairs
   val baseValidators: List[((LabeledPoint => Boolean), String)] = List(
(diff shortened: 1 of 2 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/data/DataValidators.scala
```diff
@@ -157,15 +158,17 @@ object DataValidators extends Logging {
...
 
   /**
    * Verify that a data point has a finite weight.
    *
    * @param labeledPoint The input data point
    * @return Whether the weight of the input data point is finite
    */
-  def finiteWeight(labeledPoint: LabeledPoint): Boolean = !(labeledPoint.weight.isNaN || labeledPoint.weight.isInfinite)
+  def nonNegativeWeight(labeledPoint: LabeledPoint): Boolean =
+    !(labeledPoint.weight.isNaN || labeledPoint.weight.isInfinite) && (labeledPoint.weight > 0)
```
joshvfleming: Here we're comparing with 0, but below in `rowHasNegativeWeight`, we compare with `MathConst.EPSILON`. Should be consistent.

PR #330, opened 2017-12-11
Upgrade to Vector interfaces in spark.ml

joshvfleming:
New Vector interfaces have been introduced in spark.ml (presumably to replace the spark.mllib interfaces). This change upgrades us to the new interfaces, and fixes a bug where DataFrames produced from AvroDataReader didn't work with estimators in spark.ml.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
@@ -15,7 +15,7 @@
 package com.linkedin.photon.ml.data
 
 import org.apache.spark.broadcast.Broadcast
-import org.apache.spark.mllib.linalg.SparseVector
+import org.apache.spark.ml.linalg.SparseVector
 import org.apache.spark.rdd.RDD
 import org.apache.spark.sql.{DataFrame, Row}
 
@@ -81,7 +81,7 @@ object GameConverters {
 
     val featureShardContainer = featureShards.map { shardId =>
       val features = row.getAs[SparseVector](shardId)
-      (shardId, VectorUtils.mllibToBreeze(features))
+      (shardId, VectorUtils.mlToBreeze(features))
     }.toMap
 
     val response = if (isResponseRequired) {
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
@@ -17,6 +17,7 @@ package com.linkedin.photon.ml.cli.game.training
 import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkContext
 import org.apache.spark.ml.param.{Param, ParamMap, ParamValidators, Params}
+import org.apache.spark.ml.linalg.{Vector => SparkMLVector}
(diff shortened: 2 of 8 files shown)


photon-api/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorIntegTest.scala
```diff
@@ -384,7 +386,28 @@ class GameEstimatorIntegTest extends SparkTestUtils with GameTestUtils {
...
+
+    val schema = new StructType(df.schema.map { field =>
+      field.dataType match {
+        case t: VectorUDT => StructField(field.name, VectorType)
+        case other => field
+      }
+    }.toArray)
+
+    spark.createDataFrame(rows, schema)
+  }
```
ashelkovnykov: Is this for backwards compatibility - does the schema of the `DataFrame`s use `mllib.Vector` and `VectorUDT`? Should we just run a job once to save them in the new format instead of adding in this code?

photon-api/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
```diff
@@ -81,7 +82,8 @@ object GameConverters {
 
     val featureShardContainer = featureShards.map { shardId =>
       val features = row.getAs[SparseVector](shardId)
-      (shardId, VectorUtils.mllibToBreeze(features))
+      val breezeVector = new BreezeSparseVector[Double](features.indices, features.values, features.size)
+      (shardId, breezeVector)
```
ashelkovnykov: We should add a TODO that this line should return to as it was before once `mllib.Vector` is replaced throughout all of Photon ML.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -473,7 +475,10 @@ object GameTrainingDriver extends GameDriver {
       featureShards: Iterable[FeatureShardId]): FeatureShardStatistics =
     featureShards.map { featureShardId =>
       // Calling rdd explicitly here to avoid a typed encoder lookup in Spark 2.1
-      (featureShardId, BasicStatisticalSummary(data.select(featureShardId).rdd.map(_.getAs[SparkVector](0))))
+      (featureShardId, BasicStatisticalSummary(
+        data.select(featureShardId)
+          .rdd
+          .map(row => Vectors.fromML(row.getAs[SparkMLVector](0)))))
```
ashelkovnykov: Similar to above: should add a TODO for when `ml.Vector` is the Photon ML standard.

photon-api/src/integTest/scala/com/linkedin/photon/ml/estimators/GameEstimatorIntegTest.scala
joshvfleming: Ideally, yes. Can we easily recreate the parquet files? Is there a script?

joshvfleming: Oh, I see what you mean. It should be easy enough to just load up the DataFrame, run this code, and then re-save it. I'll do that tomorrow morning.

PR #334, opened 2018-01-04
Fix normalization for IndexMapProjection and RandomProjection in RE

alice2008:
@ashelkovnykov @joshvfleming 
Fix normalization for IndexMapProjection and RandomProjection in RE.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/projector/ProjectionMatrixBroadcast.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/projector/ProjectionMatrixBroadcast.scala
@@ -20,8 +20,10 @@ import org.apache.spark.rdd.RDD
 import com.linkedin.photon.ml.constants.MathConst
 import com.linkedin.photon.ml.data.{LabeledPoint, RandomEffectDataSet}
 import com.linkedin.photon.ml.model.Coefficients
+import com.linkedin.photon.ml.normalization.NormalizationContext
 import com.linkedin.photon.ml.spark.BroadcastLike
 import com.linkedin.photon.ml.supervised.model.GeneralizedLinearModel
+import com.linkedin.photon.ml.util.{BroadcastWrapper, PhotonBroadcast}
 
 /**
  * Represents a broadcast projection matrix.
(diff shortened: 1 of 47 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/util/BroadcastWrapper.scala
```diff
@@ -0,0 +1,35 @@
...
+  def value: T
+}
+
+case class PhotonBroadcast[T](va: Broadcast[T]) extends BroadcastWrapper[T] {
+  def value = va.value
+
+  def unpersist(): Unit = { va.unpersist() }
+}
+
+case class NonPhotonBroadcast[T](value: T) extends BroadcastWrapper[T]
```
joshvfleming: I'm finding the `NonPhotonBroadcast` naming super confusing. I think something like `PhotonNonBroadcast` or `PhotonNullBroadcast` would be clearer.

photon-api/src/main/scala/com/linkedin/photon/ml/projector/ProjectionMatrixBroadcast.scala
```diff
@@ -72,6 +74,24 @@ protected[ml] class ProjectionMatrixBroadcast(projectionMatrixBroadcast: Broadca
...
   }
 
+  /**
+   *
+   * @param randomEffectDataSet The input sharded data set in the original space
+   * @param normalizationContext The original broadcast form of NormalizationContext
+   * @return The broadcast form of NormalizationContext in projected space
+   */
+  def projectNormalizationContext(randomEffectDataSet: RandomEffectDataSet, normalizationContext: BroadcastWrapper[NormalizationContext])
+    : BroadcastWrapper[NormalizationContext] = {
```
joshvfleming: The convention here is to give each arg its own line, indented 4 spaces, e.g.:

```scala
def projectNormalizationContext(
    randomEffectDataSet: RandomEffectDataSet,
    normalizationContext: BroadcastWrapper[NormalizationContext]): BroadcastWrapper[NormalizationContext] = {
```

photon-api/src/main/scala/com/linkedin/photon/ml/projector/ProjectionMatrixBroadcast.scala
```diff
@@ -72,6 +74,24 @@ protected[ml] class ProjectionMatrixBroadcast(projectionMatrixBroadcast: Broadca
...
+    val sc = randomEffectDataSet.sparkContext
+    val originalNormalizationContext = normalizationContext.value
+    val factors = originalNormalizationContext.factors.map(factors => projectionMatrixBroadcast.value.projectFeatures(factors))
+    val shifts = originalNormalizationContext.shifts.map(shifts => projectionMatrixBroadcast.value.projectFeatures(shifts))
+    val interceptId = originalNormalizationContext.interceptId.map(_ =>
+      projectionMatrixBroadcast.value.projectedSpaceDimension - 1)
```
joshvfleming: It seems odd that we'd need to know the intercept's position in the vector in this function. I think we should add a function / property to ProjectionMatrix that serves the same purpose as `projector.originalToProjectedSpaceMap(interceptId)` above.

joshvfleming:
One more thing -- please rebase and squash all the minor / temp commits.

PR #335, opened 2018-01-06
Fixes instructions in readme wrt getting started

skrawcz:
Problem: LinkedIn uses a modified 2.10. The repo defaults to using it. The readme does not make this clear.

Solution: Add specific instructions on how to change the scala version so that someone outside of LinkedIn can actually run it.

Also the diagnostic report isn't produced unless you specify it as an argument.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,8 +209,27 @@ For more information, see the [Spark docs](http://spark.apache.org/docs/latest/i
 ```
 git clone git@github.com:linkedin/photon-ml.git
 cd photon-ml
+```
+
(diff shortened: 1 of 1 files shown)


ashelkovnykov:
Thanks for bringing this up. The README in general needs to be updated for the latest changes - I don't know if we want to push this exact text, but we'll definitely make sure to add a note about this in the updated README.

joshvfleming:
Photon now crossbuilds to 2.10 and 2.11.

PR #336, opened 2018-01-10
Patch to fix evaluator mismatch error

ashelkovnykov:
- Added comment explaining problem and temporary solution
- Added new unit test for hyper-parameter tuning which doubles as test for evaluator mismatch error
- Replace hard-coded names with references to constants in the GameTrainingDriverIntegTest

### DIFF
--- a/photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
+++ b/photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
@@ -25,7 +25,7 @@ import org.testng.Assert._
 import org.testng.annotations.{DataProvider, Test}
 
 import com.linkedin.photon.avro.generated.BayesianLinearModelAvro
-import com.linkedin.photon.ml.{DataValidationType, TaskType}
+import com.linkedin.photon.ml.{DataValidationType, HyperparameterTuningMode, TaskType}
 import com.linkedin.photon.ml.cli.game.GameDriver
 import com.linkedin.photon.ml.constants.StorageLevel
 import com.linkedin.photon.ml.data.{FixedEffectDataConfiguration, GameConverters, RandomEffectDataConfiguration}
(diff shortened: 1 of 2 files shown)


photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
```diff
@@ -131,6 +131,48 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
     assertTrue(AvroUtils.modelContainsIntercept(sc, bestFixedEffectModelPath))
   }
```
alice2008: It would be good idea to add integTest for hyper-parameter tuning for randomEffect as well.

photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
```diff
@@ -131,6 +131,48 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
...
+  /**
+   * Test GAME training with a fixed effect model only and hyperparameter tuning. Note that the best model may not be
+   * one of the tuned models.
+   */
+  @Test
+  def testFixedEffectsWithHyperparameterTuning(): Unit =
+    sparkTest("testFixedEffectsWithHyperparameterTuning", useKryo = true) {
+
+      // This is a baseline RMSE capture from an assumed-correct implementation on 4/14/2016
+      val errorThreshold = 1.7
```
alice2008: errorThreshold is 2.37 for other tests. 1.7 is lower but i wonder which one is correct.

ashelkovnykov:
Please ignore the man behind the curtain

ashelkovnykov:
@alice2008 

The RMSE thresholds before this PR were as follows:
```
Fixed effect only = 1.7
Random effect only = 2.34
Mixed effect = 2.2
```
I did a check on what the tests are returning now and found the following:
```
Fixed effect only = 1.1
Random effect only = 2.33
Mixed effect = 0.95
```
I've update the tests with new thresholds to reflect this.

As to why the random effects are so poor on their own: the data set used for this integration test is quite small (34810 records). For the `per-song` coordinate the stats are as follows: `33110 unique entities, 1.05 avg. # samples, 8 max # samples`. For the `per-artist` coordinate, the stats are better: `4471 unique entities, 7.8 avg. # samples, 189 max # samples. The validation data set is even smaller. I think it's fair to say that the ratio of signal to noise isn't great for this small data set, but large enough to test the individual Photon ML features comparatively.

photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
ashelkovnykov: Updated to have the hyper-parameter tuning test run on mixed effects and compare the tuned models against the best model.

PR #346, opened 2018-03-22
Update to use prior data in autotuning

basukinjal:
The code change necessary for using prior data across different dataset to speed up autotuning. The code is backward compatible and tested.

### DIFF
--- a/photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
+++ b/photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
@@ -94,7 +94,7 @@ class GaussianProcessSearch[T](
 
         // Finding the overall bestEval
         val currentMean =  mean(evals)
-        val overallBestEval = if(evaluator.betterThan(priorBestEval, bestEval - currentMean)) {
+        val overallBestEval = if (evaluator.betterThan(priorBestEval, bestEval - currentMean)) {
           priorBestEval
         } else {
           bestEval - currentMean
--- a/photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
+++ b/photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
@@ -14,11 +14,8 @@
  */
 package com.linkedin.photon.ml.hyperparameter.search
 
-import math.max
-
 import breeze.linalg.{DenseMatrix, DenseVector}
-import breeze.numerics.sqrt
-import breeze.stats.variance
+import breeze.stats.mean
 
 import com.linkedin.photon.ml.evaluation.Evaluator
 import com.linkedin.photon.ml.hyperparameter.EvaluationFunction
@@ -67,6 +64,9 @@ class GaussianProcessSearch[T](
   private var observedPoints: Option[DenseMatrix[Double]] = None
   private var observedEvals: Option[DenseVector[Double]] = None
   private var bestEval: Double = evaluator.defaultScore
+  private var priorObservedPoints: Option[DenseMatrix[Double]] = None
+  private var priorObservedEvals: Option[DenseVector[Double]] = None
+  private var priorBestEval: Double = evaluator.defaultScore
   private var lastModel: GaussianProcessModel = _
 
   /**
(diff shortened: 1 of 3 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -92,17 +92,33 @@ class GaussianProcessSearch[T](
...
         val estimator = new GaussianProcessEstimator(
           kernel = kernel,
-          normalizeLabels = true,
+          normalizeLabels = false,
           noisyTarget = noisyTarget,
           predictionTransformation = Some(transformation),
           seed = seed)
 
-        val model = estimator.fit(points, evals)
+        //Union of points and evals with priorData
```
joshvfleming: Missing space after `//`

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -92,17 +92,33 @@ class GaussianProcessSearch[T](
...
+        val (overallPoints, overallEvals) = if(priorObservedPoints.isDefined) {
+          (DenseMatrix.vertcat(points, priorObservedPoints.get),
+            DenseVector.vertcat(evals - currentMean, priorObservedEvals.get))
+        } else {
+          (points, evals - currentMean)
+        }
```
joshvfleming: It would be more idiomatic to set this up as a pattern match, e.g.:

```scala
val (overallPoints, overallEvals) = (priorObservedPoints, priorObservedEvals) match {
  case (Some(priorPoints), Some(priorEvals)) => (
    DenseMatrix.vertcat(points, priorPoints),
    DenseVector.vertcat(evals - currentMean, priorEvals))
  case _ =>
    (points, evals - currentMean)
}
```

In general, when you find yourself using the `get` function to unwrap an `Option`, there's usually a better way to handle it.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -92,17 +92,33 @@ class GaussianProcessSearch[T](
         // hyperparameter spaces.
         val kernel = new Matern52
 
+        //Finding the overall bestEval
```
joshvfleming: Missing space after `//`.

Also, we should include a brief explanation that prior evals are expected to be mean-centered, and that's why we're doing this.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -116,22 +132,65 @@ class GaussianProcessSearch[T](
...
+   *
+   * @param point the observed point in the space
+   * @param eval the observed value
+   */
+  protected[search] override def onObservation(
+      point: DenseVector[Double],
+      eval: Double,
+      priorData: Boolean = false): Unit = {
+
+    if(priorData) {
```
joshvfleming: If `priorData` is true only within this class, it would be better to handle this with a separate `onPriorObservation` function, IMO.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -53,17 +51,29 @@ class RandomSearch[T](
...
     observations.init.foreach { case (candidate, value) =>
       onObservation(candidate, value)
     }
 
+    if(priorObservations.isDefined) {
+      // Load the prior observations. We add all of them since we do not iterate over these.
+      priorObservations.get.foreach { case (candidate, value) =>
+        onObservation(candidate, value, priorObservations.isDefined)
+      }
+    }
```
joshvfleming: This is another case where pattern matching would be more idiomatic:

```scala
priorObservations match {
  case Some(priorObs) => priorObs.foreach { case (candidate, value) =>
    onObservation(candidate, value, priorObservations.isDefined)
  }
}
```

Also, adding an `onPriorObservation` function as suggested above would eliminate the need for the `isDefined`.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
basukinjal: Done

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
basukinjal: Done

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
basukinjal: Thanks Josh. I've made the changes accordingly.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
basukinjal: Thanks! I'll do the change.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
basukinjal: Thanks. This is now changed.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -92,17 +92,34 @@ class GaussianProcessSearch[T](
         // hyperparameter spaces.
         val kernel = new Matern52
 
+        // Finding the overall bestEval
+        val currentMean =  mean(evals)
+        val overallBestEval = if(evaluator.betterThan(priorBestEval, bestEval - currentMean)) {
```
joshvfleming: Missing space after `if`. Otherwise LGTM.

PR #348, opened 2018-04-06
Fix three partial retraining bugs

ashelkovnykov:
1. Would attempt to build index map using inefficient legacy code for coordinates with missing feature shard definitions
2. Would always attempt to load all coordinates from model

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ModelProcessingUtils.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ModelProcessingUtils.scala
@@ -130,22 +130,20 @@ object ModelProcessingUtils {
   /**
    * Load a GAME model from HDFS.
    *
-   * This method can be called with or without a feature index. If a feature index is not provided, one is created
-   * by scanning the loaded models. In that case, the indexes ranges are [0..numNonZeroFeatures], even if the feature
-   * index used before saving the model was sparse. In other words, when no feature index is provided for the load,
-   * the feature index before the save might not be the same as after the load (it will be more "compact" after the
-   * load, using only contiguous indexes).
-   *
+   * @param sc The execution [[SparkContext]]
+   * @param modelsDir The path to the directory on HDFS where the models are stored
+   * @param storageLevel The level of Spark caching to perform on the loaded model
    * @param featureShardIdToIndexMapLoader A map of feature shard to index map loader
-   * @param modelsDir The directory on HDFS where the models are stored
-   * @param sc The Spark context
+   * @param coordinatesToLoadOpt An optional set of coordinates to load by name (if not provided, all coordinates are
+   *                             loaded)
    * @return The GAME model and feature index
    */
   def loadGameModelFromHDFS(
       sc: SparkContext,
       modelsDir: Path,
       storageLevel: StorageLevel,
-      featureShardIdToIndexMapLoader: Map[FeatureShardId, IndexMapLoader]): GameModel = {
+      featureShardIdToIndexMapLoader: Map[FeatureShardId, IndexMapLoader],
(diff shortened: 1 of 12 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ModelProcessingUtils.scala
```diff
@@ -159,53 +153,96 @@ object ModelProcessingUtils {
...
-        val Array(randomEffectType, featureShardId) = IOUtils.readStringsFromHDFS(idInfoPath, configuration).toArray
+          // Load the model ID info
+          val idInfoPath = new Path(innerPath, AvroConstants.ID_INFO)
 
-        // Load the models
-        val featureMapLoader = featureShardIdToIndexMapLoader.flatMap(_.get(featureShardId))
-        val modelsRDDInputPath = new Path(innerPath, AvroConstants.COEFFICIENTS)
-        val (modelsRDD, featureIndexLoader) =
-          loadModelsRDDFromHDFS(modelsRDDInputPath.toString, featureMapLoader, sc)
+          println(IOUtils.readStringsFromHDFS(idInfoPath, configuration).toArray.mkString(", "))
```
joshvfleming: Looks like something left over from debugging.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ModelProcessingUtils.scala
```diff
@@ -159,53 +153,96 @@ object ModelProcessingUtils {
...
+        .map { innerPath =>
+
+          val name = innerPath.getName
+
+          // Load the model ID info
+          val idInfoPath = new Path(innerPath, AvroConstants.ID_INFO)
+          val Array(featureShardId) = IOUtils.readStringsFromHDFS(idInfoPath, configuration).toArray
+
+          require(
+            featureShardId != null && !featureShardId.isEmpty,
```
joshvfleming: Why would `featureShardId` be null? If we don't get a match on line 172, it will result in `MatchError`. Is it possible that `IOUtils.readStringsFromHDFS` would return something like `Array(null)`? On a quick reading of the function, it doesn't appear so.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ModelProcessingUtils.scala
```diff
@@ -159,53 +153,96 @@ object ModelProcessingUtils {
...
-        val featureMapLoader = featureShardIdToIndexMapLoader.flatMap(_.get(featureShardId))
-        val modelsRDDInputPath = new Path(innerPath, AvroConstants.COEFFICIENTS)
-        val (modelsRDD, featureIndexLoader) =
-          loadModelsRDDFromHDFS(modelsRDDInputPath.toString, featureMapLoader, sc)
+          println(IOUtils.readStringsFromHDFS(idInfoPath, configuration).toArray.mkString(", "))
+
+          val Array(randomEffectType, featureShardId) = IOUtils.readStringsFromHDFS(idInfoPath, configuration).toArray
+
+          require(
+            featureShardId != null && !featureShardId.isEmpty,
```
joshvfleming: Same comment as above. I don't think `featureShardId == null` is really a reachable state here.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/ModelProcessingUtils.scala
ashelkovnykov: I think I'm going to rewrite the check instead, to explicitly check for metadata input errors.

ashelkovnykov: Or maybe not - either way, you're right.

PR #349, opened 2018-04-06
Add computation of the full Hessian matrix for computing variance

joshvfleming:
Here we mean variance in the sense of the posterior model distribution. We approximate the model variance by inverting the final Hessian matrix. This replaces the approximation that used just the Hessian diagonal, because it introduced too much error. See:

https://docs.google.com/document/d/1MAAeSIefs66q0tM2dXu-sNBtBlzUzVCVO7KmhOwCtmQ/edit?usp=sharing

### DIFF
--- a/photon-api/src/integTest/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblemIntegTest.scala
+++ b/photon-api/src/integTest/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblemIntegTest.scala
@@ -16,7 +16,9 @@ package com.linkedin.photon.ml.optimization
 
 import java.util.Random
 
-import breeze.linalg.{DenseVector, Vector}
+import scala.math.abs
+
+import breeze.linalg.{DenseMatrix, DenseVector, Vector, diag, pinv}
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
 import org.mockito.Mockito._
@@ -280,7 +282,7 @@ class DistributedOptimizationProblemIntegTest extends SparkTestUtils {
   def testComputeVariancesSimple(
       dataGenerationFunction: () => Seq[LabeledPoint],
       lossFunction: PointwiseLossFunction,
-      resultDirectDerivationFunction: (Vector[Double]) => (Vector[Double], LabeledPoint) => Vector[Double]): Unit =
+      resultDirectDerivationFunction: (Vector[Double]) => (DenseMatrix[Double], LabeledPoint) => DenseMatrix[Double]): Unit =
     sparkTest("testComputeVariancesSimple") {
       val input = sc.parallelize(dataGenerationFunction())
       val coefficients = generateDenseVector(DIMENSIONS)
@@ -304,12 +306,12 @@ class DistributedOptimizationProblemIntegTest extends SparkTestUtils {
         NoRegularizationContext,
         isComputingVariances = true)
 
-      val hessianDiagonal = input.treeAggregate(new DenseVector[Double](DIMENSIONS).asInstanceOf[Vector[Double]])(
+      val hessianMatrix = input.treeAggregate(DenseMatrix.zeros[Double](DIMENSIONS, DIMENSIONS))(
         seqOp = resultDirectDerivationFunction(coefficients),
-        combOp = (vector1: Vector[Double], vector2: Vector[Double]) => vector1 + vector2,
+        combOp = (matrix1: DenseMatrix[Double], matrix2: DenseMatrix[Double]) => matrix1 + matrix2,
         depth = 1)
-      // Simple estimate of the diagonal of the covariance matrix (instead of a full inverse).
-      val expected = hessianDiagonal.map( v => 1D / (v + MathConst.EPSILON))
+
+      val expected = diag(pinv(hessianMatrix))
       val actual: Vector[Double] = optimizationProblem.computeVariances(input, coefficients).get
 
       assertEquals(actual.length, DIMENSIONS)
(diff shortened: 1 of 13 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
```diff
@@ -55,13 +55,38 @@ object Linalg {
     DenseVector(bArr)
   }
 
+  /**
+   * Inverts the matrix $A$, given the Cholesky factorization $l$ of $A$.
+   *
+   * @param l the Cholesky factorization of matrix A
+   * @return the solution
```
ashelkovnykov: Capitalize 'the' please for both lines above (for consistency, but mostly for me)

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
```diff
@@ -55,13 +55,38 @@ object Linalg {
...
+    val n = l.rows
+
+    // Lapack will write the result to this array
+    val lArr = l.toArray
+
+    val info = new intW(0)
+    lapack.dpotri("L", n, lArr, n, info)
+
+    if (info.`val` < 0) {
+      throw new RuntimeException(
```
ashelkovnykov: Can fit on one line

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
```diff
@@ -55,13 +55,38 @@ object Linalg {
...
+        s"The value at position ${-info.`val`} had an illegal value.")
+    }
```
ashelkovnykov: Please add some comments to this function. What is the `dpotri` call? What is `info`? What is `'val'`? What is an illegal value in this context?

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
```diff
@@ -123,8 +124,7 @@ protected[ml] class DistributedOptimizationProblem[Objective <: DistributedObjec
       logger.info(s"History tracker information:\n $tracker")
       val modelsPerIteration = tracker.getTrackedStates.map { x =>
         val coefficients = x.coefficients
-        val variances = computeVariances(input, coefficients)
-        createModel(normalizationContext, coefficients, variances)
+        createModel(normalizationContext, coefficients, variances = None)
```
ashelkovnykov: Variances permanently disabled?

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
joshvfleming: IMO it's better for readability to break `if` statements like this into multiple lines, even if they could technically be one-liners.

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
joshvfleming: I think what I'll do instead is refer to LAPACK documentation, besides a brief description of `dpotri`. It's really their job to document these things.

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
joshvfleming: Will do.

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
```diff
@@ -55,13 +55,38 @@ object Linalg {
...
   }
 
+  /**
+   * Inverts the matrix $A$, given the Cholesky factorization $l$ of $A$.
+   *
+   * @param l the Cholesky factorization of matrix A
+   * @return the solution
+   */
+  protected[ml] def choleskyInverse(
+      l: DenseMatrix[Double]): DenseMatrix[Double] = {
```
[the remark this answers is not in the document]
joshvfleming: Will do.

photon-api/src/integTest/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblemIntegTest.scala
```diff
@@ -374,6 +376,86 @@ class DistributedOptimizationProblemIntegTest extends SparkTestUtils {
...
+      0.01292064,
+      10.37188)
+
+    val actual: Vector[Double] = optimizationProblem.computeVariances(input, coefficients).get
+
+    assertEquals(actual.length, expected.length)
+    for (i <- 0 until expected.length) {
+      val relDiff = abs(actual(i) / expected(i) - 1)
+      assertTrue(relDiff < 0.001)
+    }
```
ashelkovnykov: Take a look at `VectorUtils.areAlmostEqual` - it's not quite the same thing but might be passable for this case

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
ashelkovnykov: My bad, got this call confused with the one 6 lines down

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
ashelkovnykov: Fair enough

photon-lib/src/main/scala/com/linkedin/photon/ml/util/Linalg.scala
ashelkovnykov: I meant that `throw new RuntimeException(...)` could fit on one line.

I don't think it's right, since it would return an `IllegalArgumentException` and it's not an input, but we could replace the whole `if` with a `require`.

photon-api/src/integTest/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblemIntegTest.scala
joshvfleming: I wanted to do a relative comparison here, since we have variances across different orders of magnitude.

ashelkovnykov: Sure - wasn't sure how big the difference was between R and our predictions in practice

PR #354, opened 2018-04-19
Add rescaling hyper-parameters in covariance functions

cmjiang:
@basukinjal @joshvfleming @ashelkovnykov Please review the code.

### DIFF
--- a/photon-all/build.gradle
+++ b/photon-all/build.gradle
@@ -52,6 +52,12 @@ sourcesJar {
 }
 */
 
+configurations {
+  all {
+    exclude group: 'org.xerial.snappy'
+  }
+}
+
 shadowJar {
   // Remove the classifier as we want the shadow jar to be the main jar
   classifier = ''
@@ -89,11 +95,6 @@ shadowJar {
   relocate 'org.tukaani', 'photonml.shaded.org.tukaani'
   relocate 'scopt', 'photonml.shaded.scopt'
   relocate 'org.joda.time', 'photonml.shaded.org.joda.time'
-
-  // Deliberately not relocating Snappy. The reason: Snappy includes native libraries that speed up the
-  // compression / decompression process. After relocation, it can't find the native libraries anymore.
-  //relocate 'org.xerial.snappy', 'org.xerial.snappy'
-
 }
 
 artifacts {


photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/estimators/kernels/StationaryKernel.scala
```diff
@@ -56,6 +58,22 @@ abstract class StationaryKernel(
...
+   * Unwraps hyper-parameters from scaled values to true values
+   *
+   * @param x the matrix of points, where each of the m rows is a point in the space and some columns contain scaled values
+   * @return the unwrapped x that all the columns have true values
+   */
+  protected[kernels] def unwrap(x: DenseMatrix[Double]): DenseMatrix[Double] = {
+    val u = x.copy
+    indexedTransformMap.map { case (index: Int, transform: String) => transform match {
+        case "LOG" => u(::, index) := u(::, index).map(element => Math.pow(10, element))
+        case "SQRT" => u(::, index) := u(::, index).map(element => Math.pow(element, 2))
```
joshvfleming: Let's pass the function directly (e.g. `Map(0 -> (x: Double) => Math.pow(10, x))`. It will make this code cleaner, and it will also remove the burden from keeping a list of transformations here -- the kernel doesn't really need to have this information.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/estimators/kernels/StationaryKernel.scala
```diff
@@ -56,6 +58,22 @@ abstract class StationaryKernel(
    */
   protected[kernels] def fromPairwiseDistances(dists: DenseMatrix[Double]): DenseMatrix[Double]
 
+  /**
+   * Unwraps hyper-parameters from scaled values to true values
+   *
+   * @param x the matrix of points, where each of the m rows is a point in the space and some columns contain scaled values
+   * @return the unwrapped x that all the columns have true values
+   */
+  protected[kernels] def unwrap(x: DenseMatrix[Double]): DenseMatrix[Double] = {
```
joshvfleming: I think "transform" would be a better name for this function, because from the point of view of the kernel, the values are being transformed. It's not really aware of any prior "wrapping".

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -56,10 +57,11 @@ class GaussianProcessSearch[T](
     evaluationFunction: EvaluationFunction[T],
     evaluator: Evaluator,
     discreteParams: Seq[Int] = Seq(),
+    indexedTransformMap: Map[Int, String] = Map(),
```
joshvfleming: Let's pass the kernel here instead of the transform map. `GaussianProcessSearch` really operates at the "kernel" level of specificity, not the "value" level. It should default to `new Matern52`.

cmjiang:
@basukinjal @joshvfleming Changes updated accordingly.

PR #360, opened 2018-04-27
Patch the AvroUtils.saveAsAvro to pass in JobConf

alice2008:
Unit tests failed because we have a local MiniDfsCluster. Without passing-in JobConf, the unit tests always try to write to local disk instead of MiniDfsCluster.


photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroUtils.scala
```diff
@@ -96,8 +96,8 @@ object AvroUtils {
    * @param outputDir The output directory to save the data as Avro files
    * @param schemaString The schema string of the data
    */
-  protected[ml] def saveAsAvro[T <: SpecificRecord : ClassTag](data: RDD[T], outputDir: String, schemaString: String): Unit = {
-    val job = new JobConf
+  protected[ml] def saveAsAvro[T <: SpecificRecord : ClassTag](data: RDD[T], outputDir: String, schemaString: String, jobConf: Option[JobConf] = None): Unit = {
```
joshvfleming: This doesn't need to be an `Option` -- we can just default to `new JobConf` in the signature, e.g.:

```scala
protected[ml] def saveAsAvro[T <: SpecificRecord : ClassTag](
    data: RDD[T],
    outputDir: String,
    schemaString: String,
    jobConf: JobConf = new JobConf): Unit = {
```

joshvfleming:
Don't we also need to pass in `new JobConf(sc.hadoopConfiguration)` from `AvroDataWriter` now? I didn't see that here.

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroUtils.scala
alice2008: done

PR #361, opened 2018-05-08
support null types for offset and weight

alice2008:
@joshvfleming @ashelkovnykov

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataWriter.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataWriter.scala
@@ -44,8 +44,8 @@ class AvroDataWriter {
     require(columns.contains(responseColumn), s"There must be a $responseColumn column present in dataframe")
     require(columns.contains(featureColumn), s"There must be a $featureColumn column present in dataframe")
 
-    val hasOffset = columns.contains("offset")
-    val hasWeight = columns.contains("weight")
+    val hasOffset = columns.contains(OFFSET)
+    val hasWeight = columns.contains(WEIGHT)
 
     val avroDataset = df.rdd.mapPartitions { rows =>
       val indexMap = indexMapLoader.indexMapForRDD()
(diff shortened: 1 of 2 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataWriter.scala
```diff
@@ -93,6 +93,11 @@ object AvroDataWriter {
   protected[data] def getValueAsDouble(row: Row, fieldName: String): Double = {
 
     row.getAs[Any](fieldName) match {
+      case null => fieldName match {
+        case "offset" => 0.0D
+        case "weight" => 1.0D
```
joshvfleming: We should extract these out into constants for here and L57-58. Actually, the values should be in constants too. Also, we might want to abstract out the lookup of default values, e.g.:

```scala
    val OFFSET = "offset"
    val WEIGHT = "weight"

    val defaults = Map(
      OFFSET -> 0.0D,
      WEIGHT -> 1.0D)

    // ...
    row.getAs[Any](fieldName) match {
      case null => defaults.get(fieldName)
        .getOrElse {
          throw new IllegalArgumentException(s"Unsupported null for fieldName $fieldName") }
      // ..
```

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataWriter.scala
```diff
@@ -54,8 +54,8 @@ class AvroDataWriter {
       rows.map { r: Row =>
         val features = r.getAs[Vector](featureColumn)
         val response = getValueAsDouble(r, responseColumn)
-        val offset = if (hasOffset) getValueAsDouble(r, "offset") else 0.0D
-        val weight = if (hasWeight) getValueAsDouble(r, "weight") else 1.0D
+        val offset = if (hasOffset) getValueAsDouble(r, OFFSET) else 0.0D
+        val weight = if (hasWeight) getValueAsDouble(r, WEIGHT) else 1.0D
```
joshvfleming:
```scala
/* ... */ else DEFAULTS(OFFSET)
/* ... */ else DEFAULTS(WEIGHT)
```

photon-client/src/main/scala/com/linkedin/photon/ml/data/avro/AvroDataWriter.scala
```diff
@@ -127,4 +133,4 @@ object AvroDataWriter {
     }
     avroFeatures.toList
   }
-}
+}
```
ashelkovnykov: Missing new line at the end of file

PR #368, opened 2018-06-06
Normalize hyper-parameters in auto-tuning

cmjiang:
@basukinjal Please review this PR for the hyper-parameter normalization.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
@@ -16,7 +16,6 @@ package com.linkedin.photon.ml.cli.game.training
 
 import scala.math.log
 
-import breeze.linalg.DenseVector
 import org.apache.hadoop.fs.Path
 import org.apache.spark.SparkContext
 import org.apache.spark.ml.param.{Param, ParamMap, ParamValidators, Params}
@@ -649,21 +648,22 @@ object GameTrainingDriver extends GameDriver {
         // TODO: Match on this to make it clearer
         val evaluator = models.head._2.get.head._1
         val baseConfig = models.head._3
-        val evaluationFunction = new GameEstimatorEvaluationFunction(estimator, baseConfig, trainingData, testData)
-
+        val dimension = baseConfig.toSeq.length
         val range = getOrDefault(hyperParameterTuningRange)
-        val ranges = List.fill(evaluationFunction.numParams)(DoubleRange(log(range.start), log(range.end)))
+        val ranges = List.fill(dimension)(DoubleRange(log(range.start), log(range.end)))
+
+        val evaluationFunction = new GameEstimatorEvaluationFunction(estimator, baseConfig, trainingData, testData, ranges)
 
         val searcher = getOrDefault(hyperParameterTuning) match {
           case HyperparameterTuningMode.BAYESIAN =>
-            new GaussianProcessSearch[GameEstimator.GameResult](ranges, evaluationFunction, evaluator)
+            new GaussianProcessSearch[GameEstimator.GameResult](dimension, evaluationFunction, evaluator)
 
           case HyperparameterTuningMode.RANDOM =>
-            new RandomSearch[GameEstimator.GameResult](ranges, evaluationFunction)
+            new RandomSearch[GameEstimator.GameResult](dimension, evaluationFunction)
         }
-        val observations = searcher.convertObservations(models)
+        val observations = evaluationFunction.convertObservations(models)
 
-        searcher.findWithObservations(getOrDefault(hyperParameterTuningIter), observations)
+        searcher.findWithPriors(getOrDefault(hyperParameterTuningIter), observations, Seq())
 
       case _ => Seq()
     }
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
@@ -21,20 +21,26 @@ import breeze.linalg.DenseVector
 import org.apache.spark.sql.DataFrame
 
 import com.linkedin.photon.ml.estimators.GameEstimator.{GameOptimizationConfiguration, GameResult}
-import com.linkedin.photon.ml.hyperparameter.EvaluationFunction
+import com.linkedin.photon.ml.hyperparameter.{EvaluationFunction, VectorRescaling}
 import com.linkedin.photon.ml.optimization.game.{FixedEffectOptimizationConfiguration, RandomEffectOptimizationConfiguration}
-
+import com.linkedin.photon.ml.util.DoubleRange
 /**
  * Evaluation function implementation for GAME.
  *
  * An evaluation function is the integration point between the hyperparameter tuning module and an estimator, or any
  * system that can unpack a vector of values and produce a real evaluation.
+ * @param estimator The estimator for GAME model
+ * @param baseConfig The initial configuration supplied by the user
+ * @param data Training data
+ * @param validationData Validation data
+ * @param ranges Hyper-parameter searching range. If search space is [a, b], range here is [log(a), log(b)].
  */
 class GameEstimatorEvaluationFunction(
     estimator: GameEstimator,
     baseConfig: GameOptimizationConfiguration,
     data: DataFrame,
-    validationData: DataFrame)
+    validationData: DataFrame,
+    ranges: Seq[DoubleRange])
   extends EvaluationFunction[GameResult] {
 
   // CoordinateOptimizationConfigurations sorted in order by coordinate ID name
@@ -46,17 +52,37 @@ class GameEstimatorEvaluationFunction(
   /**
    * Performs the evaluation.
    *
-   * @param hyperParameters The vector of hyperparameter values under which to evaluate the function
+   * @param candidate The candidate vector of hyperparameter with values in [0, 1]
    * @return A tuple of the evaluated value and the original output from the inner estimator
    */
-  override def apply(hyperParameters: DenseVector[Double]): (Double, GameResult) = {
-    val newConfiguration = vectorToConfiguration(hyperParameters)
+  override def apply(candidate: DenseVector[Double]): (Double, GameResult) = {
+
+    val candidateScaled = VectorRescaling.scaleBackward(candidate, ranges)
+
+    val newConfiguration = vectorToConfiguration(candidateScaled)
 
     val model = estimator.fit(data, Some(validationData), Seq(newConfiguration)).head
     val (_, Some(evaluations), _) = model
 
     // Assumes model selection evaluator is in "head" position
     (evaluations.head._2, model)
+  }
+
+  /**
+   * Vectorize and scale a [[Seq]] of prior observations.
+   *
+   * @param observations Prior observations in estimator output form
+   * @return Prior observations as tuples of (vector representation of the original estimator output, evaluated value)
+   */
+  override def convertObservations(observations: Seq[GameResult]): Seq[(DenseVector[Double], Double)] = {
+
+    observations.map { observation =>
+      val candidate = vectorizeParams(observation)
+      val candidateScaled = VectorRescaling.scaleForward(candidate, ranges)
+      val value = getEvaluationValue(observation)
+
+      (candidateScaled, value)
+    }
   }
 
   /**
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
@@ -1,0 +1,134 @@
+/*
+ * Copyright 2018 LinkedIn Corp. All rights reserved.
+ * Licensed under the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License. You may obtain a
+ * copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+package com.linkedin.photon.ml.hyperparameter
+
+import breeze.linalg.DenseVector
+
+import com.linkedin.photon.ml.util.DoubleRange
+
+/**
+ * Functions to transform/scale forward/backward hyper-parameters in vectors
+ */
+object VectorRescaling {
+
+  val LOG_TRANSFORM = "LOG"
+  val SQRT_TRANSFORM = "SQRT"
+
+  /**
+   * Apply forward transformation to a subset of elements in a vector.
+   *
+   * @param vector A DenseVector.
+   * @param transformMap A Map with key-value pairs of indices and names of transform functions.
+   * @return The transformed vector.
+   */
+  def transformForward(
+      vector: DenseVector[Double],
+      transformMap: Map[Int, String]): DenseVector[Double] = {
+
+    val vectorTransformed = vector.copy
+
+    transformMap.foreach { case (index, transform) =>
+      transform match {
+        case LOG_TRANSFORM => vectorTransformed(index) = Math.log10(vectorTransformed(index))
+        case SQRT_TRANSFORM => vectorTransformed(index) = Math.sqrt(vectorTransformed(index))
+        case other => throw new IllegalArgumentException(s"Unknown transformation: $other")
+      }
+    }
+
+    vectorTransformed
+  }
+
+  /**
+   * Apply backward transformation to a subset of elements in a vector.
+   *
+   * @param vector A DenseVector.
+   * @param transformMap A Map with key-value pairs of indices and names of transform functions.
+   * @return The transformed vector.
+   */
+  def transformBackward(
+      vector: DenseVector[Double],
+      transformMap: Map[Int, String]): DenseVector[Double] = {
+
+    val vectorTransformed = vector.copy
+
+    transformMap.foreach { case (index, transform) =>
+      transform match {
+        case LOG_TRANSFORM => vectorTransformed(index) = Math.pow(10, vectorTransformed(index))
+        case SQRT_TRANSFORM => vectorTransformed(index) = Math.pow(vectorTransformed(index), 2)
+        case other => throw new IllegalArgumentException(s"Unknown transformation: $other")
+      }
+    }
+
+    vectorTransformed
+  }
+
+  /**
+   * Apply forward scaling to a vector. Given a range [a, b] and an element x of the vector,
+   * y = (x - a) / (b - a), if x is continuous;
+   * y = (x - a) / (b - a + 1), if x is discrete.
+   *
+   * @param vector A DenseVector.
+   * @param ranges A sequence of ranges for every element in the vector to be scaled.
+   * @param discreteIndexSet A Set with indices of discrete elements in the vector.
+   * @return The scaled vector.
+   */
+  def scaleForward(
+      vector: DenseVector[Double],
+      ranges: Seq[DoubleRange],
+      discreteIndexSet: Set[Int] = Set()): DenseVector[Double] = {
+
+    val vectorScaled = vector.copy
+
+    val start = DenseVector(ranges.map(_.start).toArray)
+    val end = DenseVector(ranges.map(_.end).toArray)
+    val discreteAdj = DenseVector.tabulate(ranges.length) { i =>
+      if (discreteIndexSet.contains(i)) 1.0 else 0.0
+    }
+
+    vectorScaled :-= start
+    vectorScaled :/= (end - start + discreteAdj)
+
+    vectorScaled
+  }
+
+  /**
+    * Apply backward scaling to a vector. Given range [a, b] and an element x of the vector,
+    * y = x * (b - a) + a, if x is continuous;
+    * y = x * (b - a + 1) + a, if x is discrete.
+    *
+    * @param vector A DenseVector.
+    * @param ranges A sequence of ranges for every element in the vector to be scaled.
+    * @param discreteIndexSet A Set with indices of discrete elements in the vector.
+    * @return The scaled vector.
+    */
+  def scaleBackward(
+      vector: DenseVector[Double],
+      ranges: Seq[DoubleRange],
+      discreteIndexSet: Set[Int] = Set()): DenseVector[Double] = {
+
+    val vectorScaled = vector.copy
+
+    val start = DenseVector(ranges.map(_.start).toArray)
+    val end = DenseVector(ranges.map(_.end).toArray)
+    val discreteAdj = DenseVector.tabulate(ranges.length) { i =>
+      if (discreteIndexSet.contains(i)) 1.0 else 0.0
+    }
+
+    vectorScaled :*= (end - start + discreteAdj)
+    vectorScaled :+= start
+
+    vectorScaled
+  }
+}
--- a/photon-client/src/test/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunctionTest.scala
+++ b/photon-client/src/test/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunctionTest.scala
@@ -29,6 +29,7 @@ import com.linkedin.photon.ml.estimators.GameEstimator.GameOptimizationConfigura
 import com.linkedin.photon.ml.optimization._
 import com.linkedin.photon.ml.optimization.game._
 import com.linkedin.photon.ml.test.Assertions.assertIterableEqualsWithTolerance
+import com.linkedin.photon.ml.util.DoubleRange
 
 /**
  * Unit tests for [[GameEstimatorEvaluationFunction]].
@@ -43,6 +44,7 @@ class GameEstimatorEvaluationFunctionTest {
   private val random = new Random(1)
   private val regWeights = Array.fill[Double](4) { random.nextDouble }
   private val tolerance = MathConst.EPSILON
+  private val ranges = Seq(DoubleRange(log(0.001), log(1000)))
 
   /**
    * Test that a [[GameOptimizationConfiguration]] can be correctly constructed from a hyperparameter vector.
@@ -54,7 +56,7 @@ class GameEstimatorEvaluationFunctionTest {
       ("a", FixedEffectOptimizationConfiguration(mockOptimizerConfig, mockRegContext, regWeights(0))),
       ("b", RandomEffectOptimizationConfiguration(mockOptimizerConfig, mockRegContext, regWeights(1))))
 
-    val evaluationFunction = new GameEstimatorEvaluationFunction(mockEstimator, configuration, mockData, mockData)
+    val evaluationFunction = new GameEstimatorEvaluationFunction(mockEstimator, configuration, mockData, mockData, ranges)
     val hypers = DenseVector(log(regWeights(2)), log(regWeights(3)))
     val newConfiguration = evaluationFunction.vectorToConfiguration(hypers)
 
@@ -90,7 +92,7 @@ class GameEstimatorEvaluationFunctionTest {
   @Test(dataProvider = "invalidVectorProvider", expectedExceptions = Array(classOf[IllegalArgumentException]))
   def testInvalidVectorToConfiguration(config: GameOptimizationConfiguration, hypers: DenseVector[Double]): Unit = {
 
-    val evaluationFunction = new GameEstimatorEvaluationFunction(mockEstimator, config, mockData, mockData)
+    val evaluationFunction = new GameEstimatorEvaluationFunction(mockEstimator, config, mockData, mockData, ranges)
     evaluationFunction.vectorToConfiguration(hypers)
   }
 
@@ -118,7 +120,7 @@ class GameEstimatorEvaluationFunctionTest {
   @Test(dataProvider = "configurationProvider")
   def testConfigurationToVector(config: GameOptimizationConfiguration, expected: DenseVector[Double]): Unit = {
 
-    val evaluationFunction = new GameEstimatorEvaluationFunction(mockEstimator, config, mockData, mockData)
+    val evaluationFunction = new GameEstimatorEvaluationFunction(mockEstimator, config, mockData, mockData, ranges)
     val result = evaluationFunction.configurationToVector(config)
 
     assertEquals(result.length, expected.length)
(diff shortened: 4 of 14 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -14,9 +14,7 @@
  */
 package com.linkedin.photon.ml.cli.game.training
 
-import scala.math.log
-
-import breeze.linalg.DenseVector
+import breeze.numerics.log
```
basukinjal: What is the difference between breeze.numerics.log vs scala.math.log? Was there a particular reason to change this?

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -34,7 +33,8 @@ class GameEstimatorEvaluationFunction(
     estimator: GameEstimator,
     baseConfig: GameOptimizationConfiguration,
     data: DataFrame,
-    validationData: DataFrame)
+    validationData: DataFrame,
+    range: DoubleRange)
```
basukinjal: Can you add a documentation here explaining the different parameters? Especially if the range in [0.001, 1000] or [-3, 3]. I feel the range here should be [-3,3]

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -637,21 +635,34 @@ object GameTrainingDriver extends GameDriver {
         // TODO: Match on this to make it clearer
         val evaluator = models.head._2.get.head._1
         val baseConfig = models.head._3
-        val evaluationFunction = new GameEstimatorEvaluationFunction(estimator, baseConfig, trainingData, testData)
 
         val range = getOrDefault(hyperParameterTuningRange)
-        val ranges = List.fill(evaluationFunction.numParams)(DoubleRange(log(range.start), log(range.end)))
+        val rangeWrapped = DoubleRange(log(range.start), log(range.end))
+
+        val evaluationFunction = new GameEstimatorEvaluationFunction(estimator, baseConfig, trainingData, testData, range)
```
basukinjal: Shouldn't you call this with the rangeWrapped? As the candidate is scaled by this range before evaluation. So technically, we should scale [0,1] -> [-3, 3] -> then take exp. Otherwise we are directly scaling [0,1] -> [0.001, 1000]

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -46,11 +46,16 @@ class GameEstimatorEvaluationFunction(
...
+   * @param candidate The candidate vector of hyperparameter with values in [0, 1]
    * @return A tuple of the evaluated value and the original output from the inner estimator
    */
-  override def apply(hyperParameters: DenseVector[Double]): (Double, GameResult) = {
-    val newConfiguration = vectorToConfiguration(hyperParameters)
+  override def apply(candidate: DenseVector[Double]): (Double, GameResult) = {
+
+    // Rescale candidate
+    candidate *= (range.end - range.start)
+    candidate += range.start
```
basukinjal: To reiterate the comment from the Driver. I think you should scale it by rangeWrapped since you are rescaling the priors by range wrapped. And then take the exp. That is [0,1] -> [ -3, 3] -> [0.001, 1000] instead of [0,1] -> [0.001, 1000]

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -637,21 +635,34 @@ object GameTrainingDriver extends GameDriver {
...
         }
         val observations = searcher.convertObservations(models)
 
-        searcher.findWithObservations(getOrDefault(hyperParameterTuningIter), observations)
+        // Wrap priors by a log transformation
+        observations.foreach { case (prior, _) => log.inPlace(prior) }
+
+        // Rescale priors linearly by the range
+        val slope = 1.0 / rangeWrapped.end - rangeWrapped.start
+        val intercept = - range.start / (range.end - range.start)
```
basukinjal: May be im getting this wrong. Isnt range = [0.001, 1000]? If so, then the intercept becomes -0.0001/999.999. 
So, the intercept should also be -rangeWrapped.start*slope instead of using range?

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -68,65 +64,45 @@ class RandomSearch[T](
...
-    }
-
-    // Load the prior data observations
-    priorObservations.foreach { case (candidate, value) =>
-      onPriorObservation(candidate, value)
-    }
-
-    val (results, _) = (0 until n).foldLeft((List.empty[T], observations.last)) {
-      case ((models, (lastCandidate, lastObservation)), _) =>
+    if (observations.isEmpty) {
```
basukinjal: Add a comment above saying that this is the first iteration.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -184,16 +160,21 @@ class RandomSearch[T](
...
+  /**
+   * Discretize candidates whose values are integers after rescaling
+   *
+   * @param candidate candidate with values in [0, 1]
+   * @param discreteParams Map that specifies the indices of discrete parameters and their numbers of discrete values
+   *
+   */
+  protected [search] def discretizeCandidate(candidate: DenseVector[Double], discreteParams: Map[Int, Int]): Unit = {
+    discreteParams.foreach { case (index, numDiscreteValues) =>
+      candidate(index) = floor(candidate(index) * numDiscreteValues) / numDiscreteValues
```
basukinjal: Wont this throw an error since candidate it immutable?

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
cmjiang: I used breeze.numerics.log for the range vectors. Later on I realized that the ranges for different hyper-parameters are fixed. Although it was represented by a (constant) vector. So we can just use scala.math.log.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
cmjiang: Yes, it's a typo. It should be rangeWrapped in line 642

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
cmjiang: You are right. It should be
 val intercept = - rangeWrapped.start / (rangeWrapped.end - rangeWrapped.start)

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
cmjiang: Yes, range is [-3, 3]. Documentation will be added in the code.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
cmjiang: Yes. Actually the variable `range` should always have [-3, 3], never should be [0.001, 1000]. This is true in XGBoost. However in GAME, we use a "LOG" by default and specify 0.001--1000 in configuration. That's the reason for having rangeWrapped, which takes the role of range.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
cmjiang: Strictly speaking, this is the first iteration when `observations` is empty. But in our code, `observations` is initialized explicitly, given by the initial model, the last argument in `runHyperparameterTuning`. So this if section is never executed. It is reserved only for the `find` function for API completeness.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
cmjiang: I think you can't resign another DenseVector to candidate becasue the object reference can not be accessed. But modify the element in a DenseVector is fine.

cmjiang:
Replied to @basukinjal first reviews. Please take a look.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -637,21 +635,34 @@ object GameTrainingDriver extends GameDriver {
...
         val searcher = getOrDefault(hyperParameterTuning) match {
           case HyperparameterTuningMode.BAYESIAN =>
-            new GaussianProcessSearch[GameEstimator.GameResult](ranges, evaluationFunction, evaluator)
+            new GaussianProcessSearch[GameEstimator.GameResult](dimension, evaluationFunction, evaluator)
 
           case HyperparameterTuningMode.RANDOM =>
-            new RandomSearch[GameEstimator.GameResult](ranges, evaluationFunction)
+            new RandomSearch[GameEstimator.GameResult](dimension, evaluationFunction)
         }
         val observations = searcher.convertObservations(models)
```
cmjiang: @basukinjal @joshvfleming I think here was a bug. In the previous version, `models` had values in [0.001, 1000], and observations had [-3, 3], because `convertObservations` did a wrapping. But we should keep all the prior data in [0.001, 1000]. So I suggest we move wrapping/unwrapping out of mapToVec/vecToMap and implement it separately.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -16,13 +16,12 @@ package com.linkedin.photon.ml.estimators
 
 import scala.collection.mutable
 import scala.math.{exp, log}
-
```
ashelkovnykov: Please add this newline back, it was correct style

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -16,13 +16,12 @@ package com.linkedin.photon.ml.estimators
...
 import breeze.linalg.DenseVector
 import org.apache.spark.sql.DataFrame
-
```
ashelkovnykov: Please add this newline back, it was correct style

photon-client/src/test/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunctionTest.scala
```diff
@@ -43,6 +44,7 @@ class GameEstimatorEvaluationFunctionTest {
   private val random = new Random(1)
   private val regWeights = Array.fill[Double](4) { random.nextDouble }
   private val tolerance = MathConst.EPSILON
+  private val range = DoubleRange(0.001, 1000)
```
ashelkovnykov: Based on your conversations with Kinjal, this should actually be `DoubleRange(-3, 3)` - correct?

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -68,65 +64,45 @@ class RandomSearch[T](
...
-        val (observation, model) = evaluationFunction(candidate)
+      val (_, model) = evaluationFunction(candidate)
 
-        (models :+ model, (candidate, observation))
+      Seq(model) ++ (if (n == 1) Seq() else findWithPriors(n - 1, convertObservations(Seq(model)), priorObservations))
     }
 
-    results
-  }
+    else {
```
ashelkovnykov: Lines 77 - 79 should be one line: `} else {`

Any reason that these two functions were merged into one? IMO it's more confusing to follow

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
```diff
@@ -32,9 +32,6 @@ class GaussianProcessSearchTest {
   val seed = 1L
   val dim = 3
   val n = 5
-  val lower = 1e-5
-  val upper = 1e5
-  val ranges: Seq[DoubleRange] = Seq.fill(dim)(DoubleRange(lower, upper))
```
ashelkovnykov: `DoubleRange` import can be removed - it's no longer used

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/estimators/kernels/StationaryKernelTest.scala
```diff
@@ -26,9 +26,7 @@ import com.linkedin.photon.ml.test.Assertions.assertIterableEqualsWithTolerance
 class StationaryKernelTest {
 
   private val tol = 1e-7
-  private val kernel = new RBF(
-    noise = 0.0,
-    indexedTransformMap = Map(1 -> ((x:Double) => Math.pow(10, x)), 3 -> ((x:Double) => Math.pow(x, 2))))
+  private val kernel = new RBF(noise = 0.0)
```
ashelkovnykov: Import on line 21 no longer used

photon-client/src/test/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunctionTest.scala
cmjiang: Yes, you are right. More precisely in GAME model, the code used log to the base e. Here should be DoubleRange(log(0.001), log(1000)).

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
cmjiang: I guess in the previous version, the reason to keep four "find" functions is for four cases with `observations` and `priorObservations` be present or not. Therefore, if statement is needed to decide which function to be called. So I prefer to use a general function to do all the jobs. I am open to change it back. Thanks for your comments.

cmjiang:
Thank @ashelkovnykov for the reviews. All the non-replied comments are appreciated as well and changes have been done.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
joshvfleming: @cmjiang is right -- `DenseVector` is a mutable object.

@basukinjal I'd used `floor` for descretizing in the past, and you asked me to change it to `round`. Do you remember why? I think you were worried about down-biasing candidates.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
joshvfleming: IMO it would be better for readability if we split this into 2 functions rather than one big function bifurcated by an if/else branch.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -59,6 +69,30 @@ class GameEstimatorEvaluationFunction(
...
 
+  /**
+   * Vectorize a [[Seq]] of prior observations.
+   *
+   * @param observations Prior observations in estimator output form
+   * @return Prior observations as tuples of (vector representation of the original estimator output, evaluated value)
+   */
+  override def convertObservations(observations: Seq[GameResult]): Seq[(DenseVector[Double], Double)] = {
+
+    val slope = 1.0 / range.end - range.start
```
joshvfleming: I believe this should be:

```scala
val slope = 1.0 / (range.end - range.start)
```

Also: I'm assuming the purpose for splitting it this way is to avoid recomputing slope/intercept for each observation?

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -59,6 +69,30 @@ class GameEstimatorEvaluationFunction(
     (evaluations.head._2, model)
   }
 
+  /**
+   * Vectorize a [[Seq]] of prior observations.
```
joshvfleming: Vectorize and scale

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
cmjiang: If the argument `observations` in the previous `priorWithPrior` is empty, it will cause an error. We didn't do the check for that. That's why I add an `if` statement. I agree on splitting back to the original four functions but suggest keep this if-else statement.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
cmjiang: Yes, thank you for pointing this out.
And yes, because range is fixed for all.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -117,39 +105,25 @@ class RandomSearch[T](
...
-   * @return Prior observations as tuples of (vector representation of the original estimator output, evaluated value)
-   */
-  def convertObservations(observations: Seq[T]): Seq[(DenseVector[Double], Double)] =
-    observations.map { observation =>
-      val candidate = evaluationFunction.vectorizeParams(observation)
-      val value = evaluationFunction.getEvaluationValue(observation)
-
-      (candidate, value)
-    }
+  def find(n: Int): Seq[T] = findWithPriors(n, Seq(), Seq())
```
basukinjal: I think what @joshvfleming was saying could be used here.

We can call find with the observations and priorObservations. If observations is empty we call findWithPriorObservations else we call findWithPriors. In that case, we can remove the if from the findWithPriors and the code becomes cleaner. 

Otherwise now, from find we are calling findWithPriors with empty observations which is then calling findWithPriorObservations.

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
```diff
@@ -92,32 +94,19 @@ class GaussianProcessSearchTest {
...
-    assertTrue(candidates1.forall(_.params.toArray.forall(x => x >= lower && x <= upper)))
-    assertEquals(candidates1.size, n)
-  }
-
-  @Test(dataProvider = "priorDataProvider", dependsOnMethods = Array[String]("testFindWithObservations"))
   def testFindWithPastObservations(
     currentCandidates: Seq[(DenseVector[Double], Double)],
     priorCandidates: Seq[(DenseVector[Double], Double)],
-    testSetIndex: Int): Unit = {
+    estSetIndex: Int): Unit = {
```
basukinjal: typo, t has been deleted. If the parameter is not used, you can delete it.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
cmjiang: I suggest we still add the if-statement in findWithPrior anyway for the safety of the code. After all,
findWithPrior(n, Seq(), Seq()) is a legal call but it threw an error in the previous code.

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
cmjiang: Will delete it, thanks.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
cmjiang: And, yes, we can just change `find` back to:
`def find(n: Int): Seq[T] = findWithPriorObservations(n, Seq())`

joshvfleming: @cmjiang Let's move the `if` statement to the `find` function, and add an invariant check to `findWithPrior`, e.g.:

```scala
  require(observations.nonEmpty, "There must be at least one observation")
```

joshvfleming: Actually, thinking about it more, we don't need an `if` statement. The `find` function could have the implementation that assumes no prior information, and `findWithPrior` would assume prior information as an input invariant, as before.

joshvfleming: @cmjiang Let's punt on this and go back to the way it was before (but with the `require` invariant check -- that was the actual bug here). I think we need to re-think how this is organized a little to make it "right".

basukinjal: +1 on ^

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -46,11 +53,16 @@ class GameEstimatorEvaluationFunction(
...
+   * @param candidate The candidate vector of hyperparameter with values in [0, 1]
    * @return A tuple of the evaluated value and the original output from the inner estimator
    */
-  override def apply(hyperParameters: DenseVector[Double]): (Double, GameResult) = {
-    val newConfiguration = vectorToConfiguration(hyperParameters)
+  override def apply(candidate: DenseVector[Double]): (Double, GameResult) = {
+
+    // Inverse-scale candidate
+    candidate *= (range.end - range.start)
+    candidate += range.start
```
joshvfleming: Does this mutate the candidate vector in-place? If so, we should probably make a copy first, so that the caller doesn't see any side-effects of this call.

cmjiang:
@joshvfleming Made copies of candidates in scaling functions and move all those functions into photon-client for GAME model.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
```diff
@@ -0,0 +1,127 @@
...
+  def transformForward(
+      vector: DenseVector[Double],
+      transformMap: Map[Int, String]): DenseVector[Double] = {
+
+    val vectorTransformed = vector.copy
+
+    transformMap.foreach { case (index, transform) =>
+      transform match {
+        case "LOG" => vectorTransformed(index) = Math.log10(vectorTransformed(index))
+        case "SQRT" => vectorTransformed(index) = Math.sqrt(vectorTransformed(index))
```
joshvfleming: Let's extract `"LOG"` and `"SQRT"` out to constants.

Also, we need a catch-all case that throws a user-friendly error message in case an unsupported transformation is passed in, e.g.:

```scala
    case other => throw IllegalArgumentException(s"Unknown transformation: $other")
```

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
```diff
@@ -0,0 +1,127 @@
...
+  def transformBackward(
+      vector: DenseVector[Double],
+      transformMap: Map[Int, String]): DenseVector[Double] = {
+
+    val vectorTransformed = vector.copy
+
+    transformMap.foreach { case (index, transform) =>
+      transform match {
+        case "LOG" => vectorTransformed(index) = Math.pow(10, vectorTransformed(index))
+        case "SQRT" => vectorTransformed(index) = Math.pow(vectorTransformed(index), 2)
```
joshvfleming: Same as above.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
```diff
@@ -0,0 +1,127 @@
...
+      discreteIndexSet: Set[Int] = Set()): DenseVector[Double] = {
+
+    val vectorScaled = vector.copy
+
+    val start = DenseVector(ranges.map(_.start).toArray)
+    val end = DenseVector(ranges.map(_.end).toArray)
+    val discreteAdj = DenseVector.tabulate(ranges.length) { i =>
+      if (discreteIndexSet.contains(i)) 1.0 else 0.0
+    }
+    vectorScaled :-= start
```
joshvfleming: Minor: this is a subjective thing, but I feel there should be an empty line above `vectorScaled :-= start`, to create more of a visual separation between this and the declarations above.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
```diff
@@ -0,0 +1,127 @@
...
+      discreteIndexSet: Set[Int] = Set()): DenseVector[Double] = {
+
+    val vectorScaled = vector.copy
+
+    val start = DenseVector(ranges.map(_.start).toArray)
+    val end = DenseVector(ranges.map(_.end).toArray)
+    val discreteAdj = DenseVector.tabulate(ranges.length) { i =>
+      if (discreteIndexSet.contains(i)) 1.0 else 0.0
+    }
+    vectorScaled :*= (end - start + discreteAdj)
```
joshvfleming: Same as above.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
```diff
@@ -0,0 +1,127 @@
...
+package com.linkedin.photon.ml.hyperparameter
+
+import breeze.linalg.DenseVector
+
+import com.linkedin.photon.ml.util.DoubleRange
+
+/**
+ * Functions to transform/scale forward/backward hyper-parameters in vectors
+ */
+object VectorRescaling {
```
joshvfleming: Nice! I like how you're separating this out.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/VectorRescaling.scala
```diff
@@ -0,0 +1,127 @@
...
+
+  /**
+   * Apply forward transformation to a subset of elements in a vector.
+   *
+   * @param vector A DenseVector.
+   * @param transformMap A Map with key-value pairs of indices and names of transform functions.
+   * @return The transformed vector.
+   */
+  def transformForward(
```
joshvfleming: @ashelkovnykov I like this forward/backward naming scheme. Maybe we should also adopt it for the projectors?

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -84,28 +79,16 @@ class RandomSearch[T](
         val candidate = next(lastCandidate, lastObservation)
 
         // Discretize values specified as discrete
-        discreteParams.foreach { index =>
-          candidate(index) = round(candidate(index))
-        }
+        val candidateDiscrete = discretizeCandidate(candidate, discreteParams)
```
basukinjal: Minor: May be a better idea to call this candidateWithDiscrete, since candidateDiscrete might give a wrong impression.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -184,16 +153,28 @@ class RandomSearch[T](
...
+   *
+   * @param candidate candidate with values in [0, 1]
+   * @param discreteParams Map that specifies the indices of discrete parameters and their numbers of discrete values
+   * @return candidate with the specified discrete values
+   */
+  protected[search] def discretizeCandidate(
+      candidate: DenseVector[Double],
+      discreteParams: Map[Int, Int]): DenseVector[Double] = {
+
+    val candidateDiscrete = candidate.copy
```
basukinjal: Same here.

cmjiang:
@joshvfleming @basukinjal Thank you for your reviews. Changes have been done.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearch.scala
```diff
@@ -14,35 +14,29 @@
...
+ * @param numParams The dimensionality of the hyper-parameter tuning problem
  * @param evaluationFunction The function that evaluates points in the space to real values
- * @param discreteParams Specifies the indices of parameters that should be treated as discrete values
+ * @param discreteParams Specifies the indices of discrete parameters and their numbers of discrete values
  * @param kernel Specifies the indices and transformation function of hyper-parameters
  * @param seed A random seed
  */
 class RandomSearch[T](
-    ranges: Seq[DoubleRange],
+    numParams: Int,
```
joshvfleming: Need to make sure this is nonnegative, e.g.:

```scala
  require(numParams, "Number of parameters must be nonnegative")
```

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -42,26 +41,26 @@ import com.linkedin.photon.ml.util.DoubleRange
...
+ * @param discreteParams specifies the indices of discrete parameters and their numbers of discrete values
+ * @param kernel specifies the covariance kernel for hyper-parameters
  * @param candidatePoolSize the number of candidate points to draw at each iteration. Larger numbers give more precise
  *   results, but also incur higher computational cost.
  * @param noisyTarget whether to include observation noise in the evaluation function model
  * @param seed the random seed value
  */
 class GaussianProcessSearch[T](
-    ranges: Seq[DoubleRange],
+    numParams: Int = 1,
```
joshvfleming: Why does this default to 1?

cmjiang: Oh yea, it should have been removed.

PR #369, opened 2018-06-07
Fix bug which allowed one column to be read for multiple purposes

ashelkovnykov:
- Multiple columns in `InputColumnNames` could have the same name
- Columns used for grouping of random effects could have the same names as columns in `InputColumnNames`

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
@@ -48,6 +48,13 @@ object GameConverters {
       isResponseRequired: Boolean,
       inputColumnsNames: InputColumnsNames = InputColumnsNames()): RDD[(UniqueSampleId, GameDatum)] = {
 
+    val colNamesSet = inputColumnsNames.getNames
+
+    // Cannot use response, offset, weight, or uid fields as fields for grouping random effects or queries
+    require(
+      idTagSet.intersect(colNamesSet).isEmpty,
+      s"Cannot use required columns (${colNamesSet.mkString(", ")}) for random effect/validation grouping.")
+
     val inputColumnsNamesBroadcast = data.sqlContext.sparkContext.broadcast(inputColumnsNames)
 
     data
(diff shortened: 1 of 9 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/data/GameConverters.scala
```diff
@@ -48,6 +48,11 @@ object GameConverters {
       isResponseRequired: Boolean,
       inputColumnsNames: InputColumnsNames = InputColumnsNames()): RDD[(UniqueSampleId, GameDatum)] = {
 
+    // Cannot use response, offset, weight, or uid fields as fields for grouping random effects or queries
+    require(
+      idTagSet.intersect(inputColumnsNames.getNames).isEmpty,
+      "Cannot use required columns for random effect/validation grouping.")
```
joshvfleming: We should list the required columns in the error message.

photon-api/src/main/scala/com/linkedin/photon/ml/data/InputColumnsNames.scala
```diff
@@ -20,21 +20,14 @@ package com.linkedin.photon.ml.data
...
-  /**
-   * Same as Map.updated, but the types are fixed.
-   *
-   * @param key The key to update
-   * @param value The value corresponding to key
-   * @return A new InputColumnsNames instance containing the update
-   */
-  def updated(key: InputColumnsNames.Value, value: String): InputColumnsNames = { array(key.id) = value; this }
+  // Check that there are no duplicate column names
+  require(array.length == columnNames.size, "Each column must have unique a name.")
```
joshvfleming: Why not accept `Set` instead of `Array`?

ashelkovnykov: You mean a `Map`, correct? Right now, `InputColumnNames` uses an `Array` and uses the index of the `Enum` as the key to access the column name value. We convert to a `Set` to check that the keySet is unique, and also as a convenience for the safety check on which you have commented above.

joshvfleming: Yeah, looks like I missed that we were using the array indices here.

PR #372, opened 2018-06-15
Add hyper-parameter helper object

cmjiang:
@basukinjal @joshvfleming @ashelkovnykov Can I ask for merging these backward-incompatible changes as well before upgrading the major version?


photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterHelper.scala
```diff
@@ -0,0 +1,183 @@
...
+  }
+
+  /**
+   * Helper method to read a file from HDFS as text.
+   *
+   * @param fs The Hadoop file system
+   * @param inputPath The HDFS input path from which to read the file
+   * @return The contents of the file as text
+   */
+  def readFile(fs: FileSystem, inputPath: String): String = {
```
basukinjal: I think it is a good idea to keep these read and write code here. In the future, this will allow us to separate the module out.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterDefaults.scala
```diff
@@ -0,0 +1,48 @@
...
+ * License for the specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.linkedin.photon.ml.hyperparameter
+
+/**
+ * Hyper-parameter default values
+ */
+object HyperparameterDefaults {
```
ashelkovnykov: This whole class seems very problem specific.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterConfigurationType.scala
```diff
@@ -0,0 +1,27 @@
...
+import com.linkedin.photon.ml.HyperparameterTuningMode.HyperparameterTuningMode
+import com.linkedin.photon.ml.util.DoubleRange
+
+/**
+ * Configurations of hyper-parameters, containing (tuningMode, names, ranges, discreteParams, transformMap).
+ * The Map discreteParams has (index, number of discrete values) as (key, value) pairs. The Map transformMap
+ * has (index, transformation function name) as (key, value) pairs.
+ */
+object HyperparameterConfigurationType {
+  type HyperparameterConfig = (HyperparameterTuningMode, Seq[String], Seq[DoubleRange], Map[Int, Int], Map[Int, String])
```
ashelkovnykov: You should make this a case class instead - it'll be easier to pass around and reference the data inside, instead of creating a type short had for a 5 field tuple.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterDefaults.scala
cmjiang: Yes, this default is for GAME. We are trying to figure out a general config for GAME. If the problem has different numbers of regularizer, we can set others to be zero.

cmjiang:
@basukinjal @ashelkovnykov Your suggestions have been applied. Thank you.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterDefaults.scala
basukinjal: Lets call this object GameHyperparameterDefaults to be more explicit.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterHelper.scala
cmjiang: I am wondering if `readFile` and `writeJson` can be replaced by `IOUtils.readStringsFromHDFS` and `IOUtils.writeStringsToHDFS`. If so, why don't we use them in xgboost-ml?

basukinjal: Let's do that change. I didn't know about this file while that initial version was created. Let's reuse existing files as much as we can.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/HyperparameterHelper.scala
```diff
@@ -0,0 +1,137 @@
...
+import breeze.linalg.DenseVector
+
+import com.linkedin.photon.ml.HyperparameterTuningMode
+import com.linkedin.photon.ml.util.DoubleRange
+
+
+/**
+ * Helper class to read prior observations and configurations of hyper-parameters, and output the results.
+ */
+object HyperparameterHelper {
```
joshvfleming: This should have a more specific name. It looks like this module is in charge of deserializing hyperparameter evaluations and configuration, so something like `HyperparameterSerialization` might work.

PR #373, opened 2018-06-15
Add new threshold parameter for training random effects

ashelkovnykov:
- Add a lower bound for active data; any entity which does not have more data samples than the threshold will not have a random effect model trained
- Added option to ignore the threshold during warm-start when training models for random effect IDs that do not have an existing model
- Added/modified unit tests for new parameter
- Added integration test for new functionality in RandomEffectDataSet


ashelkovnykov:
One more commit likely coming soon for this PR

PR #377, opened 2018-07-13
Decouple hyper-parameter autotuning from evaluator

cmjiang:
Hi @joshvfleming @basukinjal, 
I didn't let the `GaussianProcessSearch` do minimization and let the `EvaluationFunction` handle flipping the sign based on two reasons:
1. `GaussianProcessSearch` will know that the problem is a min or max anyway because it will select the best eval of the prior data from past data sets, which all contain positive evals.
2. `findWithPriors` needs to know that the problem is a min or max to decide whether to flip the result for the output.
So I just keep the `EvaluationFunction` unchanged. Do you think it is a good way to implement? Thanks.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
@@ -649,13 +649,21 @@ object GameTrainingDriver extends GameDriver {
         val iteration = getOrDefault(hyperParameterTuningIter)
         val dimension = baseConfig.toSeq.length
         val mode = getOrDefault(hyperParameterTuning)
-        val evaluationFunction = new GameEstimatorEvaluationFunction(estimator, baseConfig, trainingData, testData)
+
         val evaluator = evaluationResults.get.head._1
+        val isOptMax = evaluator.betterThan(1.0, 0.0)
+        val evaluationFunction = new GameEstimatorEvaluationFunction(
+          estimator,
+          baseConfig,
+          trainingData,
+          testData,
+          isOptMax)
+
         val observations = evaluationFunction.convertObservations(models)
 
         val hyperparameterTuner = HyperparameterTunerFactory[GameEstimator.GameResult](getOrDefault(hyperParameterTunerName))
 
-        hyperparameterTuner.search(iteration, dimension, mode, evaluationFunction, evaluator, observations)
+        hyperparameterTuner.search(iteration, dimension, mode, evaluationFunction, observations)
 
       case _ => Seq()
     }
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
@@ -35,12 +35,14 @@ import com.linkedin.photon.ml.util.DoubleRange
  * @param baseConfig The initial configuration supplied by the user
  * @param data Training data
  * @param validationData Validation data
+ * @param isOptMax a Boolean indicates that the problem is a maximization (true) or minimization (false).
  */
 class GameEstimatorEvaluationFunction(
     estimator: GameEstimator,
     baseConfig: GameOptimizationConfiguration,
     data: DataFrame,
-    validationData: DataFrame)
+    validationData: DataFrame,
+    isOptMax: Boolean)
   extends EvaluationFunction[GameResult] {
 
   import GameEstimatorEvaluationFunction._
@@ -90,8 +92,11 @@ class GameEstimatorEvaluationFunction(
     val model = estimator.fit(data, Some(validationData), Seq(newConfiguration)).head
     val (_, Some(evaluations), _) = model
 
+    // If this is a maximization problem, flip signs of evaluation values
+    val direction = if (isOptMax) -1 else 1
+
     // Assumes model selection evaluator is in "head" position
-    (evaluations.head._2, model)
+    (direction * evaluations.head._2, model)
   }
 
   /**
@@ -105,7 +110,11 @@ class GameEstimatorEvaluationFunction(
     observations.map { observation =>
       val candidate = vectorizeParams(observation)
       val candidateScaled = VectorRescaling.scaleForward(candidate, ranges)
-      val value = getEvaluationValue(observation)
+
+      // If this is a maximization problem, flip signs of evaluation values
+      val direction = if (isOptMax) -1 else 1
+
+      val value = direction * getEvaluationValue(observation)
 
       (candidateScaled, value)
     }
--- a/photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
+++ b/photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
@@ -17,7 +17,6 @@ package com.linkedin.photon.ml.hyperparameter.search
 import breeze.linalg.{DenseMatrix, DenseVector}
 import breeze.stats.mean
 
-import com.linkedin.photon.ml.evaluation.Evaluator
 import com.linkedin.photon.ml.hyperparameter.EvaluationFunction
 import com.linkedin.photon.ml.hyperparameter.criteria.ExpectedImprovement
 import com.linkedin.photon.ml.hyperparameter.estimators.{GaussianProcessEstimator, GaussianProcessModel, PredictionTransformation}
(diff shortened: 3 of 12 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
```diff
@@ -64,12 +63,14 @@ class GaussianProcessSearch[T](
 
   private var observedPoints: Option[DenseMatrix[Double]] = None
   private var observedEvals: Option[DenseVector[Double]] = None
-  private var bestEval: Double = evaluator.defaultScore
+  private var bestEval: Double = 0.0
```
ashelkovnykov: Won't this be an issue if it's a minimization problem?

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
```diff
@@ -204,14 +205,5 @@ object GaussianProcessSearchTest {
...
-    override protected[ml] val labelAndOffsetAndWeights: RDD[(Long, (Double, Double, Double))] =
-      mock(classOf[RDD[(Long, (Double, Double, Double))]])
-    override val evaluatorType: EvaluatorType = EvaluatorType.AUC
-    override protected[ml] def evaluateWithScoresAndLabelsAndWeights(
-        scoresAndLabelsAndWeights: RDD[(Long, (Double, Double, Double))]): Double = 0.0
-
-    override def betterThan(score1: Double, score2: Double): Boolean = score1 > score2
-  }
+  val IS_OPT_MAX = true
 }
```
ashelkovnykov: While you're editing this file, please add a newline at end of file

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
```diff
@@ -204,14 +205,5 @@ object GaussianProcessSearchTest {
...
-
-    override protected[ml] val labelAndOffsetAndWeights: RDD[(Long, (Double, Double, Double))] =
-      mock(classOf[RDD[(Long, (Double, Double, Double))]])
-    override val evaluatorType: EvaluatorType = EvaluatorType.AUC
-    override protected[ml] def evaluateWithScoresAndLabelsAndWeights(
-        scoresAndLabelsAndWeights: RDD[(Long, (Double, Double, Double))]): Double = 0.0
-
-    override def betterThan(score1: Double, score2: Double): Boolean = score1 > score2
-  }
+  val IS_OPT_MAX = true
```
ashelkovnykov: Should have unit tests where the problem is a minimization problem as well

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/RandomSearchTest.scala
```diff
@@ -115,9 +117,9 @@ class RandomSearchTest {
 object RandomSearchTest {
 
   val SEED = 1L
-  val DIM = 5
-  val N = 25
-  val DISCRETE_PARAMS = Map(0 -> 5, 1 -> 5, 2 -> 5, 4 -> 9)
+  val DIM = 3
+  val N = 5
+  val DISCRETE_PARAMS = Map(0 -> 5, 2 -> 9)
```
ashelkovnykov: Thanks for this speed-up!

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
cmjiang: I don't think so. (1) The evaluation value is demeaned. (2) Looks like over all the evaluators the defaults are always 0.

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
cmjiang: Will do. Thanks.

photon-lib/src/test/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearchTest.scala
cmjiang: Will add it. Thanks.

joshvfleming:
@cmjiang I think it would be better / simpler to make the change in `EvaluationFunction` instead of having the system be aware of minimization vs maximization in all these places. Let's chat about it when I'm there in person next week.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -639,7 +638,7 @@ object GameTrainingDriver extends GameDriver {
 
         val searcher = getOrDefault(hyperParameterTuning) match {
           case HyperparameterTuningMode.BAYESIAN =>
-            new GaussianProcessSearch[GameEstimator.GameResult](dimension, evaluationFunction, evaluator)
+            new GaussianProcessSearch[GameEstimator.GameResult](dimension, evaluationFunction, isOptMax)
```
joshvfleming: @cmjiang I think it would be better / simpler to make the change in EvaluationFunction instead of having the system be aware of minimization vs maximization in all these places. Let's chat about it when I'm there in person next week.

cmjiang:
@joshvfleming Re-open this PR for decoupling autotune from photon since the convergence bug was found. The kernel is changed to minimization problem only. Evaluation function will handle the choice of min or max optimization. The changes of XGBoost evaluation function will posted correspondingly.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
basukinjal: Just to make sure, lets test this out with an AUC and RMSE test function.

photon-client/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimatorEvaluationFunction.scala
```diff
@@ -105,7 +110,11 @@ class GameEstimatorEvaluationFunction(
     observations.map { observation =>
       val candidate = vectorizeParams(observation)
       val candidateScaled = VectorRescaling.scaleForward(candidate, ranges)
-      val value = getEvaluationValue(observation)
+
+      // If this is a maximization problem, flip signs of evaluation values
+      val direction = if (isOptMax) -1 else 1
+
+      val value = direction * getEvaluationValue(observation)
```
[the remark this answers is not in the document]
cmjiang: Yes, `convertObservations` is only for the prior data.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
cmjiang: @joshvfleming Will it cause a numerical overflow?

cmjiang:
Rebased this PR on the latest version.

basukinjal:
@cmjiang This looks great now. Just to make sure everything is working as expected, please test the overall flow with the Normal PDF / Shekel function using both EI and UCB.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
joshvfleming: @cmjiang Not if we're using it only for comparison.

cmjiang: @joshvfleming I am wondering if we do a calculation with `bestEval` at line 91 `val overallBestEval = Math.min(priorBestEval, bestEval - currentMean)` and `currentMean` is negative, then the result is not expected.

cmjiang:
@basukinjal Tests are done with hyper-parameter simulation examples and results are good as expected.

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/search/GaussianProcessSearch.scala
cmjiang: @joshvfleming Never mind. The `bestEval` will be updated by line 140 `bestEval = Math.min(bestEval, eval)` before executing line 91 `val overallBestEval = Math.min(priorBestEval, bestEval - currentMean)`.

joshvfleming:
LGTM. Will merge once the tests finish.

PR #379, opened 2018-07-13
Added support for constraint maps in TRON optimizer

### LINKED ISSUE #378: TRON Optimizer Does Not Support Coefficient Constraints
I am working with @wjohnson on an implementation of Photon ML at one of our customers.  They are very interested in the coefficient constraints feature, which they view as a major differentiator between Photon ML and other solutions.

However, the customer also wants to use the TRON optimizer because it more closely replicates their existing solution.

Photon ML does not currently support coefficient constraints for the TRON optimizer.  This feature is only supported in the LBFGS family.

(Pull request forthcoming)

AnalyticJeremy:
Added in very crude support for coefficient constraint maps in the TRON optimizer

This resolves #378

### DIFF
--- a/photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/TRON.scala
+++ b/photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/TRON.scala
@@ -249,6 +249,22 @@ class TRON(
 
     objectiveFunction.cleanupCoefficients(convertedPrevCoefficients)
 
+    // If any coefficients exceed the bounds imposed by the constraintMap, knock them back into range
+    if (constraintMap.isEmpty == false) {
+      constraintMap.foreach { map => 
+        map.keys.foreach { key =>
+          val range = map(key)
+          val currentCoef = finalState.coefficients(key)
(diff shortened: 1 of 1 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/TRON.scala
```diff
@@ -249,6 +249,22 @@ class TRON(
...
+          val currentCoef = finalState.coefficients(key)
+          if (currentCoef < range._1) {
+            finalState.coefficients(key) = range._1
+          }
+          else if (currentCoef > range._2) {
+            finalState.coefficients(key) = range._2
+          }
+        }
+      }
+    }
```
ashelkovnykov: I think what you would actually want to do is drop these changes and replace line 228 with:
```
OptimizationUtils.projectCoefficientsToSubspace(coefficients, constraintMap),
```

AnalyticJeremy:
Thanks for the help, @ashelkovnykov.  That's way easier than the way I was doing it.  I've updated and tested the code.

PR #381, opened 2018-07-17
shrink search range functionality added

yunboouyang:
This is a helper function to implement autotune with sub-sampling for machine learning models. This function takes hyperparameters and evaluation values, fits a Gaussian process regression model and outputs a small range which centers at the optimal value. It will be a part of atlas-ml hyperparameter tuning library.

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
@@ -1,0 +1,133 @@
+package com.linkedin.photon.ml.hyperparameter
+
+import scala.math.{floor, max, min}
+
+import breeze.linalg.{DenseMatrix, DenseVector}
+import org.apache.commons.math3.random.SobolSequenceGenerator
+
+import com.linkedin.photon.ml.hyperparameter.estimators.GaussianProcessEstimator
+import com.linkedin.photon.ml.hyperparameter.estimators.kernels.Matern52
+
+/**
+  * An object to shrink search range given prior data.
+  */
+object ShrinkSearchRange {
+
+  /**
+    * Compute the lower bound and upper bound of the new prior config
+    *
+    * @param hyperParams Configurations of hyper-parameters
+    * @param priorJsonString JSON string containing prior observations
+    * @param priorDefault Default values for missing hyper-parameters
+    * @param radius The radius of search range after transformation to [0, 1]
+    * @return A tuple of lower bounds and upper bounds for each hyperparameter. Lower bounds and upper bounds are
+    *         discrete for discrete parameters. Lower bounds and upper bounds are dense vectors.
+    */
+  def getBounds(
+      hyperParams: HyperparameterConfig,
+      priorJsonString: String,
+      priorDefault: Map[String, String],
+      radius: Double,
+      candidatePoolSize: Int = 1000,
+      seed: Long = System.currentTimeMillis): (DenseVector[Double], DenseVector[Double]) = {
+
+    val hyperparameterList = hyperParams.names
+    val ranges = hyperParams.ranges
+    val discreteParams = hyperParams.discreteParams
+    val numParams = ranges.length
+
+    // Get a [[Seq]] of (vectorized hyper-parameters, evaluationValue) tuples
+    val hyperparameterPairs = HyperparameterSerialization.priorFromJson(priorJsonString, priorDefault, hyperparameterList)
+
+    // Rescale the hyperparameters to [0,1]
+    val hyperparameterRescaled = VectorRescaling.rescalePriors(hyperparameterPairs, hyperParams)
+
+    // Combine hyperparameters as a dense matrix and evaluation value as a dense vector
+    val (overallPoints, overallEvals) = hyperparameterRescaled.map(x => (x._1.asDenseMatrix, DenseVector(x._2))).reduce(
+      (a, b) => (DenseMatrix.vertcat(a._1, b._1), DenseVector.vertcat(a._2, b._2))
+    )
+
+    // Fit Gaussian process regression model
+    val estimator = new GaussianProcessEstimator(kernel = new Matern52)
+    val model = estimator.fit(overallPoints, overallEvals)
+
+    // Sobol generator
+    val paramDistributions = {
+      val sobol = new SobolSequenceGenerator(numParams)
+      sobol.skipTo((seed % (Int.MaxValue.toLong + 1)).toInt)
+      sobol
+    }
+
+    // Draw candidates from a Sobol generator
+    val candidates = (1 until candidatePoolSize).foldLeft(DenseMatrix(paramDistributions.nextVector)) { case (acc, _) =>
+      DenseMatrix.vertcat(acc, DenseMatrix(paramDistributions.nextVector))
+    }
+
+    // Select the best candidate
+    val predictions = model.predict(candidates)
+    val bestCandidate = selectBestCandidate(candidates, predictions._1)
+
+    // compute lower bound and upper bound
+    val upperBound = VectorRescaling.scaleBackward(
+      discretizeCandidate(bestCandidate.map(x => x + radius), discreteParams),
+      ranges,
+      discreteParams.keySet)
+    val lowerBound = VectorRescaling.scaleBackward(
+      discretizeCandidate(bestCandidate.map(x => x - radius), discreteParams),
+      ranges,
+      discreteParams.keySet)
+
+    (0 until numParams).foreach{
+      index =>
+        upperBound(index) = min(upperBound(index), ranges(index).end)
+        lowerBound(index) = max(lowerBound(index), ranges(index).start)
+    }
+
+    (lowerBound, upperBound)
+  }
+
+  /**
+    * Selects the best candidate according to the predicted values, where "best" is defined as the largest
+    *
+    * @param candidates matrix of candidates
+    * @param predictions predicted values for each candidate
+    * @return the candidate with the best value
+    */
+  private def selectBestCandidate(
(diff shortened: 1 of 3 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,126 @@
...
+
+    // Sobol generator
+    val paramDistributions = {
+      val sobol = new SobolSequenceGenerator(numParams)
+      sobol.skipTo((seed % (Int.MaxValue.toLong + 1)).toInt)
+      sobol
+    }
+
+    // Draw candidates from a Sobol generator
+    val candidates = (1 until 1000).foldLeft(DenseMatrix(paramDistributions.nextVector)) { case (acc, _) =>
```
basukinjal: It might be a good idea to set this 1000 as a parameter with a default being 1000.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,126 @@
...
+    val upperBound = VectorRescaling.scaleBackward(
+      discretizeCandidate(bestCandidate.map(x => min(x + radius, 1.0)), discreteParams),
+      ranges,
+      discreteParams.keySet)
+    val lowerBound = VectorRescaling.scaleBackward(
+      discretizeCandidate(bestCandidate.map(x => max(x - radius, 0.0)), discreteParams),
+      ranges,
+      discreteParams.keySet)
+
+    discreteParams.foreach{
```
basukinjal: Should we always +- 1? I guess it might be a better idea to work with the radius on this domain as well. In the case we get only one integer then we expand that a bit. For example shrinking [4, 10] to [0,1] with 0.01 radius will usually give a single integer where we can extend. But [1, 10000] to [0,1] with 0.01 will contain more than 1 integer.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,126 @@
...
+
+    val candidateWithDiscrete = candidate.copy
+
+    discreteParams.foreach { case (index, numDiscreteValues) =>
+      candidateWithDiscrete(index) = floor(candidate(index) * numDiscreteValues) / numDiscreteValues
+    }
+
+    candidateWithDiscrete
+  }
+}
```
basukinjal: Please add a new line at the end.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,126 @@
...
+  /**
+    * Selects the best candidate according to the predicted values, where "best" is defined as the largest
+    *
+    * @param candidates matrix of candidates
+    * @param predictions predicted values for each candidate
+    * @return the candidate with the best value
+    */
+  private def selectBestCandidate(candidates: DenseMatrix[Double],
+                                  predictions: DenseVector[Double]): DenseVector[Double] = {
+
```
cmjiang: Is there a reason to copy and paste this function from `GaussianProcessSearch`?

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,126 @@
...
+  /**
+    * Discretize candidates with specified indices.
+    *
+    * @param candidate candidate with values in [0, 1]
+    * @param discreteParams Map that specifies the indices of discrete parameters and their numbers of discrete values
+    * @return candidate with the specified discrete values
+    */
+  private def discretizeCandidate(candidate: DenseVector[Double],
+                                  discreteParams: Map[Int, Int]): DenseVector[Double] = {
```
cmjiang: Same as above.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
yunboouyang: Right now in photon-ml case class DoubleRange has the requirement that start is strictly less than end, therefore I manually expand the range. Should we change the case class requirements to be start <= end?

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
yunboouyang: selectBestCandidate uses betterThan method for evaluator, which we cannot use without specifying the evaluator. This function will be re-written in atlas-ml so that we can directly call it.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
yunboouyang: discretizeCandidate is inside RandomSearch class. To initialize an instance of the class, we need to specify the evaluation function of the type T related with different models. I think this function should be a utility function so that it works for any models universally. We are in the code migration process. These functions will be in the atlas-ml eventually so that it can be directly called.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,128 @@
...
+    * @param radius The radius of search range after transformation to [0, 1]
+    * @return A tuple of lower bounds and upper bounds for each hyperparameter. Lower bounds and upper bounds are
+    *         discrete for discrete parameters. Lower bounds and upper bounds are dense vectors.
+    */
+  def getBounds(hyperParams: HyperparameterConfig,
+                priorJsonString: String,
+                priorDefault: Map[String, String],
+                radius: Double,
+                candidatePoolSize: Int = 1000,
+                seed: Long = System.currentTimeMillis): (DenseVector[Double], DenseVector[Double]) = {
```
joshvfleming: Each argument should get its own line, indented 4 spaces.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,128 @@
...
+
+  /**
+    * Selects the best candidate according to the predicted values, where "best" is defined as the largest
+    *
+    * @param candidates matrix of candidates
+    * @param predictions predicted values for each candidate
+    * @return the candidate with the best value
+    */
+  private def selectBestCandidate(candidates: DenseMatrix[Double],
+                                  predictions: DenseVector[Double]): DenseVector[Double] = {
```
joshvfleming: Same as above.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,128 @@
...
+  /**
+    * Discretize candidates with specified indices.
+    *
+    * @param candidate candidate with values in [0, 1]
+    * @param discreteParams Map that specifies the indices of discrete parameters and their numbers of discrete values
+    * @return candidate with the specified discrete values
+    */
+  private def discretizeCandidate(candidate: DenseVector[Double],
+                                  discreteParams: Map[Int, Int]): DenseVector[Double] = {
+
```
joshvfleming: Same as above.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,128 @@
+package com.linkedin.photon.ml.hyperparameter
+import breeze.linalg.{DenseMatrix, DenseVector}
+import com.linkedin.photon.ml.hyperparameter.estimators.GaussianProcessEstimator
+import com.linkedin.photon.ml.hyperparameter.estimators.kernels.Matern52
+import org.apache.commons.math3.random.SobolSequenceGenerator
+
+import scala.math.{floor, min, max}
```
joshvfleming: Imports should be grouped as below, separated by newline:

1. Langage runtime packages
2. 3rd-party packages
3. Current project packages

E.g., for the above, it would be:

```scala
import scala.math.{floor, min, max}

import breeze.linalg.{DenseMatrix, DenseVector}
import org.apache.commons.math3.random.SobolSequenceGenerator

import com.linkedin.photon.ml.hyperparameter.estimators.GaussianProcessEstimator
import com.linkedin.photon.ml.hyperparameter.estimators.kernels.Matern52
```

cmjiang: Just to be a little picky, please sort it by alphabetic order. For example,
`import scala.math.{floor, max, min}`
:)

joshvfleming:
@yunboouyang LGTM. Please squash this down to one commit.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/ShrinkSearchRange.scala
```diff
@@ -0,0 +1,133 @@
+package com.linkedin.photon.ml.hyperparameter
```
cmjiang: Missing the documentation of license for this file.

PR #383, opened 2018-07-30
Fix the bug to compute EI

yunboouyang:
Add parentheses in the formula to compute EI. In breeze vector element-wise product operator :* has the lower operator priority than the elementwise addition +. Parentheses are needed.


photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/criteria/ExpectedImprovement.scala
```diff
@@ -17,7 +17,6 @@ package com.linkedin.photon.ml.hyperparameter.criteria
 import breeze.linalg.DenseVector
 import breeze.numerics.sqrt
 import breeze.stats.distributions.Gaussian
-
```
cmjiang: This blank line should be kept.

joshvfleming:
I can't get the tests to pass on Travis, but they pass locally so I'm going to merge this, since it's urgent.

PR #386, opened 2018-08-07
Add hyper-parameter tuner interface

cmjiang:
@joshvfleming @basukinjal @yunboouyang
This PR, adding a hyper-parameter tuner, has no conflicts with PR#385. After this PR and PR#385 are both merged, let's merge PR#377 rebased on them to completely decouple the auto-tuning system.


photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.spark.sql.DataFrame
+
+import com.linkedin.photon.ml.HyperparameterTuningMode
+import com.linkedin.photon.ml.cli.game.training.GameTrainingDriver.{getOrDefault, hyperParameterTuning, hyperParameterTuningIter}
+import com.linkedin.photon.ml.estimators.{GameEstimator, GameEstimatorEvaluationFunction}
+import com.linkedin.photon.ml.estimators.GameEstimator.GameResult
+import com.linkedin.photon.ml.hyperparameter.search.{GaussianProcessSearch, RandomSearch}
+
+/**
+ * A hyper-parameter tuner which depends on an internal library.
```
basukinjal: Let make it more explicit and say internal LinkedIn library.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
```diff
@@ -0,0 +1,72 @@
+/*
```
basukinjal: Bit unclear on this part. So the AtlasTuner will move out of photon-ml correct? and photon-ml will only contain the DummyTuner?

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
```diff
@@ -0,0 +1,72 @@
...
+ */
+package com.linkedin.photon.ml.hyperparameter.tuner
+
+import org.apache.spark.sql.DataFrame
+
+import com.linkedin.photon.ml.HyperparameterTuningMode
+import com.linkedin.photon.ml.cli.game.training.GameTrainingDriver.{getOrDefault, hyperParameterTuning, hyperParameterTuningIter}
+import com.linkedin.photon.ml.estimators.{GameEstimator, GameEstimatorEvaluationFunction}
+import com.linkedin.photon.ml.estimators.GameEstimator.GameResult
+import com.linkedin.photon.ml.hyperparameter.search.{GaussianProcessSearch, RandomSearch}
```
basukinjal: Related to the above comment, in that setup, we will move out all functions from hyperparameter correct?

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/HyperparameterTunerName.scala
```diff
@@ -0,0 +1,23 @@
+/*
+ * Copyright 2017 LinkedIn Corp. All rights reserved.
```
basukinjal: Make this 2018

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
cmjiang: I think both `DummyTuner` and `AtlasTuner` will stay in photon-ml and li-photon-ml because we want to keep the code exactly the same. And we will use `HyperparameterTunerName` to choose which one to be called.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
cmjiang: This tuner class implementing `runHyperparameterTuning` will be unique for different models, e.g. GAME and XGBoost since they have their own customized `runHyperparameterTuning`. It is like an integration point which I prefer to keep it in photon-client and xgboost. We are not moving all the functions from `hyperparameter` folder out. For example, `GameHyperparameterDefaults` will stay too.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/HyperparameterTunerName.scala
cmjiang: Will change it. Thanks.

photon-client/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
cmjiang: Will change it. Thank you!

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
```diff
@@ -0,0 +1,63 @@
+/*
```
basukinjal: So this will now move out of photon-ml into atlas-ml by depending on photon-ml by the trait correct?

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
```diff
@@ -0,0 +1,63 @@
...
+package com.linkedin.photon.ml.hyperparameter.tuner
+
+import breeze.linalg.DenseVector
+
+import com.linkedin.photon.ml.HyperparameterTuningMode
+import com.linkedin.photon.ml.HyperparameterTuningMode.HyperparameterTuningMode
+import com.linkedin.photon.ml.evaluation.Evaluator
+import com.linkedin.photon.ml.hyperparameter.EvaluationFunction
+import com.linkedin.photon.ml.hyperparameter.search.{GaussianProcessSearch, RandomSearch}
+
```
basukinjal: Just to confirm, the lines 19-22 will stay in photon-ml as traits and 23 will move to atlas-ml. Correct?

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
cmjiang: Yes

photon-lib/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/AtlasTuner.scala
cmjiang: Yes. And the evaluator will be removed later.

photon-api/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/HyperparameterTunerFactory.scala
```diff
@@ -0,0 +1,33 @@
...
+  /**
+   * Factory for different packages of [[HyperparameterTuner]].
+   *
+   * @param tunerName The name of the package
+   * @return The hyper-parameter tuner
+   */
+  def apply[T](tunerName: HyperparameterTunerName): HyperparameterTuner[T] = {
+    tunerName match {
+      case DUMMY => new DummyTuner[T]
+      case ATLAS => new AtlasTuner[T]
```
cmjiang: There is still a dependency issue here...

basukinjal:
Let change as per the offline discussion.

photon-api/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/HyperparameterTunerFactory.scala
```diff
@@ -0,0 +1,47 @@
...
+ * under the License.
+ */
+package com.linkedin.photon.ml.hyperparameter.tuner
+
+import com.linkedin.photon.ml.HyperparameterTunerName.{ATLAS, DUMMY, HyperparameterTunerName}
+
+object HyperparameterTunerFactory {
+
+  val DUMMY_TUNER = "com.linkedin.photon.ml.hyperparameter.tuner.DummyTuner"
+  val ATLAS_TUNER = "com.linkedin.photon.ml.hyperparameter.tuner.AtlasTuner"
```
basukinjal: Are both of these moving to Atlas? In either case, let's change the string to " com.linkedin.atlas.tuner.AtlasTuner"

cmjiang: Since it is not moved yet, the test with com.linkedin.atlas.tuner.AtlasTuner will report an error message. Please see the test result with invalid tuner in li-photon-plugin.

joshvfleming: Let's put a comment here with a brief explanation of why it's arranged this way, and that the atlas-specific code will be removed.

photon-api/src/main/scala/com/linkedin/photon/ml/hyperparameter/tuner/HyperparameterTunerFactory.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object HyperparameterTunerFactory {
+
+  // Use DUMMY_TUNER for photon-ml, which does an empty operation for hyper-parameter tuning
+  val DUMMY_TUNER = "com.linkedin.photon.ml.hyperparameter.tuner.DummyTuner"
+
+  // TODO: Move AtlasTuner into atlas-ml for the auto-tuning system migration:
+  // TODO: val ATLAS_TUNER = "com.linkedin.atlas.ml.hyperparameter.tuner.AtlasTuner".
+  // TODO: Temporarily stay in photon-ml for test purpose.
+  val ATLAS_TUNER = "com.linkedin.photon.ml.hyperparameter.tuner.AtlasTuner"
```
cmjiang: Comments added here @basukinjal @joshvfleming

ashelkovnykov:
I don't like that there is both a tuner name and a tuning mode now - one is a subtype of the other, really. The end goal is that we want to do either no tuning, random tuning, or Bayesian tuning. Now this is split over two inputs, many combinations of which are invalid (`Atlas + NONE`, `Dummy + Bayesian`, etc.).

Why even have a "dummy" tuner? Why not have an `Option` so that  `None` is the null tuner?

joshvfleming:
@ashelkovnykov The idea is that rather than an if/else branch in the driver code, the factory would default to a noop tuner in the case of no tuning. That way, the logic of deciding what to do is contained in the factory instead of the driver.

basukinjal:
@ashelkovnykov @joshvfleming I think this is being done just so that photon-ml does not need to depend on atlas-ml in the short term. Once we make atlas-ml open sourced, this branching can be removed as photon-ml will directly depend on atlas-ml and the driver will work only with None, Random or Bayesian.  We can then maybe remove this tunerFactory.

joshvfleming:
@basukinjal @ashelkovnykov Yeah. The whole point of this is that it's an intermediate step to create a functioning separation between the two codebases. It might be ugly to have the interface in place, but that's better than having the entire atlas codebase embedded here as it is now.

photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
```diff
@@ -363,7 +363,8 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
     val newArgs = mixedEffectSeriousRunArgs
       .copy
       .put(GameTrainingDriver.rootOutputDirectory, outputDir)
-      .put(GameTrainingDriver.outputMode, ModelOutputMode.TUNED)
+      .put(GameTrainingDriver.outputMode, ModelOutputMode.ALL)
```
cmjiang: Here we must change the `outputMode` to `ModelOutputMode.ALL`. If we use `ModelOutputMode.TUNED`, we don't have any tuned model with a dummy tuner. Then it returns an empty output model which causes a failure at line 377 below.

cmjiang: Actually I am wondering if we can remove this `testHyperParameterTuning` in `GameTrainingDriverIntegTest` since we won't have atlas-ml in photon-ml, all the following lines in this test become non-sense for testing.

ashelkovnykov: Probably best to comment it out, since once Atlas is open-sourced the test will be relevant again.

basukinjal: @cmjiang Great find! Let's get the ball rolling on atlas now. This LGTM. @joshvfleming Let's merge!

cmjiang: Commented the test out and put a comment on this change.

PR #389, opened 2018-08-29
Move hyper-parameter config functions into photon-lib

cmjiang:
Move common hyper-parameter config/scaling functions for models such as GAME and XGBoost from photon-client to photon lib. Therefore atlas doesn't need to add photon-client into lib jar.


ashelkovnykov:
I thought the point was that Atlas would have a copy of all of this code and that it would be locked in Photon for a temporary time period.

cmjiang:
@ashelkovnykov Yes, we temporarily keep all the copied files in photon. This PR is for moving files internally in photon.

joshvfleming:
@ashelkovnykov There has to be dependency going at least one way, or there would be no way to use them together. Since we can't have Photon depend on Atlas for now, Atlas has to depend on Photon.

ashelkovnykov:
@joshvfleming I thought the dependency was through reflection - the path to the Atlas tuner is hard-coded into a constant and Photon attempts to instantiate it at runtime. Atlas compiles independently and gets added to the Spark jars configuration.

joshvfleming:
@ashelkovnykov The runtime dependency is through reflection, but they have to share some information so that Photon can make the call into Atlas. Photon needs to know about some function to call to signal that hyperparameter tuning should be run. Atlas also needs to know about this function, so that it can implement the function. They share this information by way of an interface/trait that lives in Photon.

cmjiang:
@joshvfleming @basukinjal Since we let the evaluation function perform the normalization/re-scaling, the implementation of evaluation function for GAME needs to know how to scale those hyper-parameters. The other two classes are for pre-processing the hyper-parameters, such as converting a Json string into config tuples. They are used in XGBoost driver now but not in GAME driver. So they can be moved. But in the future, are we thinking of uniform the hyper-parameter config?

PR #390, opened 2018-09-05
Minor refactor relating to normalization, statistical summaries, and the intercept

ashelkovnykov:
- Refactor `NormalizationContext` construction, style, and documentation
- Include intercept info in the `BasicStatisticalSummary`
- Ignore intercept when writing `BasicStatisticalSummary` to file


YazhiGao:
I think the intercept inclusion in `BasicStatisticalSummary` is good for resolving intercept issues for current feature selection functionality extension.
Might need to look into the CI build fail that results from building on previously built artifact.

photon-lib/src/main/scala/com/linkedin/photon/ml/stat/BasicStatisticalSummary.scala
```diff
@@ -43,7 +43,8 @@ case class BasicStatisticalSummary(
     min: BreezeVector[Double],
     normL1: BreezeVector[Double],
     normL2: BreezeVector[Double],
-    meanAbs: BreezeVector[Double])
+    meanAbs: BreezeVector[Double],
+    interceptIndex: Option[Int])
```
joshvfleming: I think it makes sense to package it up this way, but the name `BasicStatisticalSummary` doesn't really make sense anymore. It's more like `DataSummary` or something more general like that.

ashelkovnykov: Going to take care of this in a separate PR.

PR #399, opened 2018-10-18
Modify AvroDataReader to perform partition count check in all circumstances

ashelkovnykov:
- Fix bug: previously AvroDataReader would only repartition when IndexMapLoaders provided
- Removed integration test assertions that depended on strict ordering of data in DataFrame
- Added integration test to check that explicit repartition is called


ashelkovnykov:
I'm looking into the test failures - for some reason there's a test whose results change based on the order of the data. It's quite confusing.

PR #405, opened 2019-01-09
Remove the lower bound on passive dataset size as a tuning parameter

ashelkovnykov:
The reasoning is that it is a rarely-used tuning parameter which requires rather specific knowledge of the implementation of Photon ML (i.e. the difference between "active" and "passive" data).

Maintaining this tuning parameter increases training time (the code for computing passive datasets is run regardless - but with the tuning parameter, the number of samples per entity now also need to be counted and filtered), reduces accuracy (since records for entities which do not break the bound threshold will not be scored), and introduces additional code complexity (lots of `Option` objects).

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
@@ -166,11 +166,11 @@ protected[ml] class RandomEffectDataset(
   }
 
   /**
-   * Update the data set.
+   * Update the dataset.
    *
    * @param updatedActiveData Updated active data
    * @param updatedPassiveData Updated passive data
-   * @return A new updated data set
+   * @return A new updated dataset
    */
   def update(
       updatedActiveData: RDD[(REId, LocalDataset)],
@@ -396,7 +396,7 @@ object RandomEffectDataset {
   }
 
   /**
-   * Generate passive data set.
+   * Generate passive dataset.
    *
    * @param gameDataset The raw input dataset
    * @param activeData The active dataset
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
@@ -14,7 +14,6 @@
  */
 package com.linkedin.photon.ml.data
 
-import scala.collection.Set
 import scala.util.hashing.byteswap64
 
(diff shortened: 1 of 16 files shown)


photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
```diff
@@ -569,22 +557,14 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
...
-    val model = modelId match {
-      case Some(id) =>
-        val m = modelAvro.find { m => m.getModelId.toString == id }
-        assertTrue(m.isDefined, s"Model id $id not found.")
-        m.get
-      case _ => modelAvro.head
-    }
-
-    assertEquals(model.getMeans.count(x => x.getValue != 0), expectedNumCoefficients)
+    assertEquals(modelAvro.head.getMeans.count(x => x.getValue != 0), expectedNumCoefficients)
```
joshvfleming: Does this work? The model in `head` position might not be the one that the test case expects. This led to some nondeterminism in the past.

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
```diff
@@ -409,35 +388,29 @@ object RandomEffectDataset {
...
-    localDatasets
+        data.map { case ComparableLabeledPointWithId(_, uniqueId, LabeledPoint(label, features, offset, weight)) =>
+          (uniqueId, LabeledPoint(label, features, offset, weightMultiplierOpt.map(_ * weight).getOrElse(weight)))
+        }
+      }
   }
 
   /**
-   * Generate passive dataset.
+   * Generate passive data set.
```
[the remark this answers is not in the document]
ashelkovnykov: Will do

photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
ashelkovnykov: `assertModelSane` is only being called now for fixed effects and random effects that should have a uniform number of features across all models.

joshvfleming: Got it.

PR #406, opened 2019-02-06
Only accept string type in map schema for input data

alice2008:


photon-client/src/main/scala/com/linkedin/photon/ml/util/Utils.scala
```diff
@@ -135,9 +135,9 @@ object Utils {
           // Need to convert Utf8 values to String here, because otherwise we get schema casting errors and misleading
           // equivalence failures downstream.
           case s@(_: Utf8 | _: JString) => s.toString
-          case _ => value
+          case _ => null
         })
-      }.toMap
+      }.filter(_._2 != null).toMap
```
ashelkovnykov: I'm not a fan of this solution - it just silently drops non-`String` values from any loaded `Map`. I'm pretty sure this has the potential to break the 'metadataMap' field, which can be used to record random effect IDs.

Furthermore, I don't see this as a long-term solution. I think we should actively filter fields while loading the `DataFrame`, maybe using an optional whitelist of fields to keep.

alice2008: BTW, what's `metadataMap` is used for in glmix, to boostrap a warm retrain? You mean adding a job parameter to white-list of fields user wants to keep in the training, such as `features` or `metadataMap`?

ashelkovnykov:
@alice2008 Looks fine, as a temporary solution. Please squash the commits and add a descriptive commit message.

alice2008:
okay, done.

PR #415, opened 2019-03-13
Minor refactor to DatumScoringModel modelType

ashelkovnykov:
This change is made in preparation for the projection refactor, after which `RandomEffectModelInProjectedSpace` will no longer exist and the number of calls to `determineModelType` should no longer cause noticeable performance impact (allowing us to simplify the code)

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
@@ -40,11 +40,7 @@ class RandomEffectModel(
   extends DatumScoringModel
   with RDDLike {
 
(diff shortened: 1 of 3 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/model/RandomEffectModel.scala
```diff
@@ -234,19 +213,16 @@ object RandomEffectModel {
    *       that type - it will be faster for large numbers of random effect models. Note that it may still be a
```
cmjiang: Remove the TODO comments.

ashelkovnykov: We spoke offline - I explained the meaning of the TODO comments

PR #419, opened 2019-03-26
Fix bug in calculating RMSE

jianlingzhong:


photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/RMSEEvaluator.scala
```diff
@@ -31,7 +31,9 @@ object RMSEEvaluator extends SingleEvaluator {
    */
   override def evaluate(scoresAndLabelsAndWeights: RDD[(Double, Double, Double)]): Double = {
 
-    val squaredLoss = SquaredLossEvaluator.evaluate(scoresAndLabelsAndWeights)
+    // squared loss is calculated as 1/2 (z - y)^2 for mathematical convenience
+    // however, RMSE should be using (z - y)^2 for calculation, hence the multiplying by 2 here:
+    val squaredLoss = 2 * SquaredLossEvaluator.evaluate(scoresAndLabelsAndWeights)
```
ashelkovnykov: I think the better place for this multiplication is in the `SquaredLossEvaluator`, as that is also a validation evaluator using the training code for convenience. It's also possible that if someone were to enable both the `SquaredLossEvaluator` and the `RMSEEvaluator`, they could manually compare the two values (since one is computed from the other) and see that they don't match.

photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/RMSEEvaluator.scala
```diff
@@ -31,7 +31,9 @@ object RMSEEvaluator extends SingleEvaluator {
    */
   override def evaluate(scoresAndLabelsAndWeights: RDD[(Double, Double, Double)]): Double = {
 
-    val squaredLoss = SquaredLossEvaluator.evaluate(scoresAndLabelsAndWeights)
+    // squared loss is calculated as 1/2 (z - y)^2 for mathematical convenience
+    // however, RMSE should be using (z - y)^2 for calculation, hence the multiplying by 2 here:
```
ashelkovnykov: Please capitalize the first word in each comment

photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/RMSEEvaluator.scala
jianlingzhong: makes sense, updated.

photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/RMSEEvaluator.scala
jianlingzhong: fixed.

photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/SquaredLossEvaluator.scala
```diff
@@ -34,7 +34,9 @@ object SquaredLossEvaluator extends SingleEvaluator {
   override def evaluate(scoresAndLabelsAndWeights: RDD[(Double, Double, Double)]): Double =
     scoresAndLabelsAndWeights
       .map { case (score, label, weight) =>
-        weight * SquaredLossFunction.lossAndDzLoss(score, label)._1
+        // Squared loss function is calculated as 1/2 (z - y)^2 for mathematical convenience.
+        // However, RMSE should be using (z - y)^2 for calculation, hence the multiplying by 2 here:
```
ashelkovnykov: Comment outdated (still references RMSE from previous version of PR)

photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
```diff
@@ -77,7 +77,7 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
   def testFixedEffectsWithIntercept(): Unit = sparkTest("testFixedEffectsWithIntercept", useKryo = true) {
 
     // This is a baseline RMSE capture from an assumed-correct implementation on 01/24/2018
-    val errorThreshold = 1.2
+    val errorThreshold = 2.4
```
ashelkovnykov: Each of these test values has been multiplied by a factor of `2` - but that's incorrect.

Since the old test values were set for the old (incorrect) computation `RMSE = sqrt(MSE / 2)`, we need to correct these test values. But the correct factor is not `2`, it's `sqrt(2)`, since we need to undo the `sqrt` first.

photon-api/src/main/scala/com/linkedin/photon/ml/evaluation/SquaredLossEvaluator.scala
yunboouyang: Updated comments on PR 442

photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
yunboouyang: Changed to the correct threshold in PR 442

PR #422, opened 2019-04-28
Implement numerically stable Pearson Correlation

### LINKED ISSUE #50: Numerically Unstable Pearson Correlation Computation in LocalDataSet.scala
The Pearson Correlation computation in computePearsonCorrelationScore() uses a naive, one-pass, numerically unstable algorithm.

It should be updated to use a more stable algorithm such as the one presented here: http://prod.sandia.gov/techlib/access-control.cgi/2008/086212.pdf

matthieubulte:
The commit fixes #50 by implementing the algorithm described in http://prod.sandia.gov/techlib/access-control.cgi/2008/086212.pdf - as suggested in the issue.

I also added a pathological test case that used to fail with the previous implementation due to numerical instability. The new implementation passes the test. See: `LocalDatasetTest.testPearsonCorrelationScoreStability`.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
@@ -114,7 +114,7 @@ protected[ml] case class LocalDataset(dataPoints: Array[(UniqueSampleId, Labeled
     if (numFeaturesToKeep < numActiveFeatures) {
 
       val labelAndFeatures = dataPoints.map { case (_, labeledPoint) => (labeledPoint.label, labeledPoint.features) }
-      val pearsonScores = LocalDataset.computePearsonCorrelationScore(labelAndFeatures)
+      val pearsonScores = LocalDataset.stableComputePearsonCorrelationScore(labelAndFeatures)
 
       val filteredFeaturesIndexSet = pearsonScores
         .toArray
@@ -131,7 +131,6 @@ protected[ml] case class LocalDataset(dataPoints: Array[(UniqueSampleId, Labeled
       }
 
       LocalDataset(filteredActivities)
-
     } else {
       this
     }
@@ -180,13 +179,92 @@ object LocalDataset {
   }
 
   /**
+   * Compute Pearson correlation scores using a numerically stable algorithm.
+   *
+   * @param labelAndFeatures An array of (label, feature) tuples
+   * @return The Pearson correlation scores for each tuple
+   */
+  protected[ml] def stableComputePearsonCorrelationScore(
+    labelAndFeatures: Array[(Double, Vector[Double])]): Map[Int, Double] = {
+
+    val featureMeans = mutable.Map[Int, Double]()
+    val featureUnscaledVars = mutable.Map[Int, Double]()
+    var labelMean = 0.0
+    var labelUnscaledVariance = 0.0
+    val unscaledCovariances = mutable.Map[Int, Double]()
+    var interceptAdded = false
+    var numSamples = 0
+
+    labelAndFeatures.foreach { case (label, features) =>
+      numSamples += 1
+
+      val deltaLabel = label - labelMean
+      labelMean += deltaLabel / numSamples
+      labelUnscaledVariance += deltaLabel * (label - labelMean)
+
+      // Note that, if there is duplicated keys in the feature vector, then the following Pearson correlation scores
+      // calculation will screw up
+      features.iterator.foreach { case (key, value) =>
+        val prevFeatureMean = featureMeans.getOrElse(key, 0.0)
+        val deltaFeature = value - prevFeatureMean
+        val featureMean = prevFeatureMean + deltaFeature / numSamples
+
+        val prevFeatureUnscaledVar = featureUnscaledVars.getOrElse(key, 0.0)
+        val featureUnscaledVar = prevFeatureUnscaledVar + deltaFeature * (value - featureMean)
+
+        val prevCovariance = unscaledCovariances.getOrElse(key, 0.0)
+        val unscaledCovariance = prevCovariance + deltaFeature * deltaLabel * (numSamples - 1) / numSamples
+
+        featureMeans.update(key, featureMean)
+        featureUnscaledVars.update(key, featureUnscaledVar)
+        unscaledCovariances.update(key, unscaledCovariance)
+      }
+    }
+
+    val labelStd = math.sqrt(labelUnscaledVariance)
+
+    featureMeans
+      .iterator
+      .map { case (key, featureMean) =>
+        val featureStd = math.sqrt(featureUnscaledVars(key))
+        val covariance = unscaledCovariances(key)
+
+        // When the standard deviation of the feature is close to 0 we treat it as the intercept term.
+        val score = if (featureStd < math.sqrt(numSamples) * MathConst.EPSILON) {
+          // Note that if the mean and standard deviation are equal to zero, it either means that the feature is constant
+          if (featureMean == 1.0 && !interceptAdded) {
+            interceptAdded = true
+            1.0
+          } else {
+            0.0
+          }
+        } else {
+          covariance / (labelStd * featureStd + MathConst.EPSILON)
+        }
+
+        require(math.abs(score) <= 1 + MathConst.EPSILON,
+          s"Computed pearson correlation score is $score, while the score's magnitude should be less than 1. " +
+            s"(Diagnosis:\n" +
+            s"featureKey=$key\n" +
+            s"featureStd=$featureStd\n" +
+            s"labelStd=$labelStd\n" +
+            s"covariance=$covariance\n" +
+            s"numSamples=$numSamples\n" +
+            s"labelAndFeatures used to compute Pearson correlation score:\n${labelAndFeatures.mkString("\n")}})")
+
+        (key, score)
+      }
+      .toMap
+  }
+
+  /**
    * Compute Pearson correlation scores.
    *
    * @param labelAndFeatures An array of (label, feature) tuples
    * @return The Pearson correlation scores for each tuple
    */
   protected[ml] def computePearsonCorrelationScore(
-      labelAndFeatures: Array[(Double, Vector[Double])]): Map[Int, Double] = {
+    labelAndFeatures: Array[(Double, Vector[Double])]): Map[Int, Double] = {
 
     val featureLabelProductSums = mutable.Map[Int, Double]()
     val featureFirstOrderSums = mutable.Map[Int, Double]()
@@ -233,16 +311,16 @@ object LocalDataset {
 
         require(math.abs(score) <= 1 + MathConst.EPSILON,
           s"Computed pearson correlation score is $score, while the score's magnitude should be less than 1. " +
-          s"(Diagnosis:\n" +
-          s"numerator=$numerator\n" +
-          s"denominator=$denominator\n" +
-          s"numSamples=$numSamples\n" +
-          s"featureFirstOrderSum=$featureFirstOrderSum\n" +
-          s"featureSecondOrderSum=$featureSecondOrderSum\n" +
-          s"featureLabelProductSum=$featureLabelProductSum\n" +
-          s"labelFirstOrderSum=$labelFirstOrderSum\n" +
-          s"labelSecondOrderSum=$labelSecondOrderSum\n" +
-          s"labelAndFeatures used to compute Pearson correlation score:\n${labelAndFeatures.mkString("\n")}})")
+            s"(Diagnosis:\n" +
+            s"numerator=$numerator\n" +
+            s"denominator=$denominator\n" +
+            s"numSamples=$numSamples\n" +
+            s"featureFirstOrderSum=$featureFirstOrderSum\n" +
+            s"featureSecondOrderSum=$featureSecondOrderSum\n" +
+            s"featureLabelProductSum=$featureLabelProductSum\n" +
+            s"labelFirstOrderSum=$labelFirstOrderSum\n" +
+            s"labelSecondOrderSum=$labelSecondOrderSum\n" +
+            s"labelAndFeatures used to compute Pearson correlation score:\n${labelAndFeatures.mkString("\n")}})")
 
         (key, score)
       }
--- a/photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
+++ b/photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
@@ -29,7 +29,7 @@ import com.linkedin.photon.ml.test.CommonTestUtils
 class LocalDatasetTest {
 
   /**
-   * Test the Pearson correlation score computation.
+   * Test the stable Pearson correlation score computation.
    */
   @Test(groups = Array[String]("testPearsonCorrelationScore", "testCore"))
   def testPearsonCorrelationScore(): Unit = {
@@ -38,9 +38,9 @@ class LocalDatasetTest {
     val labels = Array(1.0, 4.0, 6.0, 9.0)
     val features = Array(
       Vector(0.0, 0.0, 2.0), Vector(5.0, 0.0, -3.0), Vector(7.0, 0.0, -8.0), Vector(0.0, 0.0, -1.0))
-    val expected = Map(0 -> 0.05564149, 1 -> 1.0, 2 -> -0.40047142)
+    val expected = Map(0 -> 0.05564149, 1 -> 0.0, 2 -> -0.40047142)
     val labelAndFeatures = labels.zip(features)
-    val computed = LocalDataset.computePearsonCorrelationScore(labelAndFeatures)
+    val computed = LocalDataset.stableComputePearsonCorrelationScore(labelAndFeatures)
 
     computed.foreach { case (key, value) =>
       assertEquals(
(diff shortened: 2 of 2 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -188,61 +188,69 @@ object LocalDataset {
...
       features.activeIterator.foreach { case (key, value) =>
-        featureFirstOrderSums.update(key, featureFirstOrderSums.getOrElse(key, 0.0) + value)
-        featureSecondOrderSums.update(key, featureSecondOrderSums.getOrElse(key, 0.0) + value * value)
-        featureLabelProductSums.update(key, featureLabelProductSums.getOrElse(key, 0.0) + value * label)
+        val prevFeatureMean = featureMeans.getOrElse(key, 0.0)
+        val deltaFeature = value - prevFeatureMean
+        val featureMean = prevFeatureMean + deltaFeature / numSamples
+
+        val prevFeatureUnscaledVar = featureUnscaledVars.getOrElse(key, 0.0)
+        val featureUnscaledVar = prevFeatureUnscaledVar + (value - prevFeatureMean) * (value - featureMean)
```
cmjiang: You can use `deltaFeature` to replace `(value - prevFeatureMean)`.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -188,61 +188,69 @@ object LocalDataset {
...
+    var numSamples = 0
 
     labelAndFeatures.foreach { case (label, features) =>
       numSamples += 1
-      labelFirstOrderSum += label
-      labelSecondOrderSum += label * label
+
+      val deltaLabel = label - labelMean
+      labelMean += deltaLabel / numSamples
+      labelUnscaledVar += deltaLabel * (label - labelMean)
```
cmjiang: Maybe name it `labelUnscaledVariance` would be better.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -188,61 +188,69 @@ object LocalDataset {
...
-        featureLabelProductSums.update(key, featureLabelProductSums.getOrElse(key, 0.0) + value * label)
+        val prevFeatureMean = featureMeans.getOrElse(key, 0.0)
+        val deltaFeature = value - prevFeatureMean
+        val featureMean = prevFeatureMean + deltaFeature / numSamples
+
+        val prevFeatureUnscaledVar = featureUnscaledVars.getOrElse(key, 0.0)
+        val featureUnscaledVar = prevFeatureUnscaledVar + (value - prevFeatureMean) * (value - featureMean)
+
+        val prevCovariance = unscaledCovariances.getOrElse(key, 0.0)
+        val unscaledCovariance = prevCovariance + deltaFeature * deltaLabel * (numSamples - 1) / numSamples
```
cmjiang: This line basically implements formula (3.1) by adding one element to the set of (u, v). Very neat!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -188,61 +188,69 @@ object LocalDataset {
...
+
+        featureMeans.update(key, featureMean)
+        featureUnscaledVars.update(key, featureUnscaledVar)
+        unscaledCovariances.update(key, unscaledCovariance)
       }
     }
 
-    featureFirstOrderSums
+    val labelStd = math.sqrt(labelUnscaledVar / numSamples)
```
cmjiang: Is it necessary to scale the std by `numSamples` for `labelStd`, `featureStd` and `covariance`? Mathematically it's the same without `numSamples` for these three variable. However, it might affect the choice of `MathConst.EPSILON` for stably computing the score. Say `numSamples` is 100M, then the product of `labelStd * featureStd` will be very small. I meant it's not in the same magnitude of the `denominator` in the current code.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: Will do!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: Will do!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: I'm not sure what you suggest, but here what I think for very large `numSamples` (let's say 100M as you suggested).

Before normalization, the three "unscaled" variables are sums of 100M terms with mean 0. The rescaling brings these very large sum back to the scale of the data's variance / std.

Even if the rescaling can be skipped, it is necessary to rescale the epsilon at two different place, so it doesn't really make a difference in term of code.

I wouldn't mind implementing what you suggest, but I'm not sure which benefits there would be. I might simply be missing something though :)

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: > I suggest to keep the original function, and create a new function to do the stable computation . Then we can test both functions with the same examples in the Unit Test.
That's a great idea and I was thinking of doing it (there is no proof that I'm not lying about the instability old code ;) ) but I wan't sure how to integrate it.

I see the following option but I'm not sure which one you would prefer:
1. Put the old implementation as helper function in the test file to simply demonstrate that it fails without polluting `LocalDataset`
2. Put the old implementation in `LocalDataset`, rename it to `unstableComputePearsonCorrelationScore` and mark it as deprecated.
3. Put the old implementation in `LocalDataset`, rename it to `unstableComputePearsonCorrelationScore` and remove it before merging.


> Thank you @matthieubulte for your contribution and nice implementation!
My pleasure, thank you for open sourcing this great project!

matthieubulte:
> I suggest to keep the original function, and create a new function to do the stable computation . Then we can test both functions with the same examples in the Unit Test.
That's a great idea and I was thinking of doing it (there is no proof that I'm not lying about the instability old code ;) ) but I wan't sure how to integrate it.

I see the following option but I'm not sure which one you would prefer:
1. Put the old implementation as helper function in the test file to simply demonstrate that it fails without polluting `LocalDataset`
2. Put the old implementation in `LocalDataset`, rename it to `unstableComputePearsonCorrelationScore` and mark it as deprecated.
3. Put the old implementation in `LocalDataset`, rename it to `unstableComputePearsonCorrelationScore` and remove it before merging.


> Thank you @matthieubulte for your contribution and nice implementation!
My pleasure, thank you for open sourcing this great project!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: I might misunderstand something from your comment, but assuming we have a dataset of 100M observations:

1. Removing the  `numSamples` normalization from the var and std will require - as you mentioned - to rescale `math.epsilon`. This means we wouldn't have that much of a cosmetic advantage of doing that.

2. The variables you mention are sums of 100M positive numbers centered around 0, they are potentially very big - same order of magnitude as `numSamples`. Normalizing them brings them back to the scale of the var and std, so i'm not sure why the product you mentioned should be that small.

I might be missing something obvious, otherwise I'm not sure what could motivate this change.

cmjiang: Don't worry this is a minor issue which won't affect the result too much. My concern is that since the old code did not normalize the numerator and denominator, we might need to be careful about choosing the EPSILON.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -188,61 +188,69 @@ object LocalDataset {
...
-      labelFirstOrderSum += label
-      labelSecondOrderSum += label * label
+
+      val deltaLabel = label - labelMean
+      labelMean += deltaLabel / numSamples
+      labelUnscaledVar += deltaLabel * (label - labelMean)
+
       // Note that, if there is duplicated keys in the feature vector, then the following Pearson correlation scores
       // calculation will screw up
       features.activeIterator.foreach { case (key, value) =>
```
cmjiang: My major concern is on this `activeIterator` here. In the old implementation, we could simple ignore the zeros for sparse features. However, the new implementation should take zeros into consideration. Then first, I will dramatically increase the computational cost by treating a large sparse vector as dense. Also, this implementation might be wrong because the `activeIterator` seems to take only non-zero values.

cmjiang: This can be easily verified with some sparse features in the Unit Test.

matthieubulte: Nice catch, and thanks for the comment!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -188,61 +188,69 @@ object LocalDataset {
...
-          s"denominator=$denominator\n" +
-          s"numSamples=$numSamples\n" +
-          s"featureFirstOrderSum=$featureFirstOrderSum\n" +
-          s"featureSecondOrderSum=$featureSecondOrderSum\n" +
-          s"featureLabelProductSum=$featureLabelProductSum\n" +
-          s"labelFirstOrderSum=$labelFirstOrderSum\n" +
-          s"labelSecondOrderSum=$labelSecondOrderSum\n" +
-          s"labelAndFeatures used to compute Pearson correlation score:\n${labelAndFeatures.mkString("\n")}})")
+            s"(Diagnosis:\n" +
+            s"featureKey=$key\n"+
```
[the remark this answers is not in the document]
matthieubulte: Will do!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: > @matthieubulte @cmjiang
> My preference would be similar to option 2, and exactly what Chengming suggested: rename this function to `stableComputePearsonCorrelationScore` and add another unit test with a `SparseVector`. Let's keep both functions around for now, as it's unclear what the future of feature filtering using Pearson correlation score will be.
> 
> Thank you very much for the contribution @matthieubulte

I will take care of this as well.

matthieubulte:
I have currently fixed all of the issues, but I would like to discuss one open topic before pushing the code again.

I'm not sure if the way intercepts are treated in the computation of the correlation is very consistent.

The current way it is being done is by treating the first feature encountered with a numerically 0 variance as the intercept term, assigning it a score of 1.0, and then assigning to every following zero-variance feature a score of 0.0. You can find this at https://github.com/linkedin/photon-ml/blob/549bcf3a1d3b7a53aec6e0d998ab0b7a702e48b7/photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala#L224

However, the risks I see (and it was confirmed when I changed the implementation) are:
1. The ordering of featuresMean, which is used to decide what is considered to be the intercept feature, is not reliable and is dependent on the underlying Map data structure. 
2. If the feature is a 0 constant or simply missing, it might be treated as the intercept, which clearly doesn't make sense, since one would want to get rid of such missing .

I would thus suggest the following fix: instead of just taking the first feature encountered with a zero variance, we could additionally check if the mean of the feature is non-zero. If both these conditions are fulfilled, we could treat the feature as an intercept. The resulting code would be 

```scala
// When the standard deviation of the feature is close to 0 we treat it as the intercept term.
val score = if (featureStd < math.sqrt(numSamples) * MathConst.EPSILON) {
  // Note that, if the mean and standard deviation are equal to zero, it either means that the feature is constant
  if (interceptAdded || featureMean == 0.0) {
    0.0
  } else {
    interceptAdded = true
    1.0
  }
} else {
  ...
}
```

However, this would also imply changing most Pearson correlation related unit tests, since they are all happy with treating a constant 0 feature as the intercept.

What is your opinion about this?

matthieubulte:
> I have currently fixed all of the issues, but I would like to discuss one open topic before pushing the code again.
> 
> I'm not sure if the way intercepts are treated in the computation of the correlation is very consistent.
> 
> The current way it is being done is by treating the first feature encountered with a numerically 0 variance as the intercept term, assigning it a score of 1.0, and then assigning to every following zero-variance feature a score of 0.0. You can find this at
> 
> https://github.com/linkedin/photon-ml/blob/549bcf3a1d3b7a53aec6e0d998ab0b7a702e48b7/photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala#L224
> 
> However, the risks I see (and it was confirmed when I changed the implementation) are:
> 
> 1. The ordering of featuresMean, which is used to decide what is considered to be the intercept feature, is not reliable and is dependent on the underlying Map data structure.
> 2. If the feature is a 0 constant or simply missing, it might be treated as the intercept, which clearly doesn't make sense, since one would want to get rid of such missing .
> 
> I would thus suggest the following fix: instead of just taking the first feature encountered with a zero variance, we could additionally check if the mean of the feature is non-zero. If both these conditions are fulfilled, we could treat the feature as an intercept. The resulting code would be
> 
> ```scala
> // When the standard deviation of the feature is close to 0 we treat it as the intercept term.
> val score = if (featureStd < math.sqrt(numSamples) * MathConst.EPSILON) {
>   // Note that, if the mean and standard deviation are equal to zero, it either means that the feature is constant
>   if (interceptAdded || featureMean == 0.0) {
>     0.0
>   } else {
>     interceptAdded = true
>     1.0
>   }
> } else {
>   ...
> }
> ```
> 
> However, this would also imply changing most Pearson correlation related unit tests, since they are all happy with treating a constant 0 feature as the intercept.
> 
> What is your opinion about this?

I ended up implementing the changes I've suggested. You can now find the result in the current state of the branch.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -180,11 +180,91 @@ object LocalDataset {
...
+
+        // When the standard deviation of the feature is close to 0 we treat it as the intercept term.
+        val score = if (featureStd < math.sqrt(numSamples) * MathConst.EPSILON) {
+          // Note that if the mean and standard deviation are equal to zero, it either means that the feature is constant
+          if (interceptAdded || featureMean == 0.0) {
+            0.0
+          } else {
+            interceptAdded = true
+            1.0
+          }
```
ashelkovnykov: Won't this treat any (near) constant feature as the intercept and keep it around? We can make the check for the intercept more explicit, since the mean of the intercept should be exactly `1.0`:
```
val score = if (featureStd < math.sqrt(numSamples) * MathConst.EPSILON) {
  if (featureMean == 1.0) {
    1.0
  } else {
    0.0
  }
} else {
  ...
}
```

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -180,11 +180,91 @@ object LocalDataset {
...
+        val covariance = unscaledCovariances(key)
+
+        // When the standard deviation of the feature is close to 0 we treat it as the intercept term.
+        val score = if (featureStd < math.sqrt(numSamples) * MathConst.EPSILON) {
+          // Note that if the mean and standard deviation are equal to zero, it either means that the feature is constant
+          if (interceptAdded || featureMean == 0.0) {
+            0.0
+          } else {
+            interceptAdded = true
+            1.0
```
yunboouyang: Should the score always be 0.0 when we compute the correlation between the label and any constant?

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: Yes, I did it like this because it doesn't really make any difference mathematically which column is the "real" intercept.

I will incorporate your change!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
matthieubulte: The correlation between a constant and a random variable is ill-defined. One could argue that it is 0 [1], but I think here it's more relevant to keep in mind hoe the function will be used. 

Since it will be used to filter-out features, we do want to get rid of all-but-one constant features, this allows to get rid of the co-linearity of the many constant columns.

So since  I would argue that we want to keep only one "intercept" column, we would want to give a low Pearson Correlation score in order to get rid of these correlations first.

[1] https://stats.stackexchange.com/a/267155

photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
```diff
@@ -51,10 +84,60 @@ class LocalDatasetTest {
...
+    }
+  }
+
+
+
   /**
    * Test feature filtering using Pearson correlation score.
    */
-  @Test(dependsOnGroups = Array[String]("testPearsonCorrelationScore", "testCore"))
+  @Test(dependsOnGroups = Array[String]("testStablePearsonCorrelationScore", "testCore"))
```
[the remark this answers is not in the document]
matthieubulte: Yes, will do.

photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
```diff
@@ -51,10 +84,60 @@ class LocalDatasetTest {
...
+      assertEquals(
+        expected(key),
+        value,
+        CommonTestUtils.LOW_PRECISION_TOLERANCE,
+        s"Computed Pearson correlation score is $value, while the expected value is ${expected(key)}.")
+    }
+  }
+
+
+
```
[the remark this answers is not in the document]
matthieubulte: Yes, will do.

photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
```diff
@@ -51,10 +84,60 @@ class LocalDatasetTest {
...
+  /**
+    * Test that the stable Pearson correlation score computation properly recognizes an intercept column.
+    */
+  @Test(groups = Array[String]("testStablePearsonCorrelationScoreForIntercept", "testCore"))
+  def testPearsonCorrelationScoreForIntercept(): Unit = {
+
+    // Test input data
+    val labels = Array(1.0, 4.0, 6.0, 9.0)
+    val features = Array(
+      Vector(0.0, 0.0, 1.0, 2.0), Vector(5.0, 0.0, 1.0, -3.0), Vector(7.0, 0.0, 1.0, -8.0), Vector(0.0, 0.0, 1.0, -1.0))
```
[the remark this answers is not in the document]
matthieubulte: Yes, will do.

photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
```diff
@@ -28,19 +28,52 @@ import com.linkedin.photon.ml.test.CommonTestUtils
...
 class LocalDatasetTest {
 
+
   /**
-   * Test the Pearson correlation score computation.
-   */
-  @Test(groups = Array[String]("testPearsonCorrelationScore", "testCore"))
+    * Test the stable Pearson correlation score computation.
+    */
+  @Test(groups = Array[String]("testStablePearsonCorrelationScore", "testCore"))
```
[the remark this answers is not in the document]
matthieubulte: Yes, will do.

photon-api/src/test/scala/com/linkedin/photon/ml/data/LocalDatasetTest.scala
```diff
@@ -28,19 +28,52 @@ import com.linkedin.photon.ml.test.CommonTestUtils
  */
 class LocalDatasetTest {
 
+
   /**
-   * Test the Pearson correlation score computation.
-   */
-  @Test(groups = Array[String]("testPearsonCorrelationScore", "testCore"))
+    * Test the stable Pearson correlation score computation.
+    */
```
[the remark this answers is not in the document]
matthieubulte: Yes, will do.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -180,11 +180,91 @@ object LocalDataset {
...
 
   /**
-   * Compute Pearson correlation scores.
+   * Compute Pearson correlation scores using a numerically stable algorithm.
    *
    * @param labelAndFeatures An array of (label, feature) tuples
    * @return The Pearson correlation scores for each tuple
    */
+  protected[ml] def stableComputePearsonCorrelationScore(
+      labelAndFeatures: Array[_ <: (Double, Vector[Double])]): Map[Int, Double] = {
```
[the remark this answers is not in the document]
matthieubulte: To be honest, I only introduced this to be able to test the function with a `SparseVector` features. Trying to use the function with `SparceVector`s wouldn't compile with the previous signature.

This is probably simply due to my lack of experience with Scala, I would be very thankful to get a hint from you there if there is a better way to do this.

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
```diff
@@ -180,11 +180,91 @@ object LocalDataset {
...
+      .toMap
+  }
+
+
+  /**
+    * Compute Pearson correlation scores.
+    *
+    * @param labelAndFeatures An array of (label, feature) tuples
+    * @return The Pearson correlation scores for each tuple
+    */
```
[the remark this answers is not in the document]
matthieubulte: Will do!

photon-api/src/main/scala/com/linkedin/photon/ml/data/LocalDataset.scala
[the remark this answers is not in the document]
ashelkovnykov: Sure - I just pulled the code locally.

Looks like the problem is the `Array` in 'testStableComputePearsonCorrelationScore' (line 65 of `LocalDatasetTest`). Its type is being detected by Scala as `Array[SparseVector[Double]]` and this is causing the compilation error you encountered when Scala later attempts to cast it to `Array[Vector[Double]]`.

If you explicitly cast it to use `Vector` instead, there shouldn't be an issue:
```
val features = Array[Vector[Double]](
  new SparseVector[Double](...),
  ...
)
```

matthieubulte:
Thank you very much for the comments and improvements. I have now pushed the changes, trying to implement as much as I could. Two remarks:

1. A couple of tests only apply to the new version of the Pearson correlation since the new function fixes the issues in the old code.

2. I messed up something with the code formater in my before-last commit which automatically removed imports, breaking the code. Now everything is good but the commit history is kind of stupid. Do we want to leave it that way?

Thanks!

ashelkovnykov:
@matthieubulte 

Looking great, thanks for working with us to make a great PR. With regards to the commit history: no, please crush the changes into one commit. Our standard practice for commit messages is as follows:
```
One line summary of all changes or overall purpose of the commit

- Summary of individual changes
- One change per line
- Uses list format
```

matthieubulte:
Well, thank you all for helping out with this PR, this was a lot of work for you as well!

The current version should now be satisfying.

Thanks!

ashelkovnykov:
@matthieubulte Sorry for the delay - it's merged now. Thanks again!

matthieubulte:
It was a pleasure to contribute! Thank you very much to all of you for your help and support.

PR #424, opened 2019-05-07
Minor changes to command line parameter parsers.

ashelkovnykov:


photon-client/src/test/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverTest.scala
```diff
@@ -180,8 +180,6 @@ class GameTrainingDriverTest {
       Array(validParamMap.copy.remove(GameTrainingDriver.inputDataDirectories)),
       // No root output directory
       Array(validParamMap.copy.remove(GameTrainingDriver.rootOutputDirectory)),
-      // No feature bags directory
```
cmjiang: Does this change allow user to input an empty feature bag without raising an error? Therefore users don't need to manually change the params at job configurations.

jianlingzhong:
looks good to me.

photon-client/src/test/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverTest.scala
ashelkovnykov: No, but that's a good point - let me add that change.

yunboouyang: Can you illustrate why you want to delete this test?

ashelkovnykov: Feature white lists are not a mandatory feature - not having them present will not cause parameter parsing to fail, and thus is not invalid input, which is what this test checks.

PR #425, opened 2019-05-29
Refactor random effect projection code

ashelkovnykov:
This PR removes many of the classes used to project random effect data to a lower dimension. The option to have different projection schemes was removed, and the linear subspace projection was made mandatory.


photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/ModelProjection.scala
```diff
@@ -0,0 +1,85 @@
...
+            model.updateCoefficients(newCoefficients)
+          }
+          .getOrElse(model)
+      }
+
+    randomEffectModel.update(newModels)
+  }
+
+  /**
+   * Project a [[RandomEffectModel]] from the projected space to the original space.
```
cmjiang: I don't quite understand the logic here for projecting backward. How can you project from the projected space to the original space? I guess the projection is not invertible.

cmjiang: I think I got this idea. This projection is for projecting a sparse vector to dense (forward) and dense to sparse (backward) for inactive features/weights.

PR #426, opened 2019-05-30
Fix output stream writer to UTF-8 format

yunboouyang:
Fix output stream writer to UTF-8 format to solve generating inconsistent feature list issue when running NameAndTermFeatureBagsDriver in photon-client

### DIFF
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
@@ -222,7 +222,7 @@ object IOUtils {
     val stream = fs.create(outputPath, forceOverwrite)
     val writer = new PrintWriter(
       new BufferedWriter(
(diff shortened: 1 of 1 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -222,7 +223,7 @@ object IOUtils {
     val stream = fs.create(outputPath, forceOverwrite)
     val writer = new PrintWriter(
       new BufferedWriter(
-        new OutputStreamWriter(stream)
+        new OutputStreamWriter(stream, UTF_8.name())
```
joshvfleming: You can pass the `UTF_8` charset directly -- no need to call `name()`.

yunboouyang: thank you for your suggestion! Works as well

PR #432, opened 2019-07-26
[WIP] Separate Regularization for intercepts and coefficients

biyan-linkedin:
[Early-stage dev]

Add interceptIndex in L2Regularization and subtract its penalization in L2RegValue(). 
Update related classes while passing interceptIndex around using curry.


photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -225,6 +231,8 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
     val normalizationContextsOpt = get(coordinateNormalizationContexts)
     val ignoreThreshold = getOrDefault(ignoreThresholdForNewModels)
     val numUniqueCoordinates = updateSequence.toSet.size
+    val indices = get(interceptIndices)
+
```
cmjiang: Remove extra line.

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -642,12 +652,13 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
...
+    val indices = get(interceptIndices).getOrElse(Seq.empty[Int])
 
     // Create the optimization coordinates for each component model
     val coordinates: Map[CoordinateId, C forSome { type C <: Coordinate[_] }] =
-      updateSequence
-        .map { coordinateId =>
-          val coordinate: C forSome { type C <: Coordinate[_] } = if (lockedCoordinates.contains(coordinateId)) {
+      updateSequence.zip(indices)
+        .map { case (coordinateId, interceptIndex) => {
+          val coordinate: C forSome {type C <: Coordinate[_]} = if (lockedCoordinates.contains(coordinateId)) {
```
cmjiang: Add the white spaces back.

photon-api/src/main/scala/com/linkedin/photon/ml/function/glm/DistributedGLMLossFunction.scala
```diff
@@ -160,17 +160,17 @@ object DistributedGLMLossFunction {
   def apply(
       configuration: GLMOptimizationConfiguration,
       singleLossFunction: PointwiseLossFunction,
-      treeAggregateDepth: Int): DistributedGLMLossFunction = {
+      treeAggregateDepth: Int,
+      index: Int = -1): DistributedGLMLossFunction = {
```
cmjiang: Can you add a comment for this parameter?

photon-api/src/main/scala/com/linkedin/photon/ml/function/glm/GLMLossFunction.scala
```diff
@@ -33,14 +33,15 @@ object GLMLossFunction {
    */
   def buildFactory(
       lossFunction: PointwiseLossFunction,
-      treeAggregateDepth: Int): (CoordinateOptimizationConfiguration) => ObjectiveFunction =
+      treeAggregateDepth: Int,
+      interceptIndex: Int = -1): (CoordinateOptimizationConfiguration) => ObjectiveFunction =
```
cmjiang: Can you add a comment for this parameter?

photon-api/src/main/scala/com/linkedin/photon/ml/function/glm/SingleNodeGLMLossFunction.scala
```diff
@@ -148,7 +148,8 @@ object SingleNodeGLMLossFunction {
    */
   def apply(
       configuration: GLMOptimizationConfiguration,
-      singleLossFunction: PointwiseLossFunction): SingleNodeGLMLossFunction = {
+      singleLossFunction: PointwiseLossFunction,
+      index: Int = -1): SingleNodeGLMLossFunction = {
```
cmjiang: Add comment here too.

photon-api/src/main/scala/com/linkedin/photon/ml/function/svm/DistributedSmoothedHingeLossFunction.scala
```diff
@@ -113,14 +113,16 @@ object DistributedSmoothedHingeLossFunction {
    */
   def apply(
       configuration: GLMOptimizationConfiguration,
-      treeAggregateDepth: Int): DistributedSmoothedHingeLossFunction = {
+      treeAggregateDepth: Int,
+      index: Int = -1): DistributedSmoothedHingeLossFunction = {
```
cmjiang: Same.

photon-api/src/main/scala/com/linkedin/photon/ml/function/svm/SingleNodeSmoothedHingeLossFunction.scala
```diff
@@ -99,7 +99,9 @@ object SingleNodeSmoothedHingeLossFunction {
    * @param configuration The optimization problem configuration
    * @return A new SingleNodeSmoothedHingeLossFunction
    */
-  def apply(configuration: GLMOptimizationConfiguration): SingleNodeSmoothedHingeLossFunction = {
+  def apply(
+    configuration: GLMOptimizationConfiguration,
+    index: Int = -1): SingleNodeSmoothedHingeLossFunction = {
```
cmjiang: Same.

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
biyan-linkedin: Ack.

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -131,6 +131,11 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
     "use warm start",
     "Whether to train the current model with coefficients initialized by the previous model.")
 
+  val interceptIndices: Param[Seq[Int]] = ParamUtils.createParam[Seq[Int]](
+    "intercept indices",
+    "A list of integers that indicate the index of the intercept in each feature bag."
+  )
```
li-ashelkov: There's too much implicit information here - the relationship between intercept indices and coordinates should be made explicit. The indices should also be optional. We can create a default value which will return `None` for all coordinates, for the case where no input is given.
```suggestion
  val coordinateInterceptIndices: Param[Map[CoordinateId, Option[Int]]] = ParamUtils.createParam[Map[CoordinateId, Option[Int]]](
  "coordinate intercept indices",
  "A map of coordinate ID to intercept index, if enabled.")
```
However, now I wonder if we can insert this information into another class so that this parameter doesn't need to exist.

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -170,6 +175,7 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
 
   def setUseWarmStart(value: Boolean): this.type = set(useWarmStart, value)
 
+  def setInterceptIndices(value: Seq[Int]): this.type = set(interceptIndices, value)
```
li-ashelkov: See above recommended change.

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
li-ashelkov: See above change - it would make both of these lines unnecessary.

In addition, we would need to add the following default value for `coordinateInterceptIndices` in `setDefaultParams`:
```
setDefault(coordianteInterceptIndices, Map().withDefaultValue(None))
```

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -642,12 +652,13 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
     }
     val downSamplerFactory = DownSamplerHelper.buildFactory(task)
     val lockedCoordinates = get(partialRetrainLockedCoordinates).getOrElse(Set())
+    val indices = get(interceptIndices).getOrElse(Seq.empty[Int])
```
li-ashelkov: Please remove this line - such code is best avoided by setting a default value and replacing the call with `getOrDefault`.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -119,7 +133,7 @@ trait L2RegularizationDiff extends DiffFunction with L2Regularization {
       normalizationContext: BroadcastWrapper[NormalizationContext]): (Double, Vector[Double]) = {
 
     val (baseValue, baseGradient) = super.calculate(input, coefficients, normalizationContext)
-    val valueWithRegularization = baseValue + l2RegValue(convertToVector(coefficients))
+    val valueWithRegularization = baseValue + l2RegValue(convertToVector(coefficients), interceptIndex)
```
li-ashelkov: We shouldn't need to pass `interceptIndex` as it is a property of the trait, thus can already be accessed from `l2RegValue`.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -59,16 +67,22 @@ trait L2Regularization extends ObjectiveFunction {
...
 
   /**
    * Compute the L2 regularization value for the given model coefficients.
    *
    * @param coefficients The model coefficients
    * @return The L2 regularization value
    */
-  protected def l2RegValue(coefficients: Vector[Double]): Double =
-    l2RegWeight * coefficients.dot(coefficients) / 2
+  protected def l2RegValue(coefficients: Vector[Double], interceptIndex: Int = -1): Double = {
```
li-ashelkov: See comment below, can remove `interceptIndex` from function parameters.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -59,16 +67,22 @@ trait L2Regularization extends ObjectiveFunction {
...
+    val l2Regularization = interceptIndex match {
+      case -1 => l2RegWeight * coefficients.dot(coefficients) / 2
+      case x if x >= 0 => l2RegWeight * (coefficients.dot(coefficients) - pow(coefficients(x), 2)) / 2
+      case other => throw new IllegalArgumentException(s"Encountered invalid intercept index $other")
```
li-ashelkov: We can remove this check by having a `require` statement above the L2 weight getter:
```
require(interceptIndexOpt.forall(_ >= 0), "Invalid intercept index ${interceptIndexOpt.get}")
```

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -59,16 +67,22 @@ trait L2Regularization extends ObjectiveFunction {
...
+    }
+    l2Regularization
```
li-ashelkov: Can drop some repeated calculations:
```
val interceptModifier = interceptIndexOpt match {
  case Some(x) => pow(coefficients(x), 2)
  case None => 0
}

l2RegWeight * (coefficients.dot(coefficients) - interceptModifier) / 2
```

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -26,6 +27,7 @@ import com.linkedin.photon.ml.util.BroadcastWrapper
 trait L2Regularization extends ObjectiveFunction {
 
   protected var l2RegWeight: Double = 0D
+  protected var interceptIndex: Int = -1
```
li-ashelkov: I think that we should have this be an `Option` instead. It will simplify the safety check and save us from having to declare a constant with value `-1` which is then referenced by `ObjectiveFunctionHelper` and the loss function classes.

In this case, we should also rename the member to `interceptIndexOpt`.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -34,6 +36,12 @@ trait L2Regularization extends ObjectiveFunction {
    */
   def l2RegularizationWeight: Double = l2RegWeight
 
+  /**
+    * Getter of interceptIndex
+    * @return
+    */
+  def getIntercetIndex: Int = interceptIndex
```
li-ashelkov: Typo in name. Also, see above for recommended member type change.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -59,16 +67,22 @@ trait L2Regularization extends ObjectiveFunction {
       input: Data,
       coefficients: Coefficients,
       normalizationContext: BroadcastWrapper[NormalizationContext]): Double =
-    super.value(input, coefficients, normalizationContext) + l2RegValue(convertToVector(coefficients))
+    super.value(input, coefficients, normalizationContext) + l2RegValue(convertToVector(coefficients), interceptIndex)
```
li-ashelkov: We shouldn't need to pass `interceptIndex` as it is a property of the trait, thus can already be accessed from `l2RegValue`.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -119,7 +133,7 @@ trait L2RegularizationDiff extends DiffFunction with L2Regularization {
       normalizationContext: BroadcastWrapper[NormalizationContext]): (Double, Vector[Double]) = {
 
     val (baseValue, baseGradient) = super.calculate(input, coefficients, normalizationContext)
-    val valueWithRegularization = baseValue + l2RegValue(convertToVector(coefficients))
+    val valueWithRegularization = baseValue + l2RegValue(convertToVector(coefficients), interceptIndex)
     val gradientWithRegularization = baseGradient + l2RegGradient(convertToVector(coefficients))
```
li-ashelkov: Need to modify `l2RegGradient` as well, if intercept is present.

photon-api/src/main/scala/com/linkedin/photon/ml/function/svm/SingleNodeSmoothedHingeLossFunction.scala
li-ashelkov: In addition, parameter name and type should change (see below):
```
interceptIndexOpt: Option[Int] = None
```

photon-api/src/main/scala/com/linkedin/photon/ml/function/svm/DistributedSmoothedHingeLossFunction.scala
li-ashelkov: In addition, parameter name and type should change (see below):
```
interceptIndexOpt: Option[Int] = None
```

photon-api/src/main/scala/com/linkedin/photon/ml/function/glm/SingleNodeGLMLossFunction.scala
li-ashelkov: In addition, parameter name and type should change (see below):
```
interceptIndexOpt: Option[Int] = None
```

photon-api/src/main/scala/com/linkedin/photon/ml/function/glm/GLMLossFunction.scala
li-ashelkov: In addition, parameter name and type should change (see below):
```
interceptIndexOpt: Option[Int] = None
```

photon-api/src/main/scala/com/linkedin/photon/ml/function/glm/DistributedGLMLossFunction.scala
li-ashelkov: In addition, parameter name and type should change (see below):
```
interceptIndexOpt: Option[Int] = None
```

photon-api/src/main/scala/com/linkedin/photon/ml/function/ObjectiveFunctionHelper.scala
```diff
@@ -36,11 +36,11 @@ object ObjectiveFunctionHelper {
    * @return A function which builds the appropriate type of [[ObjectiveFunction]] for a given [[Coordinate]] type and
    *         optimization settings.
    */
-  def buildFactory(taskType: TaskType, treeAggregateDepth: Int): ObjectiveFunctionFactory =
+  def buildFactory(taskType: TaskType, treeAggregateDepth: Int)(interceptIndex: Int = -1): ObjectiveFunctionFactory =
```
li-ashelkov: Don't need to curry here - these parameters are all available at the same time:
```suggestion
  def buildFactory(taskType: TaskType, treeAggregateDepth: Int, interceptIndexOpt: Option[Int] = None): ObjectiveFunctionFactory =
```

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
li-ashelkov: See above comments. Don't `zip` with `Seq` - change parameter type and access it by coordinate ID like the datasets and configurations below.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -363,6 +363,12 @@ object GameTrainingDriver extends GameDriver {
       readValidationData(avroDataReader, featureIndexMapLoaders)
     }
 
+    val interceptIndices = featureIndexMapLoaders.map {
+      case (featureShardId, indexMapLoader) =>
+        indexMapLoader.indexMapForDriver().getIndex(Constants.INTERCEPT_KEY)
+      case _ => -1
+    }.toSeq
```
li-ashelkov: See above comments - we should either have a `Map[CoordinateId, Option[Int]]` or encode the information elsewhere in a parameter that we're already passing to the `GameEstimator`.

I'm not sure where we could encode this information in existing parameters - let me think about it. While I do, you should feel free to ignore that advice and focus on the `Map[CoordinateId, Option[Int]]`.

li-ashelkov:
@cmjiang I'll look into how OWLQN handles regularization of the intercept.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
li-ashelkov: This link explains the reason for calling `forall`:
https://blog.tmorris.net/posts/scalaoption-cheat-sheet/

biyan-linkedin: Is it always true that `INTERCEPT_KEY` exists in `featureIndexMapLoaders`? I set "-1" as default value when it does not exists, so no subtraction. 

Besides, I may need to add another condition that `_ < coefficients.size` and throw exceptions accordingly.

ashelkovnykov: No, an intercept is not guaranteed to exist in `featureIndexMapLoaders`. So the code to fetch it will need to look something like this (in the `GameTrainingDriver`):
```
val featureShardIndices = featureIndexMapLoaders.mapValues { indexMap =>
  indexMap.indexMapForDriver().getIndex(Constants.INTERCEPT_KEY) match {
    case IndexMap.NULL_KEY => None
    case i if i >= 0 => Some(i)
    case i => throw new IndexOutOfBoundsException("...")
  }
}
val coordinateInterceptIndices = getRequiredParam(coordinateConfigurations).mapValues { config =>
  featureShardIndices(config.dataConfiguration.featureShardId)
}
```

You're right - we should do a safety check against `coefficients.size`. Actually, I just realized that because of random effect projection, the index of the intercept will change for each random effect. Let's take a look together later, I'll show you what I mean.

photon-api/src/main/scala/com/linkedin/photon/ml/function/ObjectiveFunctionHelper.scala
biyan-linkedin: I think we still need to keep curry here. `lossFunctionFactory` was created outside of `coordinates` (L658 in `GameEstimator.scala`) and `lossFunctionFactory` takes different `interceptIndex` each time while generating `coordinates` in `GameEstimator.scala`. Do you have any other suggestions?

PR #433, opened 2019-08-02
Make optimization state tracking mandatory

ashelkovnykov:
When using GAME, optimization state tracking already did not have the option of being disabled. When using the legacy driver (i.e. `Driver`), in practice LinkedIn users were always keeping the option enabled. There is no performance penalty for enabling optimization state tracking. Thus, to simplify the code this PR seeks to make it always enabled.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/ModelTraining.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/ModelTraining.scala
@@ -45,8 +45,6 @@ object ModelTraining extends Logging {
    * @param maxNumIter Maximum number of iterations to run
    * @param tolerance The optimizer's convergence tolerance, smaller value will lead to higher accuracy with the cost
    *                  of more iterations
-   * @param enableOptimizationStateTracker Whether to enable the optimization state tracker, which stores the
-   *                                       per-iteration log information of the running optimizer
    * @param constraintMap An optional mapping of feature indices to box constraints
    * @param treeAggregateDepth The depth for tree aggregation
    * @return The trained models in the form of Map(key -> model), where key is the String typed corresponding
@@ -61,10 +59,9 @@ object ModelTraining extends Logging {
       normalizationContext: NormalizationContext,
       maxNumIter: Int,
       tolerance: Double,
-      enableOptimizationStateTracker: Boolean,
       constraintMap: Option[Map[Int, (Double, Double)]],
       treeAggregateDepth: Int,
-      useWarmStart: Boolean): List[(Double, _ <: GeneralizedLinearModel, Option[OptimizationStatesTracker])] =
+      useWarmStart: Boolean): List[(Double, GeneralizedLinearModel, OptimizationStatesTracker)] =
     trainGeneralizedLinearModel(
       trainingData,
       taskType,
@@ -74,7 +71,6 @@ object ModelTraining extends Logging {
       normalizationContext,
       maxNumIter,
       tolerance,
-      enableOptimizationStateTracker,
       constraintMap,
       Map.empty,
       treeAggregateDepth,
@@ -94,8 +90,6 @@ object ModelTraining extends Logging {
    * @param maxNumIter Maximum number of iterations to run
    * @param tolerance The optimizer's convergence tolerance, smaller value will lead to higher accuracy with the cost
    *                  of more iterations
-   * @param enableOptimizationStateTracker Whether to enable the optimization state tracker, which stores the
-   *                                       per-iteration log information of the running optimizer
    * @param constraintMap An optional mapping of feature indices to box constraints
    * @param warmStartModels Map of (lambda -> model) to use for warm start training
    * @param treeAggregateDepth The depth for tree aggregation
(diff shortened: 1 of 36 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -299,40 +292,40 @@ object RandomEffectCoordinate {
...
-              (updatedModel, stateTrackers)
+              (updatedModel, Some(stateTrackers))
 
             case (localModel, _) =>
               (localModel, None)
           }
+        modelsAndTrackers.persist(StorageLevel.MEMORY_ONLY_SER)
+
+        val models = modelsAndTrackers.mapValues(_._1)
+        val optimizationTracker = RandomEffectOptimizationTracker(modelsAndTrackers.flatMap(_._2._2))
```
cmjiang: Two methods to access elements in a tuple: 1) using underscore + index; 2) using assigned names to elements. Which one is preferred in Photon ML? I feel using underscore is less readable.

ashelkovnykov: I think it depends on the situation and the name of the tuple. If you feel it's clear what the type of the anonymous value is, then using the underscore is fine. If you feel that it benefits from additional information, then a named parameter is better. Do you think it's unclear in this situation? I thought the name `modelsAndTrackers` for the parameter makes the types in the tuple clear.

cmjiang: The name `modelsAndTrackers` makes the type clear, but it's not obvious for me of the second `_2` right after the track. Actually I am okay with both, as you already kept it quite consistent.

photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/Optimizer.scala
```diff
@@ -31,21 +31,19 @@ import com.linkedin.photon.ml.util._
...
     maxNumIterations: Int,
     normalizationContext: BroadcastWrapper[NormalizationContext],
-    constraintMap: Option[Map[Int, (Double, Double)]],
-    val isTrackingState: Boolean)
+    constraintMap: Option[Map[Int, (Double, Double)]])
   extends Serializable
   with Logging {
 
   /**
    * OptimizationStatesTracker optionally tracks all the states of the optimizer and times the optimization.
```
yunboouyang: shall we delete `optionally`?

photon-client/src/main/scala/com/linkedin/photon/ml/Params.scala
```diff
@@ -99,6 +99,7 @@ class Params extends PalDBIndexMapParams {
   /**
    * Whether to enable the optimization tracker, which stores the per-iteration log information of the running optimizer
    */
+  @Deprecated
   var enableOptimizationStateTracker: Boolean = true
```
yunboouyang: Do we need this deprecated field?

photon-api/src/main/scala/com/linkedin/photon/ml/ModelTraining.scala
```diff
@@ -112,11 +106,10 @@ object ModelTraining extends Logging {
       normalizationContext: NormalizationContext,
       maxNumIter: Int,
       tolerance: Double,
-      enableOptimizationStateTracker: Boolean,
       constraintMap: Option[Map[Int, (Double, Double)]],
       warmStartModels: Map[Double, GeneralizedLinearModel],
       treeAggregateDepth: Int,
-      useWarmStart: Boolean): List[(Double, _ <: GeneralizedLinearModel, Option[OptimizationStatesTracker])] = {
+      useWarmStart: Boolean): List[(Double, GeneralizedLinearModel, OptimizationStatesTracker)] = {
```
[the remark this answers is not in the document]
ashelkovnykov: I was getting a casting error during testing, and this resolved it. Overall, it was an unnecessary type anyway since the explicit type of the models (e.g. logistic, linear, Poisson) is never checked/used and the models are cast to `GeneralizedLinearModel`.

photon-client/src/main/scala/com/linkedin/photon/ml/Driver.scala
```diff
@@ -344,58 +353,60 @@ protected[ml] class Driver(
...
-        val finalMetrics = Evaluation.evaluate(glm, validationData)
+          logger.info(s"Model with lambda = $lambda:")
 
-        finalMetrics
-          .keys
-          .toSeq
-          .sorted
-          .foreach { metric =>
-            logger.info(f"Metric: [$metric] value: ${finalMetrics(metric)}")
+          val trackerOpt = if(params.enableOptimizationStateTracker) {
```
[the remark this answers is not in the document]
ashelkovnykov: It can still be `false`, however in that case the optimization states are still tracked, they're just ignored afterwards. That's why this parameter is marked as deprecated - if we're always tracking the optimization states, we may as well dump the results to the log.

photon-lib/src/main/scala/com/linkedin/photon/ml/optimization/Optimizer.scala
ashelkovnykov: Yes, good catch.

photon-client/src/main/scala/com/linkedin/photon/ml/Params.scala
ashelkovnykov: I think so - see my reply to another one of your comments.

PR #439, opened 2019-09-30
Intercept separation

biyan-linkedin:
Implement intercept separation which will exclude penalizing intercept in l2 regularization.
 
Add UT and integTest for this feature.

PR #440, opened 2019-10-03
Disable L2 Regularization on Intercept

ashelkovnykov:
This is continuation of the work started by @biyan-linkedin in PR #439. Thank you for your work Bill!

Previously, Photon would regularize the entire coefficient vector, including the intercept coefficient if one was present. One of our users had this to say:
```
Penalizing intercept makes the behavior of L2 regularization depends
on the pos/neg ratio, and often leads to hard-to-interpret intercept.
Especially when we have a lot of training data and most of the features
are extremely sparse.
```

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/function/ObjectiveFunctionHelper.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/function/ObjectiveFunctionHelper.scala
@@ -26,7 +26,9 @@ import com.linkedin.photon.ml.optimization.game.CoordinateOptimizationConfigurat
  */
 object ObjectiveFunctionHelper {
 
-  type ObjectiveFunctionFactory = (CoordinateOptimizationConfiguration) => ObjectiveFunction
+  type ObjectiveFunctionFactoryFactory = CoordinateOptimizationConfiguration => Option[Int] => ObjectiveFunction
+  type DistributedObjectiveFunctionFactory = Option[Int] => DistributedObjectiveFunction
+  type SingleNodeObjectiveFunctionFactory = Option[Int] => SingleNodeObjectiveFunction
 
   /**
    * Construct a factory function for building [[ObjectiveFunction]] objects.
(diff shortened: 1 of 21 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/function/ObjectiveFunctionHelper.scala
```diff
@@ -26,7 +26,9 @@ import com.linkedin.photon.ml.optimization.game.CoordinateOptimizationConfigurat
  */
 object ObjectiveFunctionHelper {
 
-  type ObjectiveFunctionFactory = (CoordinateOptimizationConfiguration) => ObjectiveFunction
+  type ObjectiveFunctionFactoryFactory = CoordinateOptimizationConfiguration => Option[Int] => ObjectiveFunction
```
yunboouyang: Do we need to add parentheses for readability?

photon-api/src/main/scala/com/linkedin/photon/ml/function/svm/SmoothedHingeLossFunction.scala
```diff
@@ -94,19 +94,22 @@ object SmoothedHingeLossFunction {
...
-    (config: CoordinateOptimizationConfiguration) => {
-      config match {
-        case fEOptConfig: FixedEffectOptimizationConfiguration =>
+  def buildFactory
+      (treeAggregateDepth: Int)
+      (config: CoordinateOptimizationConfiguration): Option[Int] => ObjectiveFunction =
+
+    config match {
+      case fEOptConfig: FixedEffectOptimizationConfiguration =>
+        (interceptIndexOpt: Option[Int]) =>
```
yunboouyang: To clarify, will interceptIndexOpt be used in DistributedSmoothedHingeLossFunction

li-ashelkov: No - I did some tests and found that the smoothed-hinge loss function works significantly worse than logistic regression. In a future PR, I plan to get rid of it. I told Bill in advance, so that he wouldn't waste his time updating the smoothed-hinge classes.

photon-api/src/main/scala/com/linkedin/photon/ml/function/ObjectiveFunctionHelper.scala
li-ashelkov: It's a style preference question, I suppose. We can go through later and make sure that we're being consistent. IDEA complains about unnecessary parentheses if we include them, which is why I removed them.

PR #442, opened 2019-10-16
Fix a bug in calculating RMSE off by sqrt(2)

yunboouyang:
Changes:
1. Add correct calculation of squared loss, which will be used to compute Rooted Mean Squared Error (RMSE) by taking average and squared root. 
2. Multiply the error threshold by sqrt(2) since RMSE calculation has been changed.

### DIFF
--- a/photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
+++ b/photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
@@ -433,7 +433,7 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
 
     runDriver(args.put(GameTrainingDriver.rootOutputDirectory, outputDir))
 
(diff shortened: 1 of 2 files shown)


photon-client/src/integTest/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriverIntegTest.scala
```diff
@@ -433,7 +433,7 @@ class GameTrainingDriverIntegTest extends SparkTestUtils with GameTestUtils with
 
     runDriver(args.put(GameTrainingDriver.rootOutputDirectory, outputDir))
 
-    compareModelEvaluation(new Path(outputDir, "best"), trainedMixedModelPath, 0.005)
+    compareModelEvaluation(new Path(outputDir, "best"), trainedMixedModelPath, 0.007)
```
ashelkovnykov: This value shouldn't need to be updated, unlike the others. The other values are RMSE scores which the result should either match or improve upon. This value is an epsilon value for comparing two RMSE values.

yunboouyang: Addressed comments

PR #443, opened 2019-10-18
RDD join refactor

ashelkovnykov:
- Fix join issues
- Document joins, make it easier to understand why we join RDDs as we do
- Remove unused projectors after applying data lower bound threshold

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
@@ -20,7 +20,6 @@ import org.apache.spark.storage.StorageLevel
 
 import com.linkedin.photon.ml.data._
 import com.linkedin.photon.ml.data.scoring.CoordinateDataScores
-import com.linkedin.photon.ml.function.ObjectiveFunctionHelper.SingleNodeObjectiveFunctionFactory
 import com.linkedin.photon.ml.function.SingleNodeObjectiveFunction
 import com.linkedin.photon.ml.model.{Coefficients, DatumScoringModel, RandomEffectModel}
 import com.linkedin.photon.ml.normalization.NormalizationContext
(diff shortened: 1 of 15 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
```diff
@@ -79,15 +80,19 @@ protected[ml] class RandomEffectDataset(
...
 
-    // Add scores to active data offsets
-    val scoresGroupedByRandomEffectId = scores
-      .scoresRdd
-      .join(activeUniqueIdToRandomEffectIds, uniqueIdPartitioner)
-      .map { case (uniqueId, (score, reId)) => (reId, (uniqueId, score)) }
+    // It's possible that other coordinates did not score some data. Since we're trying to add scores to the offset and
+    // the default score is 0, the result of a left join vs. an inner join is the same. However, an inner join will drop
+    // data which does not have a score. Thus, we need a left join.
+    val scoresGroupedByRandomEffectId = activeUniqueIdToRandomEffectIds
```
cmjiang: Good catch.

photon-lib/src/main/scala/com/linkedin/photon/ml/evaluation/EvaluationSuite.scala
```diff
@@ -55,10 +56,12 @@ class EvaluationSuite(
    */
   protected[ml] def evaluate(scores: RDD[(UniqueSampleId, Double)]): EvaluationResults = {
 
-    val scoresAndLabelsAndWeights = scores
-      .join(labelAndOffsetAndWeights)
-      .mapValues { case (score, (label, offset, weight)) =>
-        (score + offset, label, weight)
+    // Possible for all models to be missing a score for some datum, meaning the score for a datum is missing even after
+    // summing scores from all models. Thus, need a leftOuterJoin.
+    val scoresAndLabelsAndWeights = labelAndOffsetAndWeights
```
cmjiang: Good catch.

photon-api/src/main/scala/com/linkedin/photon/ml/algorithm/RandomEffectCoordinate.scala
```diff
@@ -228,12 +227,16 @@ object RandomEffectCoordinate {
...
+    // All 3 RDDs involved in the joins below use the same partitioner
+
+    // Optimization problems are created for each entity with a projector, and thus guaranteed to match active data
+    // exactly (see RandomEffectDataset.apply)
     val dataAndOptimizationProblems = randomEffectDataset
       .activeData
       .join(randomEffectOptimizationProblem.optimizationProblems)
 
-    // Left join the models to data and optimization problems for cases where we have a prior model but no new data
+    // Left join the models to the (data, optimization problem) tuple for cases where we have a prior model but no new
```
yunboouyang: To confirm, does a prior model only come from warm start? If so, prior models and new data might not match and we can do left join. Do we still keep this model and output it in this case?

ashelkovnykov: That's right - if Photon ML trains a warm-start model, it will keep existing models for which no new data is present.

PR #444, opened 2019-10-18
Fix bug which would keep L2 component of gradient and Hessian for intercept term

ashelkovnykov:


photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -196,25 +208,52 @@ trait L2RegularizationTwiceDiff extends TwiceDiffFunction with L2RegularizationD
...
+
+      case None =>
+    }
+
+    l2HessianVector
+  }
 
   /**
    * Compute the Hessian diagonal of the L2 regularization term.
    *
```
yunboouyang: Need to add `@param` for `DIMENSION`

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -174,7 +186,7 @@ trait L2RegularizationTwiceDiff extends TwiceDiffFunction with L2RegularizationD
       coefficients: Coefficients,
       multiplyVector: Coefficients,
       normalizationContext: BroadcastWrapper[NormalizationContext]): Vector[Double] =
-    super.hessianVector(input, coefficients, multiplyVector, normalizationContext) +
+    super.hessianVector(input, coefficients, multiplyVector, normalizationContext) +:+
```
yunboouyang: Why do we change to `+:+`? `+` supports element-wise addition

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
```diff
@@ -196,25 +208,52 @@ trait L2RegularizationTwiceDiff extends TwiceDiffFunction with L2RegularizationD
...
+
+      case None =>
+    }
+
+    l2HessianDiagonal
+  }
+
+  /**
+   * Compute the Hessian matrix of the L2 regularization term.
+   *
```
cmjiang: Miss a param here too.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/L2Regularization.scala
ashelkovnykov: Trying to remember - I think my train of thought was that `+` would be for scalar values, and `+:+` would be for vector values. I think I was trying to make it clear what type of values are being summed by looking at which symbol is being used.

PR #453, opened 2020-02-14
Fix bug which caused all per-entity data to be sent to a single partitioner before sampling

ashelkovnykov:
Results in crashes for cases which need the 'active.data.upper.bound'.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
@@ -271,37 +271,38 @@ object RandomEffectDataset {
     //
 
     val keyedGameDataset = generateKeyedGameDataset(gameDataset, randomEffectDataConfiguration)
-    keyedGameDataset.persist(StorageLevel.MEMORY_ONLY_SER).count
+    keyedGameDataset.persist(StorageLevel.MEMORY_AND_DISK_SER).count
 
     // In this RDD, there is a projector for every entity (even those which may later be filtered by the lower bound)
     val unfilteredProjectors = generateLinearSubspaceProjectors(keyedGameDataset, randomEffectPartitioner)
     unfilteredProjectors.persist(storageLevel).count
 
-    val projectedKeyedGameDataset = generateProjectedDataset(keyedGameDataset, unfilteredProjectors, randomEffectPartitioner)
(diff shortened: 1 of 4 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
```diff
@@ -406,29 +406,17 @@ object RandomEffectDataset {
...
-        val projectorLookupTable = projectorsIt.toMap
-
-        dataIt.map { case (rEID, (uID, LabeledPoint(label, features, offset, weight))) =>
-
-          val projector = projectorLookupTable(rEID)
-          val projectedFeatures = projector.projectForward(features)
-
-          (rEID, (uID, LabeledPoint(label, projectedFeatures, offset, weight)))
-        }
+      .join(projectors)
```
limkothari: Is it possible that the projectors are still causing the data to be keyed by partitions and sent to single executor to be loaded in the memory?

ashelkovnykov: I think I see what you mean - @yunboouyang and I just spoke and we're working on another solution. Yes, the same partitioner is being used even without being specified as the keyed data loses its partitioner, so Spark reuses the partitioner of the projector RDD.

limkothari: Thanks @ashelkovnykov ! Please let me know the patch and i can verify it with my dataset :)

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
```diff
@@ -271,37 +271,38 @@ object RandomEffectDataset {
     //
 
     val keyedGameDataset = generateKeyedGameDataset(gameDataset, randomEffectDataConfiguration)
-    keyedGameDataset.persist(StorageLevel.MEMORY_ONLY_SER).count
+    keyedGameDataset.persist(StorageLevel.MEMORY_AND_DISK_SER).count
```
yunboouyang: Curious to know the reason of changing the storage level here

yunboouyang:
Fixed the bug in https://github.com/linkedin/photon-ml/pull/466

PR #455, opened 2020-02-14
Remove experimental SmoothedHingeLoss training task

ashelkovnykov:
- Experimental results when comparing Smoothed Hinge Loss to Logistic Regression were substantially worse and also increased training time significantly due to slower convergence
- Removed GLMLossFunction and derived classes, since all training tasks in Photon ML are GLM tasks


ashelkovnykov:
@cmjiang @yunboouyang @lguo 
Updated code for latest master, requesting review

photon-api/src/main/scala/com/linkedin/photon/ml/SparkSessionConfiguration.scala
```diff
@@ -52,6 +52,8 @@ object SparkSessionConfiguration {
     classOf[GeneralizedLinearModel],
     classOf[GeneralizedLinearOptimizationProblem[_]],
     classOf[GLMOptimizationConfiguration],
+    classOf[HessianDiagonalAggregator],
```
yunboouyang, line 55: Why do we need to add this class?

yunboouyang (approved):
LGTM

photon-api/src/main/scala/com/linkedin/photon/ml/SparkSessionConfiguration.scala
ashelkovnykov: I noticed that we had the `ValueAndGradientAggregator` and `HessianVectorAggregator` being registered with Kryo, but not the `HessianDiagonalAggregator` or the `HessianMatrixAggregator`. Since all four function the same way, it stands to reason that they should either all be registered or none of them be registered. Since objects of these four class types will for sure be distributed, therefore they should all be registered.

Apparently, Kryo was able to figure this out on its own, but I think it best to make it explicit.

PR #457, opened 2020-02-20
Remove Pearson correlation score based feature filtering for random effects

ashelkovnykov:


yunboouyang (approved):

ashelkovnykov:
Updated for latest master.
@cmjiang @lguo Requesting review

Is Photon ML being published to an external artifact repository now? Which files need to be updated before a merge?

cmjiang:
@ashelkovnykov The migration to external artifact repo is done. Please update `release-bintray.gradle` for new versions and `CHANGELOG.md` for new changes.

ashelkovnykov:
@cmjiang 
Does each change necessarily publish a new jar? What if we want to have multiple GitHub PRs condensed into one release, what should I do for that case? Asking because all five of my "ready to go" PRs should technically be a new major version, but I don't want to suddenly take us to version 25 of Photon ML. Thank you

cmjiang:
@ashelkovnykov Good question. We don't need to push each change to a new jar. And we definitely don't want to suddenly bump 5 major version up. However the problem is that the version is explicitly pinned in the `release-bintray.gradle`. I can't think of a neat solution for now. @lguo @yunboouyang Any idea?

cmjiang:
Since we don't have to push each change to a new jar, we can just keep the same version number for those PRs without new jars. The problem is that the same version number may have different versions of code.

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
```diff
@@ -287,27 +286,24 @@ object RandomEffectDataset {
       randomEffectDataConfiguration,
       randomEffectPartitioner)
     val projectedGroupedActiveData = generateProjectedActiveData(unfilteredActiveData, unfilteredProjectors)
-    val projectedUnfilteredActiveData = featureSelectionOnActiveData(
-      projectedGroupedActiveData,
-      randomEffectDataConfiguration.numFeaturesToSamplesRatioUpperBound)
```
ashelkovnykov, line 292: Forgot to remove the `numFeaturesToSamplesRatioUpperBound` from the data configuration

PR #459, opened 2020-02-27
Fix a hyperparameter tuning dimension mismatch bug

yunboouyang:
Fix a hyperparameter tuning dimension calculation bug. 
1. The dimension should be equal to the number of lambdas (magnitude) and alphas (mixing weights). 
2. Right now the dimension is equal to the number of coordinates.


photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -660,7 +660,18 @@ object GameTrainingDriver extends GameDriver {
...
+
+        val dimension = baseConfig.toSeq.map {
+          case (_, config: GLMOptimizationConfiguration) =>
+            config.regularizationContext.regularizationType match {
+              case RegularizationType.ELASTIC_NET => 2
+              case RegularizationType.L2 => 1
+              case RegularizationType.L1 => 1
+              case RegularizationType.NONE => 0
+            }
+          case _ => throw new IllegalArgumentException(s"Unknown optimization config!")
```
cmjiang: Can you change the error message to be more explicit? Like "Unknown regularization type."

PR #460, opened 2020-02-27
Add a dummy optimization states tracker

yunboouyang:
Add a dummy optimization states tracker as a workaround for null pointer exception issue when training random effects models.


yunboouyang:
no longer needed.

PR #462, opened 2020-02-28
Construct projectors incorporating with prior models

yunboouyang:
When constructing subspace linear projectors, prior model's active indices should be union with data's active indices, otherwise some of prior model's coefficients will be dropped in the training process. An union step is added in this PR.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
@@ -26,8 +26,10 @@ import org.apache.spark.{Partitioner, SparkContext}
 import com.linkedin.photon.ml.Types.{FeatureShardId, REId, REType, UniqueSampleId}
 import com.linkedin.photon.ml.constants.MathConst
 import com.linkedin.photon.ml.data.scoring.CoordinateDataScores
+import com.linkedin.photon.ml.model.RandomEffectModel
 import com.linkedin.photon.ml.projector.LinearSubspaceProjector
 import com.linkedin.photon.ml.spark.{BroadcastLike, RDDLike}
+import com.linkedin.photon.ml.supervised.model.GeneralizedLinearModel
 import com.linkedin.photon.ml.util.VectorUtils
 
 /**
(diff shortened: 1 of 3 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -507,12 +508,32 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
...
 
-          val randomEffectDataset = RandomEffectDataset(
-            gameDataset,
-            reConfig,
-            rePartitioner,
-            existingModelKeysRddOpt,
-            StorageLevel.DISK_ONLY)
+          val randomEffectDataset = if(initialModelOpt.isDefined) {
+            val reModels = initialModelOpt.get.getModel(coordinateId).map {
+              case rem: RandomEffectModel =>
```
cmjiang: The parameter `rem` is not self-explanatory. Maybe use `reModel`?

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
```diff
@@ -507,12 +508,32 @@ class GameEstimator(val sc: SparkContext, implicit val logger: Logger) extends P
...
+                rem
+
+              case other =>
```
cmjiang: If the `initialModelOpt` contains a fixed-effect model, this will throw an exception but we should let it be ignored instead.

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
```diff
@@ -251,14 +253,17 @@ object RandomEffectDataset {
    * Build a new [[RandomEffectDataset]] from the raw data using the given configuration.
    *
    * @param gameDataset The [[RDD]] of [[GameDatum]] used to generate the random effect dataset
+   * @param priorRandomEffectModelOpt The optional prior random effects models used in building projectors
    * @param randomEffectDataConfiguration The data configuration for the random effect dataset
    * @param randomEffectPartitioner A specialized partitioner to co-locate all data from a single entity, while keeping
    *                                the data distribution equal amongst partitions
    * @param existingModelKeysRddOpt Optional set of entities that have existing models
+   * @param storageLevel Spark storage level for RDDs
```
cmjiang: Why do we need to add this parameter?

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
yunboouyang: fixed

photon-api/src/main/scala/com/linkedin/photon/ml/estimators/GameEstimator.scala
yunboouyang: `coordinateId` will point to a random effect coordinate since the config is `RandomEffectDataConfiguration`

photon-api/src/main/scala/com/linkedin/photon/ml/data/RandomEffectDataset.scala
yunboouyang: Only added description here

PR #464, opened 2020-03-06
Update optimization problems to return updated model and states tracker all together to avoid NPEs.

lguo:
Update optimization problems to return updated model and states tracker
all together to avoid NPEs.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
@@ -14,7 +14,7 @@
  */
 package com.linkedin.photon.ml.optimization
 
-import breeze.linalg.{Vector, cholesky, diag}
+import breeze.linalg.{cholesky, diag, Vector => BVector}
 import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.StorageLevel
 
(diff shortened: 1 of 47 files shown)


photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
```diff
@@ -81,7 +81,7 @@ protected[ml] class DistributedOptimizationProblem[Objective <: DistributedObjec
    * @param coefficients The feature coefficients means
    * @return An optional feature coefficient variances vector
    */
-  override def computeVariances(input: RDD[LabeledPoint], coefficients: Vector[Double]): Option[Vector[Double]] = {
+  override def computeVariances(input: RDD[LabeledPoint], coefficients: BVector[Double]): Option[BVector[Double]] = {
```
cmjiang: Why do we need this?

lguo: It is not necessary in this fix. I copy it from the migration codes and we would need spark ml vector so we need to rename breeze vector to BVector. We could remove it in this commit.

yunboouyang: I think we can remove it to keep this commit clean.

PR #469, opened 2020-04-13
Minor Feature Normalization Refactor

yunboouyang:
Had a minor refactor of feature normalization related code. Changes are:

1. Fixed a bug in calculating variance when feature normalization is enabled. Previous variance calculation function `createModel` was wrong because means and variance cannot share the same transformation. The correct way of calculating variance when feature normalization is enabled is: first transform the means back to the original space, and then call `computeVariances` function. 

2. Added `normalizationContext` in `PriorDistribution.scala` to enable mean and variance normalization in incremental learning.

### DIFF
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
@@ -125,9 +125,10 @@ protected[ml] class DistributedOptimizationProblem[Objective <: DistributedObjec
 
     val normalizationContext = optimizer.getNormalizationContext
     val (optimizedCoefficients, stateTracker, _) = optimizer.optimize(objectiveFunction, initialModel.coefficients.means)(input)
-    val optimizedVariances = computeVariances(input, optimizedCoefficients)
+    val origOptimizedCoef = normalizationContext.value.coefToOriginalSpace(optimizedCoefficients)
+    val origOptimizedVar = computeVariances(input, origOptimizedCoef)
 
-    (createModel(normalizationContext, optimizedCoefficients, optimizedVariances), stateTracker)
+    (createModel(origOptimizedCoef, origOptimizedVar), stateTracker)
   }
 
   /**
--- a/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/GeneralizedLinearOptimizationProblem.scala
+++ b/photon-api/src/main/scala/com/linkedin/photon/ml/optimization/GeneralizedLinearOptimizationProblem.scala
@@ -68,22 +68,6 @@ protected[ml] abstract class GeneralizedLinearOptimizationProblem[Objective <: O
     glmConstructor(Coefficients(coefficients, variances))
 
   /**
-   * Create a GLM from given normalized coefficients (potentially including intercept)
-   *
-   * @param normalizationContext The normalization context
-   * @param coefficients The feature coefficients means
-   * @param variances The feature coefficient variances
-   * @return A GLM with the provided feature coefficients
-   */
-  protected def createModel(
-      normalizationContext: BroadcastWrapper[NormalizationContext],
-      coefficients: Vector[Double],
-      variances: Option[Vector[Double]]): GeneralizedLinearModel =
-    createModel(
-      normalizationContext.value.modelToOriginalSpace(coefficients),
-      variances.map(normalizationContext.value.modelToOriginalSpace))
-
-  /**
    * Compute coefficient variances
    *
    * @param input The training data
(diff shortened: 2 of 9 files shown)


photon-lib/src/main/scala/com/linkedin/photon/ml/normalization/NormalizationContext.scala
```diff
@@ -126,6 +126,36 @@ protected[ml] class NormalizationContext(
 
       outputCoef
     }
+
+  /**
+   * Transform the variance of the model coefficients of the original space to the transformed space. Will be used in
+   * constructing the prior.
+   *
+   * w' = w ./ factor
+   * b' = w^T^ shift + b
```
cmjiang, line 135: Remove the extra "^".

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/SingleNodeOptimizationProblem.scala
```diff
@@ -91,9 +91,10 @@ protected[ml] class SingleNodeOptimizationProblem[Objective <: SingleNodeObjecti
 
     val normalizationContext = optimizer.getNormalizationContext
     val (optimizedCoefficients, stateTracker, _) = optimizer.optimize(objectiveFunction, initialModel.coefficients.means)(input)
-    val optimizedVariances = computeVariances(input, optimizedCoefficients)
+    val origOptimizedCoef = normalizationContext.value.coefToOriginalSpace(optimizedCoefficients)
+    val origOptimizedVar = computeVariances(input, origOptimizedCoef)
```
cmjiang, line 95: Thanks for fixing this bug. We do need to transform the coefficients back to the original space then compute the variance.

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/DistributedOptimizationProblem.scala
```diff
@@ -125,9 +125,10 @@ protected[ml] class DistributedOptimizationProblem[Objective <: DistributedObjec
 
     val normalizationContext = optimizer.getNormalizationContext
     val (optimizedCoefficients, stateTracker, _) = optimizer.optimize(objectiveFunction, initialModel.coefficients.means)(input)
-    val optimizedVariances = computeVariances(input, optimizedCoefficients)
+    val origOptimizedCoef = normalizationContext.value.coefToOriginalSpace(optimizedCoefficients)
```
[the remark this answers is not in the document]
yunboouyang: Yes, you are correct. The process is as follows:
1. The data and the prior are un-normalized when we load them in the beginning. 
2. Normalize both the data and the prior; 
3. Get coefficients in the normalized space; (line 127)
4. Transform coefficients to the original space; (line 128)
5. Compute the variance using the coefficients and data in the original space (line 129)
6. Output this model as the new prior (line 131)

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/GeneralizedLinearOptimizationProblem.scala
```diff
@@ -67,22 +67,6 @@ protected[ml] abstract class GeneralizedLinearOptimizationProblem[Objective <: O
...
 
-  /**
-   * Create a GLM from given normalized coefficients (potentially including intercept)
-   *
-   * @param normalizationContext The normalization context
-   * @param coefficients The feature coefficients means
-   * @param variances The feature coefficient variances
-   * @return A GLM with the provided feature coefficients
-   */
-  protected def createModel(
```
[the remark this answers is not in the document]
yunboouyang: `createModel` function in line 67 is the function we are using (does not take `normalizationContext`). We don't use this function because variance has the wrong transformation. `createModel` function is used when we have coefficients and variances in the original space so we don't need to do any transformation inside. See line 131 in `DistributedOptimizationProblem.scala`

photon-api/src/main/scala/com/linkedin/photon/ml/optimization/SingleNodeOptimizationProblem.scala
yunboouyang: Yes, this is the correct and easiest way to compute variance when feature normalization is involved.

photon-lib/src/main/scala/com/linkedin/photon/ml/function/PriorDistribution.scala
```diff
@@ -40,6 +40,20 @@ trait PriorDistribution extends ObjectiveFunction {
...
+   * Transform prior means and prior variances to the transformed space using NormalizationContext
+   *
+   * @param normalizationContext The normalization context
+   * @return The tuple of transformed prior means and inverse transformed prior variances
+   */
+  protected[ml] def normalizePrior(
+    normalizationContext: BroadcastWrapper[NormalizationContext]): (Vector[Double], Vector[Double]) = {
+    val transformedPriorMeans = normalizationContext.value.coefToTransformedSpace(priorMeans)
+    val transformedPriorVar = normalizationContext.value.varToTransformedSpace(priorVariances)
+    val inverseTransformedPriorVar = VectorUtils.invertVectorWithZeroHandler(transformedPriorVar, l2RegWeight).toDenseVector
```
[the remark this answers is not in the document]
yunboouyang: When we prepare for training, we normalize the prior by normalizing both means and variances. In all usage of using the prior variance, we all use its inverse, like `(coefficients - prior means)^2 / prior variance`, so we inverse it element-wise here.

photon-lib/src/main/scala/com/linkedin/photon/ml/normalization/NormalizationContext.scala
yunboouyang: I saw in all previous description we use `w^T^` like line 106 and line 72, so I assume `w^T^` is `w^T`.

PR #473, opened 2020-04-21
Add per group evaluation output

cmjiang:
When the number of random effect entities is small, e.g. in the cohort-based models, users would like to check the evaluations on each group (data grouped by member id, job id, cohort id etc) and analyze the importance of each group to overall evaluation. This PR aims to output the per group evaluation in separate output files.

### DIFF
--- a/jcenter/bintray.gradle
+++ b/jcenter/bintray.gradle
@@ -43,6 +43,7 @@
  * */
 
 apply plugin: 'com.jfrog.bintray'
+apply plugin: 'maven-publish'
 
 version = libraryVersion
 
@@ -103,6 +104,23 @@ bintray {
         passphrase = properties.getProperty("bintray.gpg.password")
         //Optional. The passphrase for GPG signing'
       }
+    }
+  }
+}
+
+publishToMavenLocal.dependsOn assemble
+publishToMavenLocal.dependsOn sourcesJar
+
+publishing {
+  repositories {
+    maven {
+      url "file://~/.m2/repository"
+    }
+  }
+  publications {
+    testing(MavenPublication) {
+      from components.java
+      artifact sourcesJar
     }
   }
 }
--- a/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
+++ b/photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
@@ -70,6 +70,7 @@ object GameTrainingDriver extends GameDriver {
   protected[training] val MODELS_DIR = "models"
   protected[training] val MODEL_SPEC_DIR = "model-spec"
   protected[training] val BEST_MODEL_DIR = "best"
+  protected[training] val GROUP_EVAL_DIR = "group-eval"
 
   protected[training] var sc: SparkContext = _
   protected[training] implicit var logger: PhotonLogger = _
@@ -175,6 +176,10 @@ object GameTrainingDriver extends GameDriver {
     "incremental training",
     "Flag to enable incremental training.")
 
+  val savePerGroupEvaluationResult: Param[Boolean] = ParamUtils.createParam[Boolean](
+    "save per-group evaluation result",
+    "Flag to save group evaluation result for random effect in a separate output file.")
+
   //
   // Initialize object
   //
@@ -221,6 +226,7 @@ object GameTrainingDriver extends GameDriver {
     setDefault(timeZone, Constants.DEFAULT_TIME_ZONE)
     setDefault(ignoreThresholdForNewModels, false)
     setDefault(incrementalTraining, false)
+    setDefault(savePerGroupEvaluationResult, true) //TODO: Testing purpose, default to false before push
   }
 
   /**
@@ -470,6 +476,7 @@ object GameTrainingDriver extends GameDriver {
         .setIgnoreThresholdForNewModels(getOrDefault(ignoreThresholdForNewModels))
         .setUseWarmStart(true)
         .setIncrementalTraining(getOrDefault(incrementalTraining))
+        .setSavePerGroupEvaluationResult(getOrDefault(savePerGroupEvaluationResult))
 
       get(inputColumnNames).foreach(estimator.setInputColumnNames)
       modelOpt.foreach(estimator.setInitialModel)
@@ -498,6 +505,12 @@ object GameTrainingDriver extends GameDriver {
 
     Timed("Save models") {
       saveModelToHDFS(featureIndexMapLoaders, outputModels, bestModel)
+    }
+
+    if (getOrDefault(savePerGroupEvaluationResult)) {
+      Timed("Save per-group evaluation result") {
+        savePerGroupEvaluationToHDFS(outputModels)
+      }
     }
   }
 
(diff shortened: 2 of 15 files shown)


photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
```diff
@@ -221,6 +226,7 @@ object GameTrainingDriver extends GameDriver {
     setDefault(timeZone, Constants.DEFAULT_TIME_ZONE)
     setDefault(ignoreThresholdForNewModels, false)
     setDefault(incrementalTraining, false)
+    setDefault(savePerGroupEvaluationResult, true) //TODO: Testing purpose, default to false before push
```
junshi15, line 229: remove comment

photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -365,4 +366,44 @@ object IOUtils {
...
+  /**
+   * Save GAME per-group evaluation in terms of (random effect id, evaluation value) to HDFS.
+   *
+   * @param sc The Spark context
+   * @param outputDir The directory in HDFS where to save the evaluation
+   * @param evaluationResultsOpt An Option of evaluation results
+   * @param logger The logger instance for the application
+   */
+  def saveGameEvaluationToHDFS(
+    sc: SparkContext,
```
junshi15, line 379: Since you are using Spark, and the number of groups can be large, potentially in millions, why don't you use spark.dataframe to save in avro format. Faster and more scalable.

photon-lib/src/main/scala/com/linkedin/photon/ml/evaluation/EvaluationSuite.scala
```diff
@@ -68,10 +70,14 @@ class EvaluationSuite(
     val evaluations = evaluators
       .map {
         case evaluator: SingleEvaluator =>
-          (evaluator.evaluatorType, evaluator.evaluate(scoresAndLabelsAndWeights.values))
+          (evaluator.evaluatorType, (evaluator.evaluate(scoresAndLabelsAndWeights.values), None))
 
         case multiEvaluator: MultiEvaluator =>
-          (multiEvaluator.evaluatorType, multiEvaluator.evaluate(scoresAndLabelsAndWeights))
+          if (savePerGroupEvaluation) {
+            (multiEvaluator.evaluatorType, multiEvaluator.evaluatePerGroup(scoresAndLabelsAndWeights))
```
yunboouyang, line 77: What if we need to train multiple random effects? For example we want to first train per-member and then train per-job models based on the residuals from per-member models. Previously we get metrics from per-member models only as well as metrics from per-member + per-job models. If we want to know member-level metrics but per-member models train first, we need to get the final score and choose `member` instead of `job` as the group. Is it implemented in this way?

cmjiang: This is a good question. I think no matter of the order of random effect training, the final per-member level metric is computed by grouping on the member level, same for per-job metric. That's the same as computing the average of per-member metric previously.

photon-client/src/main/scala/com/linkedin/photon/ml/cli/game/training/GameTrainingDriver.scala
cmjiang: This is for testing purpose. The default value should be false. Will remove before push.

photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
cmjiang: Output format is changed to Avro.

photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -365,4 +367,41 @@ object IOUtils {
...
+    evaluationResultsOpt match {
+      case Some(evaluationResults) => evaluationResults.evaluations.foreach {
+        case (evaluatorType, (_, Some(groupEval))) => {
+          val evaluatorName = evaluatorType.name.split(MultiEvaluatorType.shardedEvaluatorIdNameSplitter)
+          logger.debug(s"Save per-group evaluation of ${evaluatorName(0)} on ${evaluatorName(1)}")
+          val evalOutputDir = new Path(new Path(outputDir, evaluatorName(0)), evaluatorName(1))
+          sparkSession
+            .createDataFrame(groupEval)
+            .toDF(evaluatorName(1), evaluatorName(0))
+            .coalesce(1)
```
junshi15, line 394: not sure you want to do coalesce here, 
1) it is slow
2) if there were 1 million groups, the single file will be large.

cmjiang: No, the single file will be small even with 1 million groups. Note that each record is a tuple of (String, Double). I generated an example output which contains 400 records in 2.75KB. So a million records cost roughly ~6.9MB.

cmjiang: If not apply coalesce, even for this 2.75KB Avro file, it outputs tens of tiny files.

photon-lib/src/main/scala/com/linkedin/photon/ml/evaluation/MultiEvaluator.scala
```diff
@@ -64,6 +64,37 @@ abstract class MultiEvaluator(
...
+    // EvaluationSuite guarantees that all validation data is in scoresAndLabelsAndWeights RDD, and ids RDD is directly
+    // mapped from validation data. Thus, inner join should be guaranteed to not lose any data.
+    val groupedData = scoresAndLabelsAndWeights
+      .join(ids)
+      .map { case (_, (scoreLabelAndWeight, id)) => (id, scoreLabelAndWeight) }
+      .groupByKey()
+
+      val perGroupEvaluation = groupedData
+        .mapValues(scoreLabelAndWeights => localEvaluator.evaluate(scoreLabelAndWeights.toArray))
+        .filter(results => !(results._2.isInfinite || results._2.isNaN))
```
yunboouyang, line 89: Curious on what circumstances we will see `inf` or `NaN`

photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
```diff
@@ -365,4 +367,41 @@ object IOUtils {
...
+    logger: PhotonLogger): Unit = {
+
+    evaluationResultsOpt match {
+      case Some(evaluationResults) => evaluationResults.evaluations.foreach {
+        case (evaluatorType, (_, Some(groupEval))) => {
+          val evaluatorName = evaluatorType.name.split(MultiEvaluatorType.shardedEvaluatorIdNameSplitter)
+          logger.debug(s"Save per-group evaluation of ${evaluatorName(0)} on ${evaluatorName(1)}")
+          val evalOutputDir = new Path(new Path(outputDir, evaluatorName(0)), evaluatorName(1))
+          sparkSession
+            .createDataFrame(groupEval)
```
yunboouyang, line 392: Do we need to define some pre-specified schema?

photon-lib/src/main/scala/com/linkedin/photon/ml/evaluation/MultiEvaluator.scala
cmjiang: When an entity (e.g. member, job) in the validation doesn't have a model trained, the output is NaN. I'm not sure when will the inf happen, but I still put the condition here as it was previously.

photon-client/src/main/scala/com/linkedin/photon/ml/util/IOUtils.scala
cmjiang: The schema is determined by the arguments in the following `toDF`.

junshi15:
squash the commits please.

build.gradle
```diff
@@ -17,6 +17,7 @@
 
 buildscript {
   repositories {
+    mavenLocal()
```
lguo, line 20: Please don't commit this -- it is only for local test.

build.gradle
```diff
@@ -48,6 +49,7 @@ allprojects {
   apply plugin: 'idea'
 
   repositories {
+    mavenLocal()
```
lguo, line 52: Please don't commit this -- it is only for local test.

jcenter/bintray.gradle
```diff
@@ -106,3 +107,17 @@ bintray {
     }
   }
 }
+
```
lguo, line 110: Please add the following lines:

publishToMavenLocal.dependsOn assemble
publishToMavenLocal.dependsOn sourcesJar

build.gradle
cmjiang: Done.

build.gradle
cmjiang: Done.

jcenter/bintray.gradle
cmjiang: Added.

PR #476, opened 2020-11-04
Union random effect prior model coefficients only for incremental tra…

cmjiang:
The prior models of random effect are used to join the random effect dataset and then to get the union of new features (in the dataset) and old features (in the prior model). This is useful in incremental training, but costs inefficiency in other jobs since "join" is expensive. In this PR, we change this join-and-union operation only for incremental training.


yunboouyang (approved):
LGTM!

junshi15:
LGTM.

